PR #32, opened 2018-06-28
fix: improve analyzer download console logs

michael-go:
print the download path, and add some padding to play nicer with the CLI
spinner.


lib/fetch-snyk-docker-analyzer.js
```diff
@@ -65,7 +66,8 @@ function fetch() {
           reject(err);
         })
         .on('finish', function () {
-          console.log(`\t=> download complete.`);
+          // padding log() with some whitespaces to not mix with the CLI spinner
+          console.log(`  => download complete.` + ' '.repeat(50));
```
darscan: Have you tried this on Node 4?

PR #37, opened 2018-06-28
fix: show better message when no Linux packages are detected

michael-go:
before that fix, we would get a `TypeError` trying to access fields of `undefined`


darscan:
@michael-go out-of-date with master?

darscan:
@michael-go I will pick this one up, rebase and update the analyzer version

darscan:
Rebased. Still awaiting new analyzer version

PR #38, opened 2018-07-05
Feat/new analyzer

darscan:
## What this does

- switch to TypeScript
- remove the old analyzer binary download, and stop using it
- internalise the old analyzer routines (ported to TS)

## Notes on development

- Was developed using Node 4.9.1 (to ensure it works with our minimum CLI version)


lib/analyzer/apk-analyzer.ts
```diff
@@ -0,0 +1,74 @@
...
+  return getPackages(targetImage)
+    .then(pkgs => ({
+      Image: targetImage,
+      AnalyzeType: 'Apk',
+      Analysis: pkgs,
+    }));
+}
+
+function getPackages(targetImage) {
+  return subProcess.execute('docker', [
```
orkamara: I think that it will be better to create a new class for "docker".
Right now `subProcess.execute('docker')` is the only common function, but in the near future there will be more.
Plus, it will prevent the code dup

darscan: Yeh, for sure. There is tons of cleanup that can be done here. Was trying to make the smallest PR possible for now :)

lib/analyzer/apk-analyzer.ts
```diff
@@ -0,0 +1,74 @@
...
+  const value = text.substr(2);
+  switch (key) {
+    case 'P': // Package
+      curPkg = {
+        Name: value,
+        Version: null,
+        Source: null,
+        Provides: [],
+        Deps: {},
+        AutoInstalled: null,
```
orkamara: shouldn't we initialize it with false? (the same for the rest of the analyzers)

orkamara: there is no place we change this value - didn't we implement this logic?

lib/analyzer/apk-analyzer.ts
```diff
@@ -0,0 +1,74 @@
...
+    case 'V': // Version
+      curPkg.Version = value;
+      break;
+    case 'p': // Provides
+      for (let name of value.split(' ')) {
+        name = name.split('=')[0];
+        curPkg.Provides.push(name);
+      }
+      break;
+    case 'r': // Depends
```
orkamara: is there is no diff btw 'r' and 'D'? maybe the comment should be different

lib/analyzer/apk-analyzer.ts
```diff
@@ -0,0 +1,74 @@
...
+    case 'D': // Depends
+      // tslint:disable-next-line:no-duplicate-variable
+      for (let name of value.split(' ')) {
+        if (name.charAt(0) !== '!') {
```
orkamara: flip the if - better indentation

lib/analyzer/apk-analyzer.ts
```diff
@@ -0,0 +1,74 @@
...
+          name = name.split('=')[0];
```
orkamara: the use of name twice is complicated. Maybe rename the one of them? (to better describe what are you doing)

lib/analyzer/apt-analyzer.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  return {
+    Image: targetImage,
+    AnalyzeType: 'Apt',
+    Analysis: pkgs,
+  };
+}
+
+async function dockerCat(targetImage, file): Promise<string> {
```
orkamara: I think you sould use this function in all of the other places you're cat-ing files from the img

lib/analyzer/apt-analyzer.ts
```diff
@@ -0,0 +1,115 @@
...
+      break;
+    case 'Version':
+      curPkg.Version = value;
+      break;
+    case 'Source':
+      curPkg.Source = value.trim().split(' ')[0];
+      break;
+    case 'Provides':
+      for (let name of value.split(',')) {
+        name = name.trim().split(' ')[0];
```
orkamara: the use of name twice is complicated. Maybe rename the one of them? (to better describe what are you doing)

lib/analyzer/apt-analyzer.ts
```diff
@@ -0,0 +1,115 @@
...
+        curPkg.Provides.push(name);
+      }
+      break;
+    case 'Pre-Depends':
+    case 'Depends':
+      for (const depElem of value.split(',')) {
+        for (let name of depElem.split('|')) {
+          name = name.trim().split(' ')[0];
```
orkamara: same as above

lib/analyzer/apt-analyzer.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+function parseExtLine(text, curPkgName, pkgMap) {
+  const [key, value] = text.split(': ');
+  switch (key) {
+    case 'Package':
+      curPkgName = value;
+      break;
+    case 'Auto-Installed':
+      if (parseInt(value, 10) === 1) {
```
orkamara: any possible exception on failure?

lib/analyzer/index.ts
```diff
@@ -0,0 +1,26 @@
...
+export {
+  analyze,
+};
+
+function analyze(targetImage) {
+  return Promise.all([
+    imageIdDetector.detect(targetImage),
+    osReleaseDetector.detect(targetImage),
+    Promise.all([
+      apkAnalyzer.analyze(targetImage),
```
orkamara: what's happening on an error?

lib/analyzer/index.ts
```diff
@@ -0,0 +1,26 @@
...
+      aptAnalyzer.analyze(targetImage),
+      rpmAnalyzer.analyze(targetImage),
+    ]),
+  ])
+  .then(res => ({
+    imageId: res[0],
+    osRelease: res[1],
+    results: res[2],
```
orkamara: in the GO code, on a happy flow, the size of the results arr is only 1 (I'm pretty sure), but now it will always be 3. Is it OK with the output later on?

orkamara:
how do we gonna benchmark against the analyzer?

lib/analyzer/os-release-detector.ts
```diff
@@ -0,0 +1,139 @@
+import subProcess from '../sub-process';
+
+export {
+  detect,
+};
+
+async function dockerCat(targetImage, file): Promise<string> {
```
orkamara: code dup

lib/analyzer/os-release-detector.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  if (!osRelease) {
+    osRelease = await tryRedHatRelease(targetImage);
+  }
+
+  if (!osRelease) {
+    throw new Error('Failed to detect OS release');
+  }
+
+  // Oracle Linux identifies itself as "ol"
```
orkamara: maybe enter it to tryOracleRelease()?

lib/analyzer/os-release-detector.ts
```diff
@@ -0,0 +1,139 @@
...
+async function tryOSRelease(targetImage) {
+  const text = await dockerCat(targetImage, '/etc/os-release');
+  if (!text) {
+    return null;
+  }
+  const idRes = text.match(/^ID=(.+)$/m);
+  if (!idRes) {
+    throw new Error('Failed to parse /etc/os-release');
+  }
+  const name = idRes[1].replace(/"/g, '');
```
orkamara: maybe extract ".replace(/"/g, '');" to a function - you are using it multiple times

lib/analyzer/os-release-detector.ts
```diff
@@ -0,0 +1,139 @@
...
+    return null;
+  }
+  text = text.trim();
+  if (text.length < 2) {
+    throw new Error('Failed to parse /etc/alpine-release');
+  }
+  return { name: 'alpine', version: text };
+}
+
+async function tryRedHatRelease(targetImage) {
```
orkamara: tryRedHatRelease() and tryOracleRelease() are totally identical..combine into the same func pls

lib/analyzer/index.ts
darscan: > in the GO code, on a happy flow, the size of the results arr is only 1 (I'm pretty sure)

Nope :)

> Is it OK with the output later on?

Yup, because we already deal with it

lib/analyzer/os-release-detector.ts
darscan: Nope :) 2 reasons:
- The wrong abstraction is worse than some code duplication
- `dry` is brittle

lib/analyzer/apk-analyzer.ts
darscan: Not sure what `flip the if` means

lib/analyzer/apk-analyzer.ts
darscan: Nope, we didn't. The property is here to keep the type consistent

darscan: I was keeping it in line with the GO code

lib/analyzer/os-release-detector.ts
darscan: I'd rather leave them inline so that the code can be read easily

lib/analyzer/apk-analyzer.ts
darscan: happy with it as is :)

lib/analyzer/apk-analyzer.ts
darscan: I'm ok with the current style, indentation and re-use of name

PR #44, opened 2018-11-06
Parse image name with registry when a port is specified

Limess:
- [x] Ready for review

#### What does this PR do?

If using a registry with a specific port. i.e. the default of 5000, then the current parsing results in
```json5
{
  name: `docker-image|{registryName},
  version: `${registryPort}:${imageName}:${imageTag},
  ...
}
```

Instead we parse out any forward slashes and then parse the registry explicitly if it exists, resulting in:

```
docker-image|registryName/imageName
```

e.g

```
docker-image|localhost:5000/redis
```

`system.test.js` to view the expectation.

Likely unnecessary here as the existing tests pass.

We use a private internal registry with a specific port, `nexus.in.ft.com:5000`.

Not raised.

#### Screenshots

n/a

#### Additional questions

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -133,10 +133,19 @@ function parseAnalysisResults(analysisJson) {
   };
 }
 
+function parseImageName(imageName) {
+  const slashSplit = imageName.split('/');
+  const colonSplit = imageName.split(':');
+
+  const name = slashSplit[1]
+      ? `${slashSplit[0]}/${slashSplit[1]}`
+      : imageName.split(':')[0];
+  const version = colonSplit[colonSplit.length - 1];
+  return [name, version];
+}
+
 function buildTree(targetImage, depType, depInfosList, targetOS) {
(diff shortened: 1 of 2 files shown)


robcresswell:
Awesome! Really appreciate this. I've assigned myself as a reviewer; please give me a couple of days to take a detailed look.

robcresswell:
Hey @Limess! I took a look and believe I've identified a small gap in the parsing logic, where names had multiple `'/'` in them. However, I didn't want to waste your time going back and forth on the PR so I pushed a separate one (#45) and credited you in both the PR and the commit message. Really appreciate the bug catch and the work that went into the fix here; made it much easier for me to pick up and investigate. If my approach seems okay, please close this when you're ready :smile:

Limess:
No problem, this was a PR entirely to fix a specific issue in the most straightforwards path for our use case -- I didn't delve into any specs. Very happy you found a more robust way of checking!

PR #47, opened 2018-11-21
fix: dockerfile parsing when using ARG with FROM

robcresswell:
We were not previously taking ARG into account when parsing FROM
instructions; this patch pulls in an external lib to parse dockerfiles

- [x] Ready for review

### DIFF
--- a/lib/docker-file.ts
+++ b/lib/docker-file.ts
@@ -1,4 +1,5 @@
 import * as fs from 'fs';
+import { DockerfileParser } from 'dockerfile-ast';
 
 export { getBaseImageName };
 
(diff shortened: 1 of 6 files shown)


lib/docker-file.ts
```diff
@@ -1,24 +1,45 @@
 import * as fs from 'fs';
+import { DockerfileParser, Dockerfile, From } from 'dockerfile-ast';
 
-export { getBaseImageName };
+export { getDockerfileDetails };
 
-async function getBaseImageName(targetFile?: string):
-  Promise<string|undefined> {
+interface DockerfileDetails {
```
orkamara: Why do we need interface now? and not using regular string as before?

lib/docker-file.ts
```diff
@@ -1,24 +1,45 @@
 import * as fs from 'fs';
+import { DockerfileParser, Dockerfile, From } from 'dockerfile-ast';
```
orkamara:
```suggestion
import { DockerfileParser } from 'dockerfile-ast';
```

lib/docker-file.ts
```diff
@@ -10,15 +11,31 @@ async function getBaseImageName(targetFile?: string):
   }
 
   const contents = await readFile(targetFile);
+  const dockerfile = DockerfileParser.parse(contents);
+  const from = dockerfile.getFROMs().pop();
```
orkamara: Consider rename `from`. Maybe something like `fromContext`? :thinking:

lib/docker-file.ts
```diff
@@ -10,15 +11,31 @@ async function getBaseImageName(targetFile?: string):
...
+  let baseImage;
 
-  const results: string[] = [];
-  const FROM_RE = /^FROM ([^\s]+)[\s\S]*?$/gim;
-  let match: RegExpExecArray|null;
-  while (match = FROM_RE.exec(contents)) {
-    results.push(match[1]);
+  if (from) {
```
orkamara:
```suggestion
  if (!from) {
  return {};
  }
```

lib/docker-file.ts
```diff
@@ -10,15 +11,31 @@ async function getBaseImageName(targetFile?: string):
...
+    if (fromVariables) {
+      const resolvedVariables = fromVariables.reduce(
+        (resolvedVars, variable) => {
+          const line = variable.getRange().start.line;
+          const name = variable.getName();
+          resolvedVars[name] = dockerfile.resolveVariable(name, line);
+          return resolvedVars;
+        }, {});
+
+      Object.keys(resolvedVariables).forEach((variable) => {
```
orkamara: should we check that the size of resolvedVariables == 2 \ 0?

lib/docker-file.ts
robcresswell: It'll need to return undefined, rather than an empty object, but sure, can reorder the code

lib/docker-file.ts
robcresswell: This is literally a string that is the final "FROM" line in a Dockerfile. I don't think it needs renaming, and I don't know what "context" is being added here.

PR #51, opened 2018-12-18
feat: identify vulns introduced by the dockerfile

robcresswell:


package.json
```diff
@@ -14,22 +14,25 @@
...
     "prepare": "npm run build"
   },
   "dependencies": {
+    "@types/node": "^8",
     "debug": "^3",
     "dockerfile-ast": "0.0.12",
     "semver": "^5.6.0",
-    "tslib": "^1"
+    "tslib": "^1",
+    "typescript": "^3"
```
darscan: Why have `typescript` and `@types/node` become prod deps?

lib/index.ts
```diff
@@ -23,16 +26,55 @@ function inspect(root: string, targetFile?: string, options?: any) {
...
+
       return {
         plugin: metadata,
         package: pkg,
       };
     });
 }
 
+function getDockerfileDependencies(dockerfilePackages: DockerFilePackages,
+                                   dependencies) {
```
darmalovan: Can you add a type annotation to the return value of the function?

lib/index.ts
```diff
@@ -23,16 +26,55 @@ function inspect(root: string, targetFile?: string, options?: any) {
...
+      };
+
       return {
         plugin: metadata,
         package: pkg,
       };
     });
 }
 
+function getDockerfileDependencies(dockerfilePackages: DockerFilePackages,
```
darmalovan: This function is more of `adding` the docker file package dependencies if I'm understanding it correctly.
Maybe `addDockerfilePackagesDependencies` ?

Maybe a small comment here explaining why we do this would be nice :)

test/lib/instruction-parser.spec.ts
```diff
@@ -0,0 +1,106 @@
+import { getPackagesFromRunInstructions } from '../../lib/instruction-parser';
+
+describe('instruction parsers', () => {
+  describe('getPackagesFromRunInstructions', () => {
```
darmalovan: Can we add a test for empty instructions?

test/lib/instruction-parser.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+      });
+    });
+
+    test('handles multiple spaces', () => {
+      const instruction = 'RUN    apt   install   curl   vim   ';
+      const packages = getPackagesFromRunInstructions([instruction]);
+
+      expect(packages).toEqual({
+        'curl': { instruction },
+        'vim': { instruction },
```
darmalovan: VIM 💪

test/lib/instruction-parser.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+      const instruction = 'RUN apt install curl && apt install wget';
+      const packages = getPackagesFromRunInstructions([instruction]);
+
+      expect(packages).toEqual({
+        'curl': { instruction },
+        'wget': { instruction },
+      });
+    });
+
+    test('supports multiple commands using ";"', () => {
```
darmalovan: Nice ! can we add another test for this instruction:
`const instruction = 'RUN apt install curl; echo "rob"';`

darmalovan:
Looks good! some not big comments

darmalovan:
One thing I'm not sure about is the fact that a vulnerability can appear in multiple packages.
The difference will be the vulnerable path.
Our docker instruction output is per vulnerability and not per vulnerable path how it will work out?

For example the same vuln in `curl` and `wget`, let's `curl` is installed by the user - what will the CLI output will look like?

lib/instruction-parser.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * We also need to account for the multiple ways to split commands, and
+ * arbitrary whitespace
+ */
+function getPackagesFromRunInstructions(
+  runInstructions: string[],
+): DockerFilePackages {
+  return runInstructions.reduce((dockerfilePackages, instruction) => {
+    // A single RUN may include multiple commands, so remove "RUN " and split
+    const commands = instruction.slice(4).split(/\s?(;|&&)\s?/);
```
orkamara: what's 4? the len of 'RUN '?
Maybe create a const and use length?

lib/instruction-parser.ts
```diff
@@ -0,0 +1,45 @@
+export { getPackagesFromRunInstructions, DockerFilePackages };
+
+interface DockerFilePackages {
+  [packageName: string]: {
+    instruction: string;
+  };
+}
+
+// tslint:disable-next-line:max-line-length
+const installRegex = /\s*(rpm\s+-i|rpm\s+--install|apk\s+add|apt\s+install|apt-get\s+install)\s+(-\w\s+)?/;
```
orkamara: Maybe we should also support more common ones? like `yum install` and `aptitude install`?

test/lib/instruction-parser.spec.ts
```diff
@@ -0,0 +1,106 @@
+import { getPackagesFromRunInstructions } from '../../lib/instruction-parser';
+
+describe('instruction parsers', () => {
```
orkamara: Can we add another test for an install command spitted by multiple line?
`RUN apt-get update && apt-get install -y \
    python-qt4 \
    python3-pyqt5`

test/lib/instruction-parser.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    });
+
+    test('supports "rpm --install"', () => {
+      const instruction = 'RUN rpm --install curl';
+      const packages = getPackagesFromRunInstructions([instruction]);
+
+      expect(packages).toEqual({ 'curl': { instruction } });
+    });
+
+    test('ignores irrelevant flags', () => {
```
orkamara: can we add a test where the flags are in the end?
`apt-get install vim -y`

lib/index.ts
```diff
@@ -23,16 +26,55 @@ function inspect(root: string, targetFile?: string, options?: any) {
...
+                                   dependencies) {
+  for (const dependencyName in dependencies) {
+    if (dependencies.hasOwnProperty(dependencyName)) {
+      const sourceOrName = dependencyName.split('/')[0];
+      const dockerfilePackage = dockerfilePackages[sourceOrName];
+
+      if (dockerfilePackage) {
+        collectDeps(dependencies[dependencyName]).forEach((dep) => {
```
darmalovan: Let's say package `a` is installed by the base image, and package `b` is installed by the docker file.
As I understand the code, it means that if a package `b` is depending on `a` it will mark `a` as a docker file package - which will mark a vuln on this package as a docker file vuln. LMK what do you think.

lib/instruction-parser.ts
robcresswell: Good point, will do

test/lib/instruction-parser.spec.ts
robcresswell: Ah good suggestion! I don't think my code catches this.

lib/index.ts
robcresswell: Yup, it will do exactly this. Any suggestions on improving it?

lib/index.ts
robcresswell: Done

test/lib/instruction-parser.spec.ts
robcresswell: YAAAASSSSS

test/lib/instruction-parser.spec.ts
robcresswell: Updated the existing case to 

```
const instruction = 'RUN apt install curl; apt install vim; echo "bitwise"';
```

Hope that covers the case you were concerned about

lib/instruction-parser.ts
robcresswell: Done

test/lib/instruction-parser.spec.ts
robcresswell: Good catch

lib/index.ts
```diff
@@ -23,16 +26,62 @@ function inspect(root: string, targetFile?: string, options?: any) {
...
+
       return {
         plugin: metadata,
         package: pkg,
       };
     });
 }
 
+// Iterate over the dependencies list; if one is introduced by the dockerfile,
+// flatten it's dependencies and append them to the list of dockerfile
```
karniwl: sorry for nitpicking but `it's` should be `its` :)

robcresswell: Fixed

darmalovan:
Approved, Just take a look at the test failures?

PR #54, opened 2019-01-07
feat: don't fetch node binaries if they are installed by package manager

darmalovan:
- [ ] Ready for review

#### What does this PR do?

We want to remove duplication of vulnerabilities for the case
that node installed by a package manager.

### DIFF
--- a/lib/analyzer/binaries-analyzer.ts
+++ b/lib/analyzer/binaries-analyzer.ts
@@ -7,8 +7,8 @@ export {
   analyze,
 };
 
-async function analyze(targetImage: string) {
-  const binaries = await getBinaries(targetImage);
+async function analyze(targetImage: string, installedPackages: string[]) {
+  const binaries = await getBinaries(targetImage, installedPackages);
   return {
     Image: targetImage,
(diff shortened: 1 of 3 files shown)


lib/analyzer/binaries-analyzer.ts
```diff
@@ -7,15 +7,23 @@ export {
...
   return {
     Image: targetImage,
     AnalyzeType: 'binaries',
     Analysis: binaries,
   };
 }
 
+function installedByPackageManager(installedPackages: string[]): boolean {
+  return installedPackages
+    .filter((pkg) => pkg === 'node' || pkg === 'nodejs').length > 0;
```
robcresswell: Is there a way we can compare the results of the binary and system results so we don't need to keep manually updating this function?

Something like, get the binaries, then filter anything that is in the OS deps list

darmalovan: Let me work it out...

robcresswell: :crown:

darmalovan:
@robcresswell LMK if this is what you thought about

PR #58, opened 2019-01-21
feat: check openjdk installed by pkg manager

karniwl:
- [x] Ready for review

#### What does this PR do?
Adds package names for `openjdk-jre` according to package manager `APK`, `APT`, `RPM`.
Refactors logic by insertion of `installedByPackageManager` function into each binary `extractor` as it needs to be unique per binary.

snyk docker test any node image.

### DIFF
--- a/lib/analyzer/binary-version-extractors/openjdk-jre.ts
+++ b/lib/analyzer/binary-version-extractors/openjdk-jre.ts
@@ -3,11 +3,8 @@ import { Binary } from '../types';
 
 export {
   extract,
-  packageNames,
+  installedByPackageManager,
 };
-
-// todo: add common names such as 'java-common'
-const packageNames = ['java'];
 
 async function extract(targetImage: string): Promise<Binary | null> {
   try {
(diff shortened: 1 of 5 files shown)


lib/analyzer/index.ts
```diff
@@ -48,14 +49,16 @@ async function analyze(targetImage: string) {
...
   });
 
   if (!dockerAnalysis) {
-    return [];
+    return { installedPackages: [] };
   }
-  return dockerAnalysis.Analysis.map((pkg) => pkg.Name);
-
+  const installedPackages = dockerAnalysis.Analysis.map((pkg) => pkg.Name);
+  const pkgManager = dockerAnalysis.AnalyzeType;
```
darmalovan: I would lowercase `pkgManager`, if `AnalyzeType` is not null

lib/analyzer/binary-version-extractors/openjdk-jre.ts
```diff
@@ -46,3 +43,31 @@ function parseOpenJDKBinary(fullVersionOutput: string) {
     version,
   };
 }
+
+const pkgManagerPkgNames = {
```
darmalovan: Maybe change this name to `javaPkgRegexByPkgManager` ?

lib/analyzer/binary-version-extractors/openjdk-jre.ts
```diff
@@ -46,3 +43,31 @@ function parseOpenJDKBinary(fullVersionOutput: string) {
...
+  Apt: [/openjdk-\d*-jre-headless/,
```
darmalovan: Maybe give an example for every regex to which package it conforms in a comment?

PR #60, opened 2019-02-08
fix: subprocess: include both stdout and stderr

leongold:
Process may emit both stdout and stderr.

We currently only propagate one (giving precedence to stdout), potentially causing loss of data, e.g.:

```
Leons-MBP:snyk-docker-plugin leongold$ docker inspect foo:123
[] <--- stdout
Error: No such object: foo:123 <--- stderr
```

Refactor the project to return both in a single object, i.e {stdout, stderr}

- [x] Ready for review

### DIFF
--- a/lib/analyzer/apk-analyzer.ts
+++ b/lib/analyzer/apk-analyzer.ts
@@ -1,6 +1,6 @@
 import { Docker } from '../docker';
 import { AnalyzerPkg } from './types';
-
+import { Output } from '../sub-process';
 export {
   analyze,
 };
@@ -20,10 +20,10 @@ function getPackages(targetImage: string) {
     .then(parseFile);
 }
 
-function parseFile(text: string) {
+function parseFile(output: Output) {
   const pkgs: AnalyzerPkg[] = [];
   let curPkg: any = null;
-  for (const line of text.split('\n')) {
+  for (const line of output.stdout.split('\n')) {
     curPkg = parseLine(line, curPkg, pkgs);
   }
   return pkgs;
--- a/lib/analyzer/apk-analyzer.ts
+++ b/lib/analyzer/apk-analyzer.ts
@@ -1,6 +1,6 @@
 import { Docker } from '../docker';
 import { AnalyzerPkg } from './types';
-import { Output } from '../sub-process';
+
 export {
   analyze,
 };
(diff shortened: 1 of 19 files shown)


michael-go:
I suggest something a-bit different:
return `{stdout, stderr}` and let the caller decide what to use

lib/analyzer/apk-analyzer.ts
```diff
@@ -20,10 +20,10 @@ function getPackages(targetImage: string) {
     .then(parseFile);
 }
 
-function parseFile(text: string) {
+function parseFile(output: Output) {
```
michael-go: don't think this function should get `Output` - just give it the text like before.

lib/analyzer/binary-version-extractors/openjdk-jre.ts
```diff
@@ -7,16 +7,18 @@ export {
...
+  let binaryVersion: string;
   try {
-    const binaryVersion = await new Docker(targetImage).
-      run('java', [ '-version' ]);
-    return parseOpenJDKBinary(binaryVersion);
-  } catch (stderr) {
+    binaryVersion = (await new Docker(targetImage).
+      run('java', [ '-version' ])).stdout;
+  } catch (error) {
+    const stderr: string = error.stderr;
```
michael-go: the ` :string` annotation is not really needed here

lib/analyzer/binary-version-extractors/openjdk-jre.ts
```diff
@@ -7,16 +7,18 @@ export {
...
     if (typeof stderr === 'string' && stderr.indexOf('not found') >= 0) {
       return null;
     }
     throw new Error(stderr);
   }
+  return parseOpenJDKBinary(binaryVersion);
```
michael-go: why did you decided to take this out of the `try {}` block?

lib/analyzer/rpm-analyzer.ts
```diff
@@ -22,18 +23,19 @@ function getPackages(targetImage: string) {
...
+        return { stdout: '', stderr: ''};
       }
-      throw new Error(stderr);
+      throw error;
     })
     .then(parseOutput);
 }
 
-function parseOutput(text: string) {
+function parseOutput(output: Output) {
```
michael-go: if this function will just get the `stdout` string, you won't need to import the `Output` interface here.

lib/sub-process.ts
```diff
@@ -1,10 +1,16 @@
 import * as childProcess from 'child_process';
 
-export function execute(
+export { execute, Output };
```
michael-go: please add a blank line here

lib/sub-process.ts
```diff
@@ -1,10 +1,16 @@
...
+interface Output {
```
michael-go: Consider renaming to something for explicit, like `CmdOutput` - as in other places where you do:
```js
import {Output} from './sub-process'
```
the `Output` type can be confusing as it's a very generic name

lib/analyzer/binary-version-extractors/openjdk-jre.ts
leongold: The try block is intended for the subprocess call, but let's leave it out as it's an unrelated change

PR #61, opened 2019-02-08
feat: support global docker options

leongold:
- [x] Ready for review

Added support for global docker options:
host (docker server: hostname:port)
tlsverify (verify tls: true/false)
tlskey (path to tls key)
tlscert (path to tls cert)
tlscacert (path to tls ca cert)

(snyk --docker --tlsverify=true --host=localhost:1234...)

entrypoint options are now being propagated all the way
through the docker module.

A couple of additional changes that had to take place:

1) We used to assume that if there's an inspection error,
it is due to missing images. This is no longer the case, and
this commit adds stderr output to the error messages.

2) the inspection module was using subprocess directly,
and it now uses the docker module wrapper in order to
also make use of the propagated options.

docker.ts

run docker's subprocess with some options

https://github.com/snyk/snyk/issues/299

### DIFF
--- a/lib/analyzer/binaries-analyzer.ts
+++ b/lib/analyzer/binaries-analyzer.ts
@@ -1,4 +1,5 @@
 import { Binary } from './types';
+import { DockerOptions } from '../docker';
 
 export {
   analyze,
@@ -7,10 +8,10 @@ export {
 async function analyze(
   targetImage: string,
   installedPackages: string[],
-  options?: any,
-  pkgManager?: string) {
+  pkgManager?: string,
+  options?: DockerOptions) {
   const binaries = await getBinaries(
-    targetImage, installedPackages, options, pkgManager);
+    targetImage, installedPackages, pkgManager, options);
   return {
     Image: targetImage,
     AnalyzeType: 'binaries',
@@ -26,14 +27,14 @@ const binaryVersionExtractors = {
 async function getBinaries(
   targetImage: string,
   installedPackages: string[],
-  options?: any,
   pkgManager?: string,
+  options?: DockerOptions,
   ): Promise<Binary[]> {
   const binaries: Binary[] = [];
   for (const versionExtractor of Object.keys(binaryVersionExtractors)) {
     const extractor = binaryVersionExtractors[versionExtractor];
     if (extractor.installedByPackageManager(
-      installedPackages, options, pkgManager)) {
+      installedPackages, pkgManager, options)) {
       continue;
     }
     const binary = await extractor.extract(targetImage);
--- a/lib/analyzer/binaries-analyzer.ts
+++ b/lib/analyzer/binaries-analyzer.ts
@@ -7,9 +7,10 @@ export {
 async function analyze(
   targetImage: string,
   installedPackages: string[],
+  options?: any,
   pkgManager?: string) {
   const binaries = await getBinaries(
-    targetImage, installedPackages, pkgManager);
+    targetImage, installedPackages, options, pkgManager);
   return {
     Image: targetImage,
     AnalyzeType: 'binaries',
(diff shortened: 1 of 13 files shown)


lib/docker.ts
```diff
@@ -24,4 +37,27 @@ class Docker {
       throw error;
     }
   }
+
+  private createOptionsList(options: any) {
```
orkamara:
```suggestion
  private createOptionsList(options?: any) {
```

lib/docker.ts
```diff
@@ -3,16 +3,29 @@ import * as subProcess from './sub-process';
 export { Docker };
 
 class Docker {
-  constructor(private targetImage: string) {
+
+  private optionsList: string[];
+
+  constructor(
+    private targetImage: string,
+    options?: any,
```
orkamara: maybe create a type to options?

lib/analyzer/index.ts
```diff
@@ -10,21 +10,21 @@ export {
...
-      apkAnalyzer.analyze(targetImage),
-      aptAnalyzer.analyze(targetImage),
-      rpmAnalyzer.analyze(targetImage),
+      apkAnalyzer.analyze(targetImage, options),
+      aptAnalyzer.analyze(targetImage, options),
+      rpmAnalyzer.analyze(targetImage, options),
     ]).catch((err) => {
       debug(`Error while running analyzer: '${err}'`);
-      throw new Error('Failed to detect installed OS packages');
+      throw new Error(`Failed to detect installed OS packages: '${err}'`);
```
orkamara: Let's not output the customer all of the ugly errors :) - to keep it more user friendly
This is why there is a debug right before..

lib/analyzer/binary-version-extractors/openjdk-jre.ts
```diff
@@ -6,10 +6,12 @@ export {
...
 
-async function extract(targetImage: string): Promise<Binary | null> {
+async function extract(
+  targetImage: string,
+  options?: any): Promise<Binary | null> {
   try {
-    const binaryVersion = (await new Docker(targetImage).
-      run('java', [ '-version' ])).stdout;
+    const binaryVersion = (await new Docker(
+      targetImage, options).run('java', [ '-version' ])).stdout;
```
orkamara:
```suggestion
 = (await new Docker(targetImage,options)
 .run('java', [ '-version' ])).stdout;
```

lib/analyzer/binaries-analyzer.ts
```diff
@@ -7,9 +7,10 @@ export {
 async function analyze(
   targetImage: string,
   installedPackages: string[],
+  options?: any,
```
orkamara: add `options` as the last arg

lib/analyzer/binaries-analyzer.ts
```diff
@@ -23,12 +24,16 @@ const binaryVersionExtractors = {
 };
 
 async function getBinaries(
-  targetImage: string, installedPackages: string[], pkgManager?: string)
-  : Promise<Binary[]> {
+  targetImage: string,
+  installedPackages: string[],
+  options?: any,
```
orkamara: add `options` as the last arg

test/system/system.test.ts
```diff
@@ -11,9 +11,36 @@ import * as subProcess from '../../lib/sub-process';
...
+        err.message,
+        'no such host');
+    });
+});
+
 test('inspect an image that does not exist', t => {
   return plugin.inspect('not-here:latest').catch((err) => {
-    t.same(err.message, 'Docker image was not found locally: not-here:latest');
+    t.includes(
+      err.message, 'No such object: not-here:latest');
```
orkamara: Why the error is `No such object`? it's not user friendly - please keep it as before (`Docker image was not found locally`)

Maybe we can have 2 different types of error - one for `Docker image was not found` and another one for global errors (so more advanced users can handle their problems?

lib/docker.ts
leongold: Decided to keep it non-optional

lib/docker.ts
```diff
@@ -1,18 +1,39 @@
 import * as subProcess from './sub-process';
 
-export { Docker };
+export { Docker, DockerOptions};
+
+interface DockerOptions {
```
orkamara: :medal_military:

PR #67, opened 2019-02-11
fix: improve java binary analysis

leongold:
Current JRE parsing logic is too restrictive:

1) Expects RE version to appear strictly in the 2nd line --
I'm now looking for "Runtime Environment" to determine the line.
This is merely a better heuristic, but it does fix adoptOpenJDK
detection.

2) Expects the output to be exactly 3 lines long.
I don't see the necessity.

- [x] Ready for review

### DIFF
--- a/lib/analyzer/binary-version-extractors/openjdk-jre.ts
+++ b/lib/analyzer/binary-version-extractors/openjdk-jre.ts
@@ -30,15 +30,19 @@ function parseOpenJDKBinary(fullVersionOutput: string) {
   */
   const jdkVersionLines = fullVersionOutput &&
                           fullVersionOutput.trim().split('\n');
-  if (!jdkVersionLines || jdkVersionLines.length !== 3) {
+  if (!jdkVersionLines) {
     return null;
   }
   const bracketsRE = /\(build (.*)\)$/;
-  const buildVersion = jdkVersionLines[1].match(bracketsRE);
-  const version = buildVersion && buildVersion[1];
+  const runtimeEnv = 'Runtime Environment';
+  const buildVersion =
+    (jdkVersionLines.filter(line => line.includes(runtimeEnv)).pop() || '')
+    .match(bracketsRE);
+  let version = buildVersion && buildVersion[1];
   if (!version) {
     return null;
   }
+  version = version.replace('-adoptopenjdk', '');
   return {
     name: 'openjdk-jre',
     version,
--- a/lib/analyzer/binary-version-extractors/openjdk-jre.ts
+++ b/lib/analyzer/binary-version-extractors/openjdk-jre.ts
@@ -39,11 +39,10 @@ function parseOpenJDKBinary(fullVersionOutput: string) {
 
   const bracketsRE = /\(build (.*)\)$/;
   const buildVersion = runtimeLine.match(bracketsRE);
-  let version = buildVersion && buildVersion[1];
+  const version = buildVersion && buildVersion[1];
   if (!version) {
     return null;
   }
-  version = version.replace('-adoptopenjdk', '');
   return {
     name: 'openjdk-jre',
     version,
--- a/lib/analyzer/binary-version-extractors/openjdk-jre.ts
+++ b/lib/analyzer/binary-version-extractors/openjdk-jre.ts
@@ -28,16 +28,17 @@ function parseOpenJDKBinary(fullVersionOutput: string) {
    Java HotSpot(TM) 64-Bit Server VM (build 25.191-b12, mixed mode)`
   => extracting `1.8.0_191-b12`
   */
-  const jdkVersionLines = fullVersionOutput &&
-                          fullVersionOutput.trim().split('\n');
-  if (!jdkVersionLines) {
+  const runtimeEnv = 'Runtime Environment';
+  const runtimeLine = fullVersionOutput &&
+                      fullVersionOutput.trim()
+                      .split('\n')
+                      .find(line => line.includes(runtimeEnv));
+  if (!runtimeLine) {
     return null;
   }
+
   const bracketsRE = /\(build (.*)\)$/;
-  const runtimeEnv = 'Runtime Environment';
-  const buildVersion =
-    (jdkVersionLines.filter(line => line.includes(runtimeEnv)).pop() || '')
-    .match(bracketsRE);
+  const buildVersion = runtimeLine.match(bracketsRE);
   let version = buildVersion && buildVersion[1];
   if (!version) {
     return null;
--- a/test/lib/analyzer/binaries-analyzer.test.ts
+++ b/test/lib/analyzer/binaries-analyzer.test.ts
@@ -109,6 +109,32 @@ test('analyze', async t => {
         { name: 'openjdk-jre', version: '1.8.0_191-b12' },
       ],
     },
+    {
+      description: 'adoptopenjdk version with hyphen in image',
+      targetImage: 'adoptopenjdk/openjdk10',
+      binariesOutputLines: { node: '',
+                             openjdk: `Picked up JAVA_TOOL_OPTIONS: -XX:+UseContainerSupport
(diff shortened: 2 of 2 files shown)


test/lib/analyzer/binaries-analyzer.test.ts
```diff
@@ -109,6 +109,19 @@ test('analyze', async t => {
...
+      description: 'adoptOpenJDK present in image',
+      targetImage: 'adoptopenjdk/openjdk10',
+      binariesOutputLines: { node: '',
+                             openjdk: `Picked up JAVA_TOOL_OPTIONS: -XX:+UseContainerSupport
+                                       openjdk 10.0.2-adoptopenjdk 2018-07-17
+                                       OpenJDK Runtime Environment (build 10.0.2-adoptopenjdk+13)
+                                       OpenJDK 64-Bit Server VM (build 10.0.2-adoptopenjdk+13, mixed mode)` },
+      installedPackages: [],
+      expectedBinaries: [
+        { name: 'openjdk-jre', version: '10.0.2-adoptopenjdk+13' },
```
orkamara: the expected version should be `10.0.2`

test/lib/analyzer/binaries-analyzer.test.ts
```diff
@@ -109,6 +109,19 @@ test('analyze', async t => {
         { name: 'openjdk-jre', version: '1.8.0_191-b12' },
       ],
     },
+    {
+      description: 'adoptOpenJDK present in image',
```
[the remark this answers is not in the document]
leongold: I think this is bad testing practice. Essentially, this means we're going to multiply a single test ("test if we are able to detect and analyse adoptopenjdk") case many times.

If we decide to go that route, in order to keep testing integrity, we'll need to cover all tags/versions (and not for just adoptopenjdk, but for openjdk as well)... and we really shouldn't IMO.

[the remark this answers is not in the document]
leongold: BTW, I did manually check all versions (8-12) pass this test for the customer's sake :)

test/lib/analyzer/binaries-analyzer.test.ts
leongold: I understand that it still qualifies as plain opendjk and versions should ideally match, but even opendjk 10.0.2 returns a more elaborate version:

Leons-MacBook-Pro:snyk-docker-plugin leongold$ docker run -it openjdk:10.0.2 /bin/sh
# java --version
openjdk 10.0.2 2018-07-17
OpenJDK Runtime Environment (build 10.0.2+13-Debian-2)
OpenJDK 64-Bit Server VM (build 10.0.2+13-Debian-2, mixed mode)

Looking at other fixtures, we don't confine versions to strict x.y.z either, e.g. "1.8.0_191-b12"

lib/analyzer/binary-version-extractors/openjdk-jre.ts
```diff
@@ -30,11 +30,14 @@ function parseOpenJDKBinary(fullVersionOutput: string) {
   */
   const jdkVersionLines = fullVersionOutput &&
                           fullVersionOutput.trim().split('\n');
-  if (!jdkVersionLines || jdkVersionLines.length !== 3) {
```
karniwl: @leongold not very critical, but removing this condition doesn't have anything to do with adopt openjdk, right? we purposely added this to be restrictive after checking many cases of `java -version` outputs on different images. @darmalovan

karniwl:
@leongold perhaps you should add to your commit message what you're improving, which imo isn't an improvement nor a `fix` but a `feat` because you're adding support to Adopt OpenJDK

lib/analyzer/binary-version-extractors/openjdk-jre.ts
leongold: This PR fixes adopt openjdk (I hope :D) but it's more of a happy side-effect in this context.

The main purpose is to alleviate the maybe too restrictive parsing and, in the hopes that other flavours of openjdk will be supported out of the box.

We have just 1 use case of it now, but perhaps when we encounter more cases we'll find out this solution isn't quite right either...

I think that all in all parsing strings can't actually be a fail proof method, but as long as this is what we do, we can only tweak restrictions when we encounter new use cases...

leongold:
> @leongold perhaps you should add to your commit message what you're improving, which imo isn't an improvement nor a `fix` but a `feat` because you're adding support to Adopt OpenJDK

I think you're right this isn't a fix, but please see my comment to L33 about what this PR is about IMO

lib/analyzer/binary-version-extractors/openjdk-jre.ts
karniwl: I guess it's a point of view if adding support for adopt openjdk is a fix or not.
I agree regards parsing strings but we need to make sure our rules are not too loose...

karniwl: either way just `improve` isn't informative :)

leongold: I agree, I'll improve `improve` 😄

lib/analyzer/binary-version-extractors/openjdk-jre.ts
```diff
@@ -30,15 +30,19 @@ function parseOpenJDKBinary(fullVersionOutput: string) {
...
+  if (!jdkVersionLines) {
     return null;
   }
   const bracketsRE = /\(build (.*)\)$/;
-  const buildVersion = jdkVersionLines[1].match(bracketsRE);
-  const version = buildVersion && buildVersion[1];
+  const runtimeEnv = 'Runtime Environment';
+  const buildVersion =
+    (jdkVersionLines.filter(line => line.includes(runtimeEnv)).pop() || '')
```
darmalovan: I think you can use `.find` instead of `.filter` and remove the `pop`

lib/analyzer/binary-version-extractors/openjdk-jre.ts
```diff
@@ -30,15 +30,19 @@ function parseOpenJDKBinary(fullVersionOutput: string) {
...
+    .match(bracketsRE);
+  let version = buildVersion && buildVersion[1];
   if (!version) {
     return null;
   }
+  version = version.replace('-adoptopenjdk', '');
```
darmalovan: My gut feeling is that we shouldn't remove the original version - but being able to parse it by the binary version translator...

lib/analyzer/binary-version-extractors/openjdk-jre.ts
darmalovan: Maybe also moving this logic to line `32` and let the `if` discard lines without `Runtime environment`.

lib/analyzer/binary-version-extractors/openjdk-jre.ts
leongold: This is a little beyond me, @orkamara wdyt?

lib/analyzer/binary-version-extractors/openjdk-jre.ts
leongold: Nice! I like this much better

lib/analyzer/binary-version-extractors/openjdk-jre.ts
orkamara: Yes. Maybe it's a better one. Cause this way, we'll be able to display the users the actual versions. So they will see it's adopt and not a regular openjdk

orkamara: Sorry @leongold - can you revert it?

test/lib/analyzer/binaries-analyzer.test.ts
leongold: We can resolve this, right @orkamara? (decided to keep `adoptopenjdk` in version here and use the translator)

PR #68, opened 2019-02-12
feat: alter analyse dockerfile by content to support SCM

karniwl:
- [x] Ready for review

#### What does this PR do?
In order to support SCM Docker integration, this PR changes the way the Dockerfile is being parsed and analyzed, the `analyseDockerfile` function currently receives the path of the Dockerfile as an argument, but since in SCM integration the file system isn't available to `snyk-docker-plugin`, we must provide the actual contents of the Dockerfile instead of the path.

### DIFF
--- a/lib/docker-file.ts
+++ b/lib/docker-file.ts
@@ -5,21 +5,26 @@ import {
   DockerFilePackages,
 } from './instruction-parser';
 
-export { analyseDockerfile };
+export { analyseDockerfile, readDockerfileAndAnalyse, DockerFileAnalysis };
 
 interface DockerFileAnalysis {
   baseImage?: string;
   dockerfilePackages: DockerFilePackages;
 }
 
-async function analyseDockerfile(targetFile?: string):
+async function readDockerfileAndAnalyse(targetFilePath?: string):
   Promise<DockerFileAnalysis|undefined> {
 
-  if (!targetFile) {
-    return undefined;
-  }
+if (!targetFilePath) {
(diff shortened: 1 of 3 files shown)


lib/docker-file.ts
```diff
@@ -5,21 +5,26 @@ import {
...
-export { analyseDockerfile };
+export { analyseDockerfile, readDockerfileAndAnalyse, DockerFileAnalysis };
 
 interface DockerFileAnalysis {
   baseImage?: string;
   dockerfilePackages: DockerFilePackages;
 }
 
-async function analyseDockerfile(targetFile?: string):
+async function readDockerfileAndAnalyse(targetFile?: string):
```
orkamara:
```suggestion
async function readDockerfileAndAnalyse(targetFilePath?: string):
```

lib/index.ts
```diff
@@ -1,5 +1,6 @@
 const debug = require('debug')('snyk');
 
+import * as fs from 'fs';
```
orkamara: do we need `fs`?

lib/index.ts
```diff
@@ -11,12 +12,12 @@ export {
   inspect,
 };
 
-function inspect(root: string, targetFile?: string, options?: any) {
```
karniwl: @orkamara I can't remember why we remove `?` from the argument `targetFile`, isn't it an optional argument that is provided only with the `--file` option in the cli?

PR #75, opened 2019-04-07
feat: introducing dockerfileLayers & annotating dep trees with layer ids

leongold:
- [ ] Ready for review

Introducing a structure (`dockerfileLayers`) that maps dockerfile layer digests to dockerfile install instructions. Additionally, the dependency tree is now annotated with dockerfile layer digests (using the new "labels" container (https://github.com/snyk/dep-graph/pull/19)

These two changes/additions will be directly used in displaying layer information for vulnerabilities in snapshots, as well as allowing us to filter base image vulnerabilities in the UI.

### DIFF
--- a/lib/instruction-parser.ts
+++ b/lib/instruction-parser.ts
@@ -1,7 +1,19 @@
-export { getPackagesFromRunInstructions, DockerFilePackages };
+export {
+  getDockerfileLayers,
+  getPackagesFromRunInstructions,
+  DockerFilePackages,
+  DockerFileLayers,
+  instructionDigest,
+};
 
 interface DockerFilePackages {
   [packageName: string]: {
+    instruction: string;
+  };
+}
+
+interface DockerFileLayers {
+  [id: string]: {
     instruction: string;
   };
 }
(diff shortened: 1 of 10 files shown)


lib/instruction-parser.ts
```diff
@@ -1,7 +1,14 @@
-export { getPackagesFromRunInstructions, DockerFilePackages };
+export { getPackagesFromRunInstructions, DockerFilePackages, DockerFileLayers };
 
 interface DockerFilePackages {
   [packageName: string]: {
+    id: string,
```
leongold: Unfortunately this is required to later map packages to layer ids. I would want to replace `instruction` with `id` altogether, but that would be a breaking change that fixing will require a lot of effort that will also slip to other repositories.

lib/response-builder.ts
```diff
@@ -102,3 +107,17 @@ function extractDockerfileDeps(allDeps, dockerfilePkgs) {
       return extractedDeps;
     }, {});
 }
+
+function annotateLayerIds(deps, dockerfilePkgs) {
```
leongold: annotate deptree with the `dockerLayerId` field.

lib/response-builder.ts
```diff
@@ -108,3 +109,23 @@ function extractDockerfileDeps(allDeps, dockerfilePkgs) {
...
```
orkamara: why do we need `annotateLayerIds` and not using `instructionDigest` in `collectDockerfilePkgs`?

Make more sense for me...

leongold: collectDockerfilePkgs populates dependant pkgs for the **dockerfilePackages** map, whereas annoteLayerIds annotates **depTree** with layer ids...

orkamara: was thinking about doing it here https://github.com/snyk/snyk-docker-plugin/blob/master/lib/docker-file.ts#L37

don't you think it will be much more simple?

leongold: not possible since at this stage we don't have all the packages in dockerfilePackages.

one of the functions in response-builder flattens the packages and adds non-direct packages to dockerfilePackages, until that point dockerfilePackages only contain direct packages

PR #77, opened 2019-04-26
fix: Support for scratch images

mladkau:
I did not find a way to detect scratch images without the help of Dockerfiles. Here are my proposed changes to support scratch images with the help of a Dcokerfile. Once we see "FROM scratch" in the dockerfile we basically skip past some tests and only perform the key-binary tests if we don't find a package manager.

### DIFF
--- a/lib/analyzer/index.ts
+++ b/lib/analyzer/index.ts
@@ -1,4 +1,5 @@
 import { DockerOptions } from '../docker';
+import * as dockerFile from '../docker-file';
 import * as osReleaseDetector from './os-release-detector';
 import * as imageInspector from './image-inspector';
 import * as apkAnalyzer from './apk-analyzer';
@@ -11,13 +12,15 @@ export {
   analyze,
 };
 
-async function analyze(targetImage: string, options?: DockerOptions) {
+async function analyze(targetImage: string,
+  dockerfileAnalysis?: dockerFile.DockerFileAnalysis, options?: DockerOptions) {
+
   const [
     imageInspection,
     osRelease,
   ] = await Promise.all([
     imageInspector.detect(targetImage, options),
-    osReleaseDetector.detect(targetImage, options)]);
+    osReleaseDetector.detect(targetImage, dockerfileAnalysis, options)]);
 
   const results = await Promise.all([
     apkAnalyzer.analyze(targetImage, options),
--- a/lib/analyzer/os-release-detector.ts
+++ b/lib/analyzer/os-release-detector.ts
@@ -1,4 +1,5 @@
 import { Docker, DockerOptions } from '../docker';
+import * as dockerFile from '../docker-file';
 import { OSRelease } from './types';
 import { TextDecoder } from 'util';
 
@@ -6,8 +7,10 @@ export {
   detect,
 };
 
-async function detect(targetImage: string, options?: DockerOptions):
-  Promise<OSRelease> {
+async function detect(targetImage: string,
+  dockerfileAnalysis?: dockerFile.DockerFileAnalysis,
+  options?: DockerOptions): Promise<OSRelease> {
+
   const docker = new Docker(targetImage, options);
 
   let osRelease = await tryOSRelease(docker);
@@ -35,7 +38,16 @@ async function detect(targetImage: string, options?: DockerOptions):
   }
 
   if (!osRelease) {
-    throw new Error('Failed to detect OS release');
+    if (dockerfileAnalysis &&
+      dockerfileAnalysis.baseImage === 'scratch') {
+
+      // If the docker file was build from a scratch image
+      // then we don't have a known OS
+
+      osRelease = { name: 'scratch', version: '0.0' };
+    } else {
+      throw new Error('Failed to detect OS release');
+    }
   }
 
   // Oracle Linux identifies itself as "ol"
--- a/lib/docker.ts
+++ b/lib/docker.ts
@@ -68,8 +68,11 @@ class Docker {
       return await this.run('cat', [filename]);
     } catch (error) {
       const stderr: string = error.stderr;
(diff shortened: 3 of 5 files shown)


lib/index.ts
```diff
@@ -7,24 +7,39 @@ import { Docker, DockerOptions } from './docker';
...
 
 export {
+  EnvDetails,
   inspect, dockerFile,
 };
 
+// Additional details which can help the analyzer
+// make a decision
+//
+interface EnvDetails {
```
mladkau: That might be over-engineering it a bit but I think it is better to have an environment object rather then passing the dockerfile details directly around.

lib/index.ts
```diff
@@ -7,24 +7,39 @@ import { Docker, DockerOptions } from './docker';
...
   const targetImage = root;
-  return Promise.all([
-    getRuntime(dockerOptions),
-    getDependencies(targetImage, dockerOptions),
-    dockerFile.readDockerfileAndAnalyse(targetFile),
-  ])
-    .then((res) => buildResponse(res[0], res[1], res[2], options));
+
+  return dockerFile.readDockerfileAndAnalyse(targetFile)
+    .then(dockerFileResult => {
```
mladkau: Here we first resolve the dockerfile promise and then do the runtime and dependencies.

lib/index.ts
```diff
@@ -104,14 +121,29 @@ function getDependencies(targetImage: string, options?: DockerOptions) {
...
-      'Failed to detect a supported Linux package manager (deb/rpm/apk)');
+    if (envDetails.dockerFile &&
+      envDetails.dockerFile.baseImage === 'scratch') {
+
+      // Special case when we have no package management
+      // on scratch images
+
+      analysisResult = {
+        Image: targetImage,
+        AnalyzeType: 'none',
```
mladkau: We have now a new type of package management called "none".

test/lib/analyzer/os-release-detector.test.ts
```diff
@@ -109,6 +109,17 @@ test('os release detection', async t => {
       expected: { name: 'ubuntu', version: '18.04' },
       notes: 'uses /etc/os-release',
     },
+    'scratch': {
+      dir: '',
+      expected: { name: 'scratch', version: '0.0' },
+      notes: 'uses dockerfile',
+      envDetails: {
```
mladkau: I added a test case which tests a scratch image with a dockerfile

test/lib/analyzer/os-release-detector.test.ts
```diff
@@ -117,8 +128,11 @@ test('os release detection', async t => {
     sinon.match.any, 'cat', sinon.match.any,
   ])
     .callsFake(async (docker, [run, rm, entry, empty, network, none, image, cat, file]) => {
+      const example = examples[image];
+      if (example.dir === '') {
```
mladkau: Adding here an extra condition to make sure that all files in case of the scratch image resolve to "not found"

lib/docker.ts
```diff
@@ -68,8 +68,11 @@ class Docker {
       return await this.run('cat', [filename]);
     } catch (error) {
       const stderr: string = error.stderr;
-      if (typeof stderr === 'string' && stderr.indexOf('No such file') >= 0) {
-        return { stdout: '', stderr: '' };
+      if (typeof stderr === 'string') {
+        if (stderr.indexOf('No such file') >= 0 ||
+            stderr.indexOf('file not found') >= 0) {
```
mladkau: I added here the "file not found" otherwise the command line complains about an unknown return from the docker daemon.

lib/analyzer/os-release-detector.ts
```diff
@@ -35,7 +37,16 @@ async function detect(targetImage: string, options?: DockerOptions):
...
 
   if (!osRelease) {
-    throw new Error('Failed to detect OS release');
+    if (envDetails.dockerFile &&
+      envDetails.dockerFile.baseImage === 'scratch') {
+
+      // If the docker file was build from a scratch image
+      // then we don't have a known OS
+
+      osRelease = { name: 'scratch', version: '0.0' };
```
mladkau: Added this as a special case when we have a scratch image with a bare bone environment.

lib/index.ts
leongold: I do feel like this might be over-engineered, but beyond that,  "EnvDetails" does not really explain what am I expecting to find in it. I'd start with the least common denominator rather than jumping to generic terms, if we ever find a more elaborate use to it, we can readdress the name

leongold: For starters, perhaps I would actually pass dockerfile analysis.

lib/index.ts
```diff
@@ -7,24 +7,39 @@ import { Docker, DockerOptions } from './docker';
...
+  dockerFile?: dockerFile.DockerFileAnalysis;
+}
+
 function inspect(root: string, targetFile?: string, options?: any) {
   const dockerOptions = options ? {
     host: options.host,
     tlsverify: options.tlsverify,
     tlscert: options.tlscert,
-    tlscacert: options.tlscacert,
+     tlscacert: options.tlscacert,
```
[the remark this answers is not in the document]
mladkau: I did indeed ... strange that the lint didn't pick this up

lib/analyzer/index.ts
```diff
@@ -11,13 +12,15 @@ export {
...
+async function analyze(targetImage: string,
+  dockerFileResult?: dockerFile.DockerFileAnalysis, options?: DockerOptions) {
+
   const [
     imageInspection,
     osRelease,
   ] = await Promise.all([
     imageInspector.detect(targetImage, options),
-    osReleaseDetector.detect(targetImage, options)]);
+    osReleaseDetector.detect(targetImage, dockerFileResult, options)]);
```
mladkau: We are now passing directly the DockerFileAnalysis around

PR #86, opened 2019-05-28
feat: static package scan for docker images

moshikod:
- [x] Ready for review

#### What does this PR do?

Static scan of packages for docker images

Package repositories for the different OSs

Use known docker image and look for expected installed packages


lib/analyzer/apt-analyzer.ts
```diff
@@ -3,12 +3,25 @@ import { AnalyzerPkg } from "./types";
...
+const APT_EXT_STATES = "/var/lib/apt/extended_states";
+
+const APT_PKGFILES = [APT_DPKG_STATUS, APT_EXT_STATES];
+
+export { APT_PKGFILES };
+
+async function analyze(
+  targetImage: string,
+  options?: DockerOptions,
+  files?: {} | void,
```
robcresswell:
```
files? : { [key: string]: string }
```

is probably more like what you want, depending on the exact format.

lib/analyzer/apt-analyzer.ts
```diff
@@ -3,12 +3,25 @@ import { AnalyzerPkg } from "./types";
...
+) {
+  if (!files) {
+    files = await new Docker(targetImage, options).extract(APT_PKGFILES);
+  }
+  const dpkgFile = files[APT_DPKG_STATUS] || "";
```
robcresswell: What will happen if `parseDpkgFile` receives `""`? I think if `files[APT_DPKG_STATUS]` isn't present, and is needed, we should just return.

lib/tar.ts
```diff
@@ -0,0 +1,49 @@
...
+class Tar {
+  constructor(private data?: string | Buffer, private path?: string) {}
+
+  /**
+   * Extract the specified files from the associated TAR file.
+   * @param paths list of files paths to extract from the associated TAR file
+   */
+  public async extract(paths: string[]) {
+    return new Promise((resolve) => {
+      const result = {};
```
robcresswell: `result` needs typing

lib/tar.ts
```diff
@@ -0,0 +1,49 @@
...
+      const tarExtract = extract() as Extract;
```
robcresswell: ```const tarExtract: Extract = extract()```

lib/tar.ts
```diff
@@ -0,0 +1,49 @@
+import { readFileSync } from "fs";
+import { extract, Extract } from "tar-stream";
```
robcresswell: Do we really want to use both `tar` and `tar-stream`? We have two different extraction approaches now within the same library. Lets use one please. For example, using `tar` and iterating over the list of files instead of streaming is probably perfectly fine.

lib/analyzer/apt-analyzer.ts
```diff
@@ -3,12 +3,25 @@ import { AnalyzerPkg } from "./types";
...
+
+export { APT_PKGFILES };
+
+async function analyze(
+  targetImage: string,
+  options?: DockerOptions,
+  files?: {} | void,
+) {
+  if (!files) {
+    files = await new Docker(targetImage, options).extract(APT_PKGFILES);
```
robcresswell: I'm a little confused... are these not passed in by the `analyzer/index`?

lib/analyzer/apt-analyzer.ts
moshikod: Done

lib/analyzer/apt-analyzer.ts
moshikod: This is a mandatory file, should we just return or throw some error?

lib/tar.ts
moshikod: Done

lib/tar.ts
moshikod: Done

lib/tar.ts
moshikod: The only library we use is `tar-stream`

lib/analyzer/apt-analyzer.ts
moshikod: I'll remove this once I'll fix and verify that all tests pass

robcresswell: I don't follow... Why not remove now? If you're going to remove it anyway I don't understand why it's needed during review. Am I missing something here?

lib/analyzer/apt-analyzer.ts
```diff
@@ -1,16 +1,34 @@
...
+) {
+  // TODO: remove when done, backwards compatibility
+  if (!files) {
+    files = await new Docker(targetImage, options).extract(APT_PKGFILES);
+  }
+
+  const dpkgFile = files[APT_DPKG_STATUS];
+  const pkgs = parseDpkgFile(
+    dpkgFile ? dpkgFile.toString(DEFAULT_ENCODING) : "",
+  );
```
robcresswell: In what circumstance does `dpkgFile` not exist but we so want to continue the function? Shouldn't we return immediately if it does not exist?

lib/analyzer/image-extractor.ts
```diff
@@ -0,0 +1,96 @@
...
+const MANIFEST_JSON: string = "manifest.json";
+const DEFAULT_ENCODING: string = "utf8";
+
+interface Extracted {
+  Manifest?: string;
+  Layers?: { [key: string]: { [key: string]: Buffer } };
+}
+
+async function streamToString(stream: Readable): Promise<string> {
+  const chuncks: string[] = [];
```
robcresswell: typo: `chunks`

package.json
```diff
@@ -19,9 +19,15 @@
     "prepare": "npm run build"
   },
   "dependencies": {
+    "@types/steam": "0.0.27",
+    "@types/tar-stream": "^1.6.0",
+    "@types/tmp": "^0.1.0",
     "debug": "^4.1.1",
     "dockerfile-ast": "0.0.16",
     "semver": "^6.1.0",
+    "steam": "^1.4.1",
```
robcresswell: Typo, can be removed.

package.json
```diff
@@ -19,9 +19,15 @@
     "prepare": "npm run build"
   },
   "dependencies": {
+    "@types/steam": "0.0.27",
```
robcresswell: Typo, can be removed

lib/analyzer/apk-analyzer.ts
```diff
@@ -1,19 +1,30 @@
 import { Docker, DockerOptions } from "../docker";
+import { DEFAULT_ENCODING } from "./image-extractor";
 import { AnalyzerPkg } from "./types";
```
robcresswell: Making the encoding abstracted here is probably a little bit much, I think it's fine to inline "utf8" in the function call; fairly standard stuff.

lib/docker.ts
```diff
@@ -86,4 +87,85 @@ class Docker {
...
+
+      if (extracted.Layers) {
+        // reverse layer order from last to first
+        for (const layerName of layersNames.reverse()) {
+          // files found for this layer
+          if (layerName in extracted.Layers) {
+            // go over files found in this layer
+            for (const filename of Object.keys(extracted.Layers[layerName])) {
+              // file was not found in previous layer
+              if (!(filename in result)) {
```
robcresswell: ```if (!Reflect.has(result, filename))```

Given the range of filenames possible, we need to be careful about overriding object, weirdly enough

lib/analyzer/apk-analyzer.ts
moshikod: Done

lib/analyzer/apt-analyzer.ts
moshikod: Done

lib/analyzer/apt-analyzer.ts
moshikod: 👍

lib/analyzer/image-extractor.ts
moshikod: 👍

lib/docker.ts
moshikod: Thanks

package.json
moshikod: 👍

package.json
moshikod: 👍

lib/analyzer/apk-analyzer.ts
```diff
@@ -1,19 +1,23 @@
-import { Docker, DockerOptions } from "../docker";
 import { AnalyzerPkg } from "./types";
 export { analyze };
 
-function analyze(targetImage: string, options?: DockerOptions) {
-  return getPackages(targetImage, options).then((pkgs) => ({
+const APK_DB_INSTALLED = "lib/apk/db/installed";
```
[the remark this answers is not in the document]
moshikod: @leongold the `tar-stream` module use the full path with no leading slash so actually there is actually no relative path. We can add leading slash trimming to make this less confusing, wdyt?

lib/analyzer/image-extractor.ts
```diff
@@ -0,0 +1,121 @@
+import { createReadStream } from "fs";
+import * as md5 from "md5";
+import { basename } from "path";
+import { Readable } from "stream";
+import { extract, Extract } from "tar-stream";
+
+export { extractImageKeyFiles, ExtractedImage, ExtractedKeyFiles };
+
+interface ExtractedKeyFiles {
+  txt: { [key: string]: string };
```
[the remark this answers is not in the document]
moshikod: The key files extraction returns a list of strings for textual key files (in our case, repositories saved as plain text) and a list of md5 sum strings for binary key files.

lib/analyzer/image-extractor.ts
```diff
@@ -0,0 +1,121 @@
...
+  md5: { [key: string]: string };
+}
+
+interface ExtractedImage {
+  Manifest?: string;
+  Layers: { [key: string]: ExtractedKeyFiles };
+}
+
+async function streamToString(stream: Readable): Promise<string> {
```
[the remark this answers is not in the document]
moshikod: Looked for but found nothing, if there's a more generic approach I probably missed it

lib/analyzer/image-extractor.ts
```diff
@@ -0,0 +1,121 @@
...
+      resolve(chunks.join(""));
+    });
+    stream.on("error", reject);
+    stream.on("data", (chunk) => {
+      chunks.push(chunk.toString("utf8"));
+    });
+  });
+}
+
+async function streamToBuffer(stream: Readable): Promise<Buffer> {
```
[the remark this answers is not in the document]
moshikod: again, nothing that I found ?!?

lib/analyzer/image-extractor.ts
```diff
@@ -0,0 +1,121 @@
...
+  imagePath: string,
+  txtPaths: string[],
+  md5Paths: string[] = [],
+): Promise<ExtractedImage> {
+  return new Promise((resolve) => {
+    const result: ExtractedImage = { Layers: {} };
+    const imageExtract: Extract = extract();
+    imageExtract.on("entry", (header, stream, next) => {
+      if (header.type === "file") {
+        if ("layer.tar" === basename(header.name)) {
```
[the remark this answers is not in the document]
moshikod: I used for constants on the left which is a bit safer in C/C++ if you forget to put a double (or triple in our case) equal sign.
What is the preferred convention in JS/TS?

lib/analyzer/image-extractor.ts
```diff
@@ -0,0 +1,121 @@
...
+          extractLayerKeyFiles(stream, txtPaths, md5Paths).then((value) => {
```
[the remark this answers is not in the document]
moshikod: Done

lib/docker.ts
```diff
@@ -86,4 +91,106 @@ class Docker {
...
+      }
+
+      const extracted: ExtractedImage = await extractImageKeyFiles(
+        name,
+        txtPaths,
+        md5Paths,
+      );
+
+      // No manifest file
+      if (!extracted.Manifest) {
```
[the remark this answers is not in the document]
moshikod: But then how can one tell the order of the layers?

lib/analyzer/image-extractor.ts
```diff
@@ -0,0 +1,121 @@
...
+ * @param name layer name
+ * @param txtPaths list of plain text key files paths to extract and return as strings
+ * @param md5Paths list of binary key files paths to extract and return their MD5 sum
+ * @returns key files found in the specified layer TAR stream
+ */
+async function extractLayerKeyFiles(
+  stream: Readable,
+  txtPaths: string[],
+  md5Paths: string[] = [],
+): Promise<{ txt: { [key: string]: string }; md5: { [key: string]: string } }> {
```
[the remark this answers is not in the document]
moshikod: Done

lib/analyzer/image-extractor.ts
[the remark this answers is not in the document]
leongold: interesting! but let's use const to the right as it's much much more used in our codebase, not sure convention wise but I'd wager it's const to the right

lib/docker.ts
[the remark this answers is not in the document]
leongold: no no, I mean checking for it is redundant (it's guaranteed to exist unless extraction failed or it's not an OCI image. if we want ot check the latter case, make it explicit)

test/lib/analyzer/index.test.ts
```diff
@@ -16,9 +16,9 @@ const readOsFixtureFile = (...from) =>
 
 test("analyzer", async (t) => {
   const examples = {
-    "alpine:2.6": {
-      dir: "alpine_2_6_6",
-    },
+    // "alpine:2.6": {
```
leongold: remove?

lib/analyzer/image-extractor.ts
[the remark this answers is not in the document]
moshikod: Done

fake-image
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
robcresswell: Make a `scripts/` dir, put this in it, amend the file extension (`.sh`) and leave a comment explaining what it does in the file please

package.json
```diff
@@ -16,12 +16,23 @@
...
+    "@types/minimatch": "^3.0.3",
+    "@types/tar-fs": "^1.16.1",
+    "@types/tar-stream": "^1.6.0",
+    "@types/tmp": "^0.1.0",
     "debug": "^4.1.1",
     "dockerfile-ast": "0.0.16",
+    "md5": "^2.2.1",
+    "minimatch": "^3.0.4",
     "semver": "^6.1.0",
+    "tar-fs": "^2.0.0",
```
robcresswell: I don't think we need both `tar-fs` and `tar-stream`? Same with the `@types`

test/perf.ts
```diff
@@ -0,0 +1,230 @@
...
+  while (index < process.argv.length) {
+    posargs.push(process.argv[index++]);
+  }
+  return { opt: optargs, pos: posargs };
+}
+
+function main() {
+  const args = argparse();
+  if ("h" in args.opt) {
+    // tslint:disable-next-line
```
robcresswell: Disable the specific rule please; `disable-next-line no-console` for example (or whatever the rule is)

test/lib/analyzer/index.test.ts
```diff
@@ -173,12 +229,31 @@ test("analyzer", async (t) => {
...
+      //   );
+      // }
+      const expectation = JSON.parse(
+        readOsFixtureFile(example.dir, "analyzer-expect.json"),
+      );
+
+      // // tslint:disable-next-line
+      // console.error(actual.results[2]);
+      // // tslint:disable-next-line
+      // console.error(expectation.results[2]);
```
robcresswell: Can we remove all this commented out code please?

test/lib/analyzer/index.test.ts
moshikod: Fixed

lib/docker.ts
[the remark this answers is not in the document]
moshikod: Done

fake-image
moshikod: Done

package.json
moshikod: Moved from `dependencies` section to `devDependencies` section since it is used by `test/lib/image-extractor.test.ts`

test/perf.ts
moshikod: Done

test/lib/analyzer/index.test.ts
moshikod: Done

test/perf.ts
moshikod: Done, again, now only for the specific lines of code

PR #89, opened 2019-06-16
feat: limit static scan by image size

moshikod:
- [x] Ready for review

#### What does this PR do?

Limit static scan to images of size smaller than 3GB

[DC-240](https://snyksec.atlassian.net/browse/DC-240)


lib/analyzer/image-extractor.ts
```diff
@@ -1,56 +1,61 @@
...
+interface ExtractedFilesByLayers {
+  [key: string]: ExtractedFiles;
 }
 
-interface Layers {
-  [key: string]: ExtractedKeyFiles;
+type ExtractFileCallback = (stream: Readable) => Promise<string>;
+
+interface LookupEntry {
+  p: string; // path pattern to look for
```
orkamara:
```suggestion
  pattern: string; // path pattern to look for
```

lib/analyzer/image-extractor.ts
```diff
@@ -1,56 +1,61 @@
...
+  c: ExtractFileCallback; // handler to manipulate stream into a string
```
orkamara:
```suggestion
  callback: ExtractFileCallback; // handler to manipulate stream into a string
```

lib/analyzer/image-extractor.ts
```diff
@@ -1,56 +1,61 @@
...
-  layers: Layers;
+interface ExtractedFilesByLayers {
+  [key: string]: ExtractedFiles;
 }
 
-interface Layers {
-  [key: string]: ExtractedKeyFiles;
+type ExtractFileCallback = (stream: Readable) => Promise<string>;
+
+interface LookupEntry {
```
orkamara: I think Lookup isn't a good name - the name of the interface should explain the it's related to callbacks

lib/analyzer/image-extractor.ts
```diff
@@ -1,56 +1,61 @@
...
-          });
-        }
-      });
-      md5Patterns.forEach((pattern) => {
-        if (minimatch(header.name, pattern, { dot: true })) {
-          streamToBuffer(stream).then((value) => {
-            result.md5[header.name] = md5(value);
-          });
+      lookups.forEach((lookup) => {
+        if (minimatch(`/${header.name}`, lookup.p, { dot: true })) {
```
orkamara: please create a const for `/${header.name}` - it's being used twice

lib/analyzer/image-extractor.ts
```diff
@@ -60,51 +65,64 @@ async function extractLayerKeyFiles(
...
+
     imageExtract.on("finish", () => {
-      // all image level entries read
-      resolve({
-        manifest,
-        layers,
-      });
+      const result: ExtractedFiles = {};
+
+      if (layers) {
```
orkamara: let's take this code into a separate function

lib/analyzer/image-extractor.ts
```diff
@@ -60,51 +65,64 @@ async function extractLayerKeyFiles(
...
+        // reverse layer order from last to first
+        for (const layerName of layersNames.reverse()) {
```
orkamara: please change the `if`s statements, so you will have less indentations.
Instead of:
`if (x) { if (y) {return z;}` you can do `if (!x) {return} if (y) {return z}`

lib/docker.ts
```diff
@@ -92,6 +101,17 @@ class Docker {
     }
   }
 
+  public async size(): Promise<number | undefined> {
+    try {
```
orkamara: why do you need to wrap it with size? not sure if it should be a safe function

lib/analyzer/image-extractor.ts
```diff
@@ -1,56 +1,61 @@
...
-        }
-      });
-      md5Patterns.forEach((pattern) => {
-        if (minimatch(header.name, pattern, { dot: true })) {
-          streamToBuffer(stream).then((value) => {
-            result.md5[header.name] = md5(value);
-          });
+      lookups.forEach((lookup) => {
+        if (minimatch(`/${header.name}`, lookup.p, { dot: true })) {
+          lookup.c(stream).then((value) => (result[`/${header.name}`] = value));
```
orkamara: I think the results should be split into array according to the steam - otherwise there is any ovelap

lib/analyzer/image-extractor.ts
moshikod: Done, split the code into `extractLayersFromTar` and `extractFromTar`

lib/analyzer/image-extractor.ts
moshikod: Done explicitly by adding `if (!layers) { return {}; }` and implicitly after the split into separate function made 2 of the loops redundant.

lib/docker.ts
moshikod: Errors are ignored for backwards compatibility with the expected behaviour as being reflected by many tests.
I renamed the function to `sizeSafe` to better expose this,

lib/analyzer/apk-analyzer.ts
```diff
@@ -1,23 +1,23 @@
...
-async function analyze(
-  targetImage: string,
-  pkgFiles: { [key: string]: string },
-) {
-  const dbFile = pkgFiles[APK_DB_INSTALLED];
-  const pkgs = dbFile ? parseFile(dbFile) : [];
-  return {
-    Image: targetImage,
+async function analyze(docker: Docker) {
+  return getPackages(docker).then((pkgs) => ({
```
orkamara: why do you need `then`? does `getPackages` return a promise?

lib/analyzer/os-release-detector.ts
```diff
@@ -55,7 +54,10 @@ async function detect(
 }
 
 async function tryOSRelease(docker: Docker): Promise<OSRelease | null> {
-  const text = await tryRelease(docker, "/etc/os-release");
+  // /etc/os-release is a symlink to /usr/lib/os-release
+  const text =
+    (await tryRelease(docker, "/usr/lib/os-release")) ||
```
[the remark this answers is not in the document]
moshikod: Added the symlink `etc/os-release` -> `../usr/lib/os-release` to `debian_9` fixture, this simuates a real case

lib/docker.ts
```diff
@@ -132,69 +152,52 @@ class Docker {
...
+   * @param filename name of file to retrieve its associated string
+   * @param callback optional callback to call when runti,e method is used to retrieve the file
+   */
+  public async getFile(
+    filename: string,
+    callback?: ExtractFileCallback,
+  ): Promise<string> {
+    if (Reflect.has(this.extractedFiles, filename)) {
+      return this.extractedFiles[filename];
+    } else {
```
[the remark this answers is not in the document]
moshikod: Done

lib/stream-utils.ts
```diff
@@ -41,3 +42,26 @@ async function streamToBuffer(stream: Readable): Promise<Buffer> {
     });
   });
 }
+
+/**
+ * hash the specified stream with MD5
+ * @param stream a stream to hash
+ * @returns the resultant MD5 hash of the given stream
+ */
+async function md5Stream(stream: Readable): Promise<string> {
```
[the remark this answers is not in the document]
moshikod: Deprecated, the function removed

lib/stream-utils.ts
```diff
@@ -41,3 +42,26 @@ async function streamToBuffer(stream: Readable): Promise<Buffer> {
...
+  return md5(await streamToBuffer(stream));
```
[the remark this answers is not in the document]
moshikod: Deprecated, the function removed

lib/stream-utils.ts
```diff
@@ -41,3 +42,26 @@ async function streamToBuffer(stream: Readable): Promise<Buffer> {
...
+}
+
+/**
+ * convert the specified string to stream
+ * @param str string to convert to stream
+ * @returns stream with the content of the specified string
+ */
+async function stringToStream(str: string): Promise<Readable> {
```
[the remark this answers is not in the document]
moshikod: Deprecated, the function removed.
Used to support compatibility with `catSafe` but I changed the use of `stream` to `buffer` in order to support processing of a single file by multiple callbacks.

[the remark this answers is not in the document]
moshikod: Done

lib/stream-utils.ts
```diff
@@ -41,3 +42,26 @@ async function streamToBuffer(stream: Readable): Promise<Buffer> {
...
+  const s = new Readable();
```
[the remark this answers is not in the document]
moshikod: Deprecated, the function removed

lib/docker.ts
```diff
@@ -132,69 +152,52 @@ class Docker {
...
+      });
     });
   }
+
+  /**
+   * Get file product that was previously retrieved and cached or using runtime method and the specified callback
+   * @param filename name of file to retrieve its associated string
+   * @param callback optional callback to call when runti,e method is used to retrieve the file
+   */
+  public async getFile(
```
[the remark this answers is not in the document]
moshikod: Changed to `getFileProduct` (together with the more general function `getFilesProducts`)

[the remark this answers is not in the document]
moshikod: Tests were added to `test/lib/docker.test.ts`

lib/analyzer/image-extractor.ts
```diff
@@ -1,110 +1,181 @@
...
+interface SearchActionProducts {
+  [key: string]: { [key: string]: string | Buffer };
 }
 
-interface ExtractedImage {
-  manifest: string;
-  layers: Layers;
+interface SearchActionCallback {
+  name: string; // name, should be unique, for this action
+  call: (buffer: Buffer) => string | Buffer; // convert Buffer into a string or Buffer
```
orkamara:
```suggestion
  callback: (buffer: Buffer) => string | Buffer; // convert Buffer into a string or Buffer
```

lib/analyzer/apk-analyzer.ts
```diff
@@ -1,26 +1,28 @@
...
-    Image: targetImage,
+    Image: docker.getTargetImage(),
     AnalyzeType: "Apk",
     Analysis: pkgs,
   };
 }
 
-function parseFile(text: string) {
+async function getPackages(docker: Docker): Promise<AnalyzerPkg[]> {
+  const dbFileContent = await docker.getFileProduct(APK_DB_INSTALLED);
```
orkamara: can `getFileProduct` return undefined? if so, you don't need to call `parseFile`

lib/analyzer/index.ts
```diff
@@ -18,27 +18,28 @@ async function analyze(
...
   const docker = new Docker(targetImage, options);
-  let pkgFiles: ExtractedKeyFiles;
 
-  try {
-    pkgFiles = await docker.extract(pkgPaths);
-  } catch (err) {
-    debug(err);
-    throw new Error(err);
-  }
+  await docker.maybeStaticScan(
```
orkamara: can you please replace `maybeStaticScan` with something like `scanStaticalyIfNeeded`?

lib/docker.ts
```diff
@@ -132,69 +174,99 @@ class Docker {
...
-    });
+  public async getFileProduct(
+    filename: string,
+    callback?: SearchActionCallback,
+  ): Promise<string> {
+    const result = await this.getFilesProducts(
+      filename,
+      callback ? [callback] : undefined,
+    );
+    const length = Object.keys(result).length;
```
orkamara: your'e using `Object.keys(result)` twice - please extract to const

lib/docker.ts
```diff
@@ -132,69 +174,99 @@ class Docker {
...
+    if (length > 1) {
+      throw new Error(`File product ambiguity, ${length}`);
+    }
+    // return the first and only product of the file
+    return result[Object.keys(result)[0]].toString("utf8");
```
orkamara: Why do you need the `utf8`? isn't that function support content that is not only txt?

lib/analyzer/index.ts
```diff
@@ -4,41 +4,54 @@ import * as dockerFile from "../docker-file";
...
+  const size = await docker.size();
 
-  try {
-    pkgFiles = await docker.extract(pkgPaths);
-  } catch (err) {
-    debug(err);
-    throw new Error(err);
+  // ignore undefined size for backwards compatibility
+  if (size && size <= STATIC_SCAN_MAX_IMAGE_SIZE) {
+    await docker.extractAndCache(
```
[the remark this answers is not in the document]
orkamara: I cannot see the change...

lib/analyzer/image-extractor.ts
```diff
@@ -1,110 +1,181 @@
...
   });
 }
+
+/**
+ * Extract key files textual content and MD5 sum from the specified TAR file
+ * @param imageTarPath path to image file saved in tar format
+ * @param searchActions array of pattern, callbacks pairs
+ * @returns extracted files products
+ */
+async function extractFromTar(
```
orkamara: please revert the changes of this function so it will be like before (https://github.com/snyk/snyk-docker-plugin/pull/89/files#diff-1a778f82722b9f884412efd52d3c46d9L157) - I don't want this change as part of this PR

lib/analyzer/image-extractor.ts
```diff
@@ -69,23 +69,23 @@ async function extractFromLayer(
...
           if (minimatch(name, searchAction.pattern, { dot: true })) {
-            if (header.type === "link" || header.type === "symlink") {
-              debug(`${header.type} '${header.name}' -> '${header.linkname}'`);
-            } else {
-              if (!buffer) {
-                buffer = await streamToBuffer(stream);
-              }
+            if (header.type === "file") {
+              // convert stream to buffer
+              buffer = buffer || (await streamToBuffer(stream));
```
orkamara: I think it's better to do as before (if (!buffer)) - because you're not doing reassignment, but initialization of the code

lib/analyzer/image-extractor.ts
```diff
@@ -69,23 +69,23 @@ async function extractFromLayer(
...
               // initialize the files associated products dict
-              if (!result[name]) {
-                result[name] = {};
-              }
+              result[name] = result[name] || {};
```
orkamara: I think it's better to do as before (if (!result[name])) - because you're not doing reassignment, but initialization of the code

lib/analyzer/image-extractor.ts
```diff
@@ -1,110 +1,181 @@
...
+interface SearchActionCallback {
+  name: string; // name, should be unique, for this action
+  callback: (buffer: Buffer) => string | Buffer; // convert Buffer into a string or Buffer
 }
 
-interface Layers {
-  [key: string]: ExtractedKeyFiles;
+interface SearchAction {
+  pattern: string; // path pattern to look for
+  callbacks: SearchActionCallback[]; // array of handlers
```
orkamara: why `SearchActionCallback` is an array?

lib/analyzer/image-extractor.ts
```diff
@@ -1,110 +1,181 @@
...
+  SearchAction,
+};
 
-interface ExtractedKeyFiles {
-  txt: { [key: string]: string };
-  md5: { [key: string]: string };
+const debug = Debug("snyk");
+
+interface SearchActionProducts {
+  [key: string]: { [key: string]: string | Buffer };
```
orkamara: what's each one of the keys? please use different names

lib/analyzer/image-extractor.ts
```diff
@@ -1,110 +1,181 @@
...
-      });
-      md5Patterns.forEach((pattern) => {
-        if (minimatch(header.name, pattern, { dot: true })) {
-          streamToBuffer(stream).then((value) => {
-            result.md5[header.name] = md5(value);
-          });
+    layerExtract.on("entry", async (header, stream, next) => {
+      if (
+        header.type === "file" ||
+        header.type === "link" ||
```
orkamara: why do we need the ifs for `link` + `symlink` if you ignore them later?

lib/analyzer/image-extractor.ts
```diff
@@ -1,110 +1,181 @@
...
+        header.type === "symlink"
+      ) {
+        const name = `/${header.name}`;
```
orkamara:
```suggestion
        const fileName = `/${header.name}`;
```

lib/analyzer/image-extractor.ts
```diff
@@ -1,110 +1,181 @@
...
+        let buffer: Buffer | undefined;
+        for (const searchAction of searchActions) {
+          if (minimatch(name, searchAction.pattern, { dot: true })) {
+            if (header.type === "file") {
+              // convert stream to buffer
+              buffer = buffer || (await streamToBuffer(stream));
```
orkamara: not sure what are we doing here...`stream` is the same for all of the iterations of the loop, right? so why calculating `buffer` again and again?

lib/analyzer/image-extractor.ts
```diff
@@ -1,110 +1,181 @@
...
-  txtPatterns: string[],
-  md5Patterns: string[] = [],
-): Promise<ExtractedImage> {
+  searchActions: SearchAction[],
+): Promise<SearchActionProducts[]> {
   return new Promise((resolve) => {
     const imageExtract: Extract = extract();
-    let manifest: string;
-    const layers: Layers = {};
+    const layers: { [key: string]: SearchActionProducts } = {};
```
orkamara: layers is `{[key: string]: [key: string]: { [key: string]: string | Buffer };}`? if so - le'ts rethink about it - feels like overgeneric

lib/docker.ts
```diff
@@ -132,69 +174,93 @@ class Docker {
...
+  }
+
+  /**
+   * Convenience function the single product of a single file
+   * @param filename name of file to retrieve its associated single product
+   * @param callback optional callback to call when runtime method is used
+   *  to retrieve the file
+   * @returns file product, the only one
+   */
+  public async getFileProduct(
```
orkamara: we need to add function to get products by callback name

lib/docker.ts
```diff
@@ -132,69 +174,93 @@ class Docker {
...
+    filename: string,
+    searchActionCallback?: SearchActionCallback,
+  ): Promise<string> {
+    const result = await this.getFilesProducts(filename, searchActionCallback);
+    const resultNames = Object.keys(result);
+    const length = resultNames.length;
+    if (length > 1) {
```
orkamara: let's remove this `if` and get callback name as param

lib/analyzer/image-extractor.ts
orkamara: so just split into 2 different ifs - one for `file`and one for the 2 others. This way you'll remove the if inside if

lib/analyzer/image-extractor.ts
```diff
@@ -1,110 +1,164 @@
...
+                result[filename] = {};
+              }
+              try {
+                // store the product under the search action name
+                result[filename][searchAction.name] = searchAction.callback(
+                  buffer,
+                );
+              } catch (error) {
+                // prevent static scan from crash due to callback error
+                debug(`${filename} ${searchAction.name} ${error}`);
```
orkamara: I think we should throw in this case - this way, you'll not get familiar with the fact that there is any problem...

lib/analyzer/image-extractor.ts
```diff
@@ -1,110 +1,181 @@
...
-            (extractedKeyFiles) => {
-              layers[header.name] = extractedKeyFiles;
-            },
-          );
+          layers[header.name] = await extractFromLayer(stream, searchActions);
         } else if (header.name === "manifest.json") {
           streamToString(stream).then((manifestFile) => {
-            manifest = manifestFile;
+            const manifest = JSON.parse(manifestFile);
+            layersNames = manifest[0].Layers;
```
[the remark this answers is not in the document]
orkamara: I do see a change - it was `manifest = manifestFile;` but now its `manifest = manifest[0].Layers;`

lib/analyzer/index.ts
```diff
@@ -18,27 +17,31 @@ async function analyze(
...
-    throw new Error(err);
-  }
+  await docker.scanStaticalyIfNeeded(
+    [
+      ...aptAnalyzer.APT_PKGPATHS,
+      ...apkAnalyzer.APK_PKGPATHS,
+      ...osReleaseDetector.OS_VERPATHS,
+    ].map((p) => {
+      return {
+        name: "",
```
orkamara: let's add name for the string

lib/docker.ts
```diff
@@ -132,69 +174,93 @@ class Docker {
...
+   *  runtime method after applying the specified callback
+   * @param filename name of file to retrieve its associated string product
+   * @param callbacks optional array of callbacks to call when runtime method
+   *  is used to retrieve the file
+   * @returns map of file products by callback name
+   */
+  public async getFilesProducts(
+    filename: string,
+    searchActionCallback?: SearchActionCallback,
+  ): Promise<{ [key: string]: string | Buffer }> {
```
orkamara: create type for `[searchActionName: string]: string | Buffer` and use it here and in `interface SearchActionProducts` declaration

lib/docker.ts
```diff
@@ -132,69 +174,93 @@ class Docker {
...
+    if (Reflect.has(this.searchActionProducts, filename)) {
+      return this.searchActionProducts[filename];
+    }
+    const content = (await this.catSafe(filename)).stdout;
+    if (!searchActionCallback) {
+      return { "": content };
```
orkamara: let's not use `""`

lib/docker.ts
```diff
@@ -132,69 +174,93 @@ class Docker {
...
+    searchActionCallback?: SearchActionCallback,
+  ): Promise<string> {
+    const result = await this.getFilesProducts(filename, searchActionCallback);
+    const resultNames = Object.keys(result);
+    const length = resultNames.length;
+    if (length > 1) {
+      throw new Error(`File product ambiguity, ${length}`);
+    }
+    // return the first and only product of the file
+    return result[resultNames[0]].toString("utf8");
```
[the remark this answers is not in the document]
moshikod: Done
Function removed

lib/docker.ts
orkamara: let's do as we talked - create a different function for `txt` only so the call to `getFilesProducts` will be separate from `catSafe`

lib/docker.ts
moshikod: Done + test

lib/docker.ts
moshikod: Done

test/lib/analyzer/rpm-analyzer.test.ts
```diff
@@ -79,7 +80,7 @@ test("analyze", async (t) => {
 
       t.teardown(() => execStub.restore());
 
-      const actual = await analyzer.analyze("centos:6");
+      const actual = await analyzer.analyze(new Docker("centos:6"));
```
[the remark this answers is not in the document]
moshikod: Done

lib/docker.ts
```diff
@@ -132,69 +180,110 @@ class Docker {
...
+        // check for product associated with the specified action name
+        .filter((filename) =>
+          Reflect.has(
+            this.extractProductsByFilename[filename],
+            searchActionName,
+          ),
+        )
+        // create a map of this filename and product
+        .map((filename) => ({
+          [filename]: this.extractProductsByFilename[searchActionName],
```
mladkau: Sorry, but this will not work. The line should be:
```suggestion
          [filename]: this.extractProductsByFilename[filename][searchActionName]
```

lib/docker.ts
```diff
@@ -132,69 +180,110 @@ class Docker {
...
+      ? extractCallback(Buffer.from(content, "utf8"))
+      : content;
+  }
+
+  /**
+   * Retrieve all filenames with products of the specified action name
+   * @param searchActionName name of a search action
+   */
+  public getFileProducts(searchActionName: string = "txt"): FilenameProducts {
+    return Object.assign(
```
mladkau: This function is a bit convoluted - how about:
```suggestion
    return Object.keys(this.extractProductsByFilename).reduce(
      (acc : FilenameProducts, file : string) => {
        const val = this.extractProductsByFilename[file][searchActionName];
        return val ? Object.assign(acc, {[file] : val}): acc;
    }, {});
```
Or did I miss something?

lib/docker.ts
moshikod: @mladkau great catch updated the test to verify the products of the file and not just the filename existence

lib/docker.ts
moshikod: @mladkau sure, a much elegant approach, thanks

lib/docker.ts
```diff
@@ -47,12 +61,29 @@ class Docker {
...
+  public getTargetImage(): string {
+    return this.targetImage;
+  }
+
+  public GetStaticScanSizeLimit(): number {
+    return this.staticScanSizeLimitInBytes;
   }
 
-  public run(cmd: string, args: string[] = []) {
+  public run(cmd: string, args: string[] = []): Promise<CmdOutput> {
```
orkamara: why Promise?

lib/docker.ts
```diff
@@ -1,12 +1,22 @@
...
-export { Docker, DockerOptions };
+export { Docker, DockerOptions, STATIC_SCAN_MAX_IMAGE_SIZE_IN_BYTES };
+
+const debug = Debug("snyk");
+
+const KB = 1024;
+const MB = KB * 1024;
+const GB = MB * 1024;
+
+const STATIC_SCAN_MAX_IMAGE_SIZE_IN_BYTES = 3 * GB;
```
orkamara: shouldn't we make it less? I think we should limit the static scan to ~40 secs

lib/docker.ts
```diff
@@ -132,69 +180,101 @@ class Docker {
...
+   * Extract files from image and store their product
+   * @param extractActions array of pattern, callbacks pairs
+   */
+  public async extractAndCache(extractActions: ExtractAction[]): Promise<void> {
+    try {
+      this.extractProductsByFilename = Object.assign(
+        this.extractProductsByFilename,
+        await this.extract(extractActions),
+      );
+    } catch (error) {
```
orkamara: no need for this try & catch

lib/docker.ts
```diff
@@ -132,69 +180,101 @@ class Docker {
...
-  public async analyze(
-    imageTarPath: string,
-    txtPatterns: string[],
-    md5Patterns: string[] = [],
-  ) {
-    const result: ExtractedKeyFiles = { txt: {}, md5: {} };
+  public async extract(
+    extractActions: ExtractAction[],
+  ): Promise<ExtractProductsByFilename> {
+    return this.save(async (err, imageTarPath) => {
```
orkamara: change to async await

lib/docker.ts
```diff
@@ -132,69 +180,101 @@ class Docker {
...
+    return extractCallback
+      ? extractCallback(Buffer.from(content, "utf8"))
+      : content;
+  }
+
+  /**
+   * Retrieve all filenames with products of the specified action name
+   * @param searchActionName name of a search action
+   */
+  public getFileProducts(searchActionName: string = "txt"): FilenameProducts {
```
orkamara: let's remove the default value so it will be safer to use

lib/docker.ts
```diff
@@ -132,69 +180,101 @@ class Docker {
...
+      const callbackProducts = this.extractProductsByFilename[filename];
+      if (Reflect.has(callbackProducts, searchActionName)) {
+        return callbackProducts[searchActionName];
       }
-      return await this.analyze(imageTarPath, txtPatterns, md5Patterns);
-    });
+    }
+    const content = (await this.catSafe(filename)).stdout;
+    return extractCallback
+      ? extractCallback(Buffer.from(content, "utf8"))
```
orkamara: why are we using "utf8"?

lib/docker.ts
```diff
@@ -132,69 +180,101 @@ class Docker {
...
+    );
+  }
+
+  /**
+   * Backward compatibility
+   * @param filename name of file to retrieve its associated string product
+   * @param searchActionName name of a search action
+   */
+  public async getTextFile(filename: string): Promise<string> {
+    return (await this.getFileProduct(filename, "txt")).toString("utf8");
```
orkamara: please take out `await this.getFileProduct(filename, "txt")` into a different const

lib/docker.ts
```diff
@@ -132,69 +180,101 @@ class Docker {
...
+   * @returns map of file products by callback name
    */
-  public async extract(
-    txtPatterns: string[],
-    md5Patterns: string[] = [],
-  ): Promise<ExtractedKeyFiles> {
-    return this.save(async (err, imageTarPath) => {
-      if (err) {
-        throw err;
+  public async getFileProduct(
```
orkamara:
```suggestion
  public async getActionProductByFileName(
```

lib/docker.ts
moshikod: Simply because `execute` (`lib\sub-process.ts`) returns the same

lib/docker.ts
moshikod: 3GB takes about 40 secs on my laptop but I would like to change the value to comply with the `MAX_IMAGE_SIZE_IN_BYTES` default value in `docker-registry-agent` which is set to 2GB

lib/docker.ts
moshikod: Because I should convert the returned string from `catSafe` (backward compatibility) to a Buffer which is the callback parameter type

lib/docker.ts
moshikod: Done

lib/docker.ts
moshikod: Done

lib/docker.ts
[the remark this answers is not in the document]
moshikod: Done

lib/docker.ts
moshikod: Done

lib/docker.ts
moshikod: maybe I miss something, I believe it's already in async-await form ?!?

lib/docker.ts
moshikod: DOne

PR #91, opened 2019-07-01
feat: Creating sha256 hashes for node key binaries

mladkau:
Code to create hash values for key binaries and send them to registry

### DIFF
--- a/lib/analyzer/hash-analyzer.ts
+++ b/lib/analyzer/hash-analyzer.ts
@@ -1,0 +1,21 @@
+import { Docker } from "../docker";
+
+export { analyze };
+
+const HASH_PKGPATHS = ["**/node"];
+
+export { HASH_PKGPATHS };
+
+async function analyze(docker: Docker) {
+  const result = docker.getActionProducts("hash");
+  return {
+    Image: docker.getTargetImage(),
+    AnalyzeType: "Hash",
+    Analysis: Object.keys(result).map((i) => {
+      return {
+        name: i,
+        hash: result[i],
+      };
+    }),
+  };
+}
--- a/lib/analyzer/index.ts
+++ b/lib/analyzer/index.ts
@@ -1,9 +1,11 @@
+import * as crypto from "crypto";
 import * as Debug from "debug";
 import { Docker, DockerOptions } from "../docker";
 import * as dockerFile from "../docker-file";
 import * as apkAnalyzer from "./apk-analyzer";
 import * as aptAnalyzer from "./apt-analyzer";
 import * as binariesAnalyzer from "./binaries-analyzer";
+import * as hashAnalyzer from "./hash-analyzer";
 import * as imageInspector from "./image-inspector";
 import * as osReleaseDetector from "./os-release-detector";
 import * as rpmAnalyzer from "./rpm-analyzer";
@@ -23,6 +25,18 @@ const extractActions = [
   };
 });
 
+const hashActions = [...hashAnalyzer.HASH_PKGPATHS].map((p) => {
+  return {
+    name: "hash",
+    pattern: p,
+    callback: (b) =>
+      crypto
+        .createHash("sha256")
+        .update(b)
+        .digest("hex"),
+  };
+});
+
 async function analyze(
   targetImage: string,
   dockerfileAnalysis?: dockerFile.DockerFileAnalysis,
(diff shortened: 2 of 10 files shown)


orkamara:
Consider renaming the PR + commit to something more specific (maybe about supporting SHAs / key binaries)

lib/analyzer/index.ts
```diff
@@ -16,12 +18,26 @@ const extractActions = [
...
-});
+]
+  .map((p) => {
+    return {
+      name: "txt",
+      pattern: p,
+    };
+  })
+  .concat(
+    [...hashAnalyzer.HASH_PKGPATHS].map((p) => {
```
orkamara: I think it will be better to create 2 different consts (one for every mapping) and than combine them together

lib/index.ts
```diff
@@ -163,6 +164,7 @@ function parseAnalysisResults(
     type: depType,
     depInfosList: analysisResult.Analysis,
     binaries: analysisJson.binaries.Analysis,
+    hashes: analysisJson.hashes.Analysis,
```
orkamara: Maybe we should rename it to binariesHashes?
We'll use this value in other places in the code (not only the plugin) and it should be clear what's the content of it

lib/analyzer/index.ts
```diff
@@ -67,6 +83,7 @@ async function analyze(
     osRelease,
     results,
     binaries,
+    hashes: await hashAnalyzer.analyze(docker),
```
orkamara: let's do the same as we're doing for `binaries` - initialize separately and wrap with try & catch

lib/analyzer/hash-analyzer.ts
```diff
@@ -0,0 +1,21 @@
+import { Docker } from "../docker";
+
+export { analyze };
+
+const HASH_PKGPATHS = ["**/node"];
```
orkamara: please add a manual test to make sure we're doing good when there are 2 different `node` binaries (not only in the scope of the plugin but also the rest of the system) - you can also do it after merging this PR :)

test/lib/analyzer/hash-analyzer.test.ts
```diff
@@ -0,0 +1,38 @@
...
+          .update(b)
+          .digest("hex"),
+    },
+  ]);
+
+  const actual = await analyzer.analyze(docker);
+  t.same(actual, {
+    Image: "alpine:2.6",
+    AnalyzeType: "Hash",
+    Analysis: [],
```
orkamara: Not sure exactly about the tests - maybe I missed it, but failed to find any test that actually check if the hashes were calculated. In all of the tests the `hashes` array is `[]`

mladkau: You are correct this test is a placeholder in the moment - same goes for the test with two different node versions....

lib/analyzer/hash-analyzer.ts
mladkau: Yes I need to add some tests. This PR is just the bare minimum to get it working ...

lib/index.ts
mladkau: I am not sure why binary hashes is more clear? Can hashes be anything else than binary? I use the shorter terminology throughout the code so would like to keep the naming consistent.

lib/analyzer/index.ts
```diff
@@ -23,14 +25,26 @@ const extractActions = [
   };
 });
 
+const hashActions = [...hashAnalyzer.HASH_PKGPATHS].map((p) => {
```
mladkau: Now a separate variable

lib/analyzer/index.ts
```diff
@@ -62,11 +76,20 @@ async function analyze(
     throw new Error("Failed to detect binaries versions");
   }
 
+  let hashes;
+  try {
+    hashes = await hashAnalyzer.analyze(docker);
+  } catch (err) {
+    debug(`Error while running hash analyzer: '${err}'`);
+    throw new Error("Failed to get the hash of a binary");
+  }
```
mladkau: As suggested - same as with binaries

orkamara:
maybe also rename the commit? that's what actually matter in the history :)

lib/index.ts
orkamara: think about it that at the end we'll have `hashes` section inside the DB - we need to make sure that people will understand what's the meaning of those hashes - so the trivial question for me (without any context) will be "hashes of what?"

mladkau: Not sure what you mean by "inside the DB". I would then add a comment to make it clear. Naming it binariesHashes I would think that these are the actual hash values of the objects in the binaries variable. Trying to convey a description with a variable name will be confusing either way.

PR #97, opened 2019-07-26
feat: Reading of manifest files

mladkau:
This plugin is now able to read multiple manifest files. Only the manifest files which could be found are included in the response.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,4 +1,5 @@
 import * as Debug from "debug";
+import * as path from "path";
 import * as analyzer from "./analyzer";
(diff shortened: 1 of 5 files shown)


lib/index.ts
```diff
@@ -120,6 +129,26 @@ function getDependencies(
     });
 }
 
+async function getManifestFiles(targetImage: string, options?: any) {
+  if (!options.manifestGlobs) {
+    return [];
+  }
```
jurglic: I like this. It allows simpler client code with less conditionals 👍

PR #98, opened 2019-08-06
fix: support multi-stage Dockerfile with name referral

moshikod:
- [x] Ready for review

#### What does this PR do?

Support name referral in multi stage Docker file
 
https://docs.docker.com/develop/develop-images/multistage-build/


test/fixtures/dockerfiles/multi-stage-as/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+FROM alpine:latest AS base
+RUN apk --no-cache add ca-certificates
+WORKDIR /root/
+COPY --from=builder /go/src/github.com/alexellis/href-counter/app .
+CMD ["./app"]
+
+FROM base
+
+FROM base AS extended
```
adrukh: How about another fixture that uses regular image tags as `FROM` args, has `AS` names on all but the last `FROM` statements? Like the public example from the dockerhub docs

moshikod: @adrukh I changed the fixture and flipped the order of the stages. I believe this reflects the common scenario.

I also added the Dockerfile example we got from the customer:
https://snyk.slack.com/archives/CBXS33GJY/p1552396933238600?thread_ts=1552396862.238300&cid=CBXS33GJY

lib/docker-file.ts
```diff
@@ -26,10 +26,45 @@ async function readDockerfileAndAnalyse(
...
+  const variables = instruction.getVariables();
+  if (variables) {
+    const resolvedVariables = variables.reduce((resolvedVars, variable) => {
+      const line = variable.getRange().start.line;
+      const name = variable.getName();
+      resolvedVars[name] = dockerfile.resolveVariable(name, line);
+      return resolvedVars;
+    }, {});
+    Object.keys(resolvedVariables).forEach((variable) => {
+      text = text.replace(`\$\{${variable}\}`, resolvedVariables[variable]);
```
mladkau: The library claims to do "variable lookup and resolution" yet here we do a rather crude text.replace to resolve variables ourself - Is there no better way to do this? I looked briefly at the library but the documentation seems to be rather "limited".

lib/docker-file.ts
```diff
@@ -26,10 +26,45 @@ async function readDockerfileAndAnalyse(
...
+  }
+  return text;
+}
+
 async function analyseDockerfile(
   contents: string,
 ): Promise<DockerFileAnalysis | undefined> {
   const dockerfile = DockerfileParser.parse(contents);
+  // collect stages names
+  const stages = dockerfile.getFROMs().reduce((names, from) => {
```
mladkau: This reduce is getting rather long - you might want to consider moving this into a separate function which returns the stages.

lib/docker-file.ts
```diff
@@ -26,10 +26,45 @@ async function readDockerfileAndAnalyse(
   return analyseDockerfile(contents);
 }
 
+function resolveVariables(
+  text: string,
+  instruction: Instruction,
+  dockerfile: Dockerfile,
+): string {
+  const variables = instruction.getVariables();
+  if (variables) {
```
mladkau: No need for this condition. Looking at the libraries code getVariables will always return a list. We should try to reduce cyclomatic complexity whenever possible.

lib/docker-file.ts
```diff
@@ -26,10 +26,45 @@ async function readDockerfileAndAnalyse(
...
+  const stages = dockerfile.getFROMs().reduce((names, from) => {
+    const args = from.getArguments();
+    if (args.length > 2 && args[1].getValue() === "AS") {
+      const fromBaseName = resolveVariables(
+        from.getImage() as string,
+        from,
+        dockerfile,
+      );
+      const fromAsName = args[2].getValue();
+      // support nested referral
```
mladkau: Do we have an example of nested referrals in the fixtures ?

lib/docker-file.ts
```diff
@@ -26,10 +26,45 @@ async function readDockerfileAndAnalyse(
...
+      names[fromAsName] = names[fromBaseName] || fromBaseName;
+    }
+    return names;
+  }, {});
   const from = dockerfile.getFROMs().pop();
```
karniwl: consider calling `from` `fromInstruction`

lib/docker-file.ts
karniwl: make sure when you extract to a function to name the function in a way that would ease the readability and clarity of the code flow

lib/docker-file.ts
karniwl: agree with @mladkau, try to check if we can utilize the library to do this

lib/docker-file.ts
```diff
@@ -26,10 +26,45 @@ async function readDockerfileAndAnalyse(
...
 async function analyseDockerfile(
   contents: string,
 ): Promise<DockerFileAnalysis | undefined> {
   const dockerfile = DockerfileParser.parse(contents);
+  // collect stages names
+  const stages = dockerfile.getFROMs().reduce((names, from) => {
+    const args = from.getArguments();
+    if (args.length > 2 && args[1].getValue() === "AS") {
+      const fromBaseName = resolveVariables(
+        from.getImage() as string,
```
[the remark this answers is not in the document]
moshikod: The parameter is now optional and is used when the string to be resolved is not the instruction test itself like in this case where we only need to resolve the stage image name

lib/docker-file.ts
moshikod: Couldn't find any equivalent, even for variable lookup.
I updated the regexp to support either `$variable_name` or `${variable_name}` forms.
https://docs.docker.com/v17.09/engine/reference/builder/#environment-replacement

lib/docker-file.ts
moshikod: Done, created a new function for base image retrieval https://github.com/snyk/snyk-docker-plugin/pull/98/files#diff-138b9629a01f1781af8ad6dff1ea75acR109

lib/docker-file.ts
moshikod: Done, thanks

lib/docker-file.ts
moshikod: The fixture https://github.com/snyk/snyk-docker-plugin/pull/98/files#diff-7f8eb961c7fb84a0b2e8d3e780b06bbeR1
The test https://github.com/snyk/snyk-docker-plugin/pull/98/files#diff-fc4c77fa4619a1c0b7c65906e79c7d99R51

lib/docker-file.ts
moshikod: Done, thanks
See https://github.com/snyk/snyk-docker-plugin/pull/98/files#diff-138b9629a01f1781af8ad6dff1ea75acR115

lib/docker-file.ts
moshikod: Done, thanks
See https://github.com/snyk/snyk-docker-plugin/pull/98/files#diff-138b9629a01f1781af8ad6dff1ea75acR114

lib/instruction-parser.ts
```diff
@@ -31,8 +34,17 @@ const installRegex = /\s*(rpm\s+-i|rpm\s+--install|apk\s+((--update|-u)\s+)*add|
  * arbitrary whitespace
  */
 function getPackagesFromRunInstructions(
-  runInstructions: string[],
+  dockerfile: Dockerfile,
 ): DockerFilePackages {
+  const runInstructions = dockerfile
+    .getInstructions()
+    .filter((instruction) => {
+      return instruction.getInstruction() === "RUN";
```
mladkau: No need for this return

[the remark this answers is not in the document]
moshikod: Done, thanks

moshikod: Done, thanks

PR #100, opened 2019-08-09
feat: Glob processor for Docker images.

mladkau:
The manifestGlobs which can be defined in the inspect() options are now interpreted. This PR adds the ability to scan a given docker images for specific files and return their content.


lib/docker.ts
```diff
@@ -71,19 +91,48 @@ class Docker {
...
+  ) {
+    const res: string[] = [];
+    const root = await this.lsSafe(path, recursive).then((output) =>
+      lsu.parseLsOutput(output.stdout),
+    );
+
+    lsu.iterateFiles(root, (f) => {
+      const filepath = fspath.join(f.path, f.name);
+      let exclude = false;
+      exclusionGlobs.forEach((g) => {
```
moshikod: Consider using a for loop and break when exclude is set to true

lib/ls-utils.ts
```diff
@@ -0,0 +1,89 @@
...
+        });
+      }
+    });
+
+  return res;
+}
+
+/**
+ * Return a subdirectory of a given DiscoveredDirectory. Create
+ * parent directories of the subdirectory if necessary.
```
moshikod: Maybe I miss something but where is a parent directory is created if necessary?

mladkau: Line 84/85. In the code it is a subdir as we go bottom-up however in the comment I write from the perspective of a subdirectory - so its parents are created (hence I write "parent directories of the subdirectory")...

lib/docker.ts
mladkau: I would prefer a function here. Normally you are right - we should break out of a loop as soon as possible. In this case the looping effort is insignificant. We'll only expect a few exclude patterns (<10) and the expensive minimatch is no longer executed once the exclude flag has been set. I prefer iterators to loops in JavaScript as variables have function score and not block scope - as such iterator functions are an effective way to avoid polluting the variable scope of the enclosing function with loop variables.

PR #101, opened 2019-08-16
fix: Limit the amount of manifest files which are read at once

mladkau:
Limit the number of manifest files which we try to read from a given Docker image. This is to protect our system from overload. Also adds a few fixtures for findGlobs testing.


test/lib/docker.test.ts
```diff
@@ -253,13 +253,77 @@ test("findGlobs", async (t) => {
     t.same(files, ["/pom.xml", "/dir3/file6.csproj"]);
   });
 
-  t.test("find globs using registry globs and fixture", async (t) => {
+  t.test(
+    "find globs on ghost app using node_modules exclude glob",
+    async (t) => {
+      stub.resolves({
+        stdout: readFileSync(getLSOutputFixture("ghost-app.txt")).toString(),
```
karniwl: is `txt` the suitable format here?

test/lib/docker.test.ts
```diff
@@ -253,13 +253,77 @@ test("findGlobs", async (t) => {
     t.same(files, ["/pom.xml", "/dir3/file6.csproj"]);
   });
 
-  t.test("find globs using registry globs and fixture", async (t) => {
+  t.test(
+    "find globs on ghost app using node_modules exclude glob",
```
karniwl: what's a `ghost app`?

test/lib/ls-utils.test.ts
```diff
@@ -212,3 +215,14 @@ test("parse ls output", async (t) => {
     });
   });
 });
+
+const getLSOutputFixture = (file: string) =>
```
karniwl: I think we try to avoid arrow functions, leave it if you prefer it this way but I think it's better to stick with regular functions.

test/lib/docker.test.ts
```diff
@@ -253,13 +253,77 @@ test("findGlobs", async (t) => {
...
+        getLSOutputFixture("debian-10.0-manifest-files.txt"),
+      ).toString(),
+    });
+
+    const files = await docker.findGlobs(["**/package.json"]);
+
+    t.same(files, ["/app/package.json"]);
+  });
+
+  t.test("find globs on centos", async (t) => {
```
karniwl: I suggest naming this test differently than the previous https://github.com/snyk/snyk-docker-plugin/pull/101/files#diff-36abe35061e760a3bbe9c05b3d215723R305
specifically with a verbose name that differentiates it with its purpose and test subject.

test/lib/docker.test.ts
```diff
@@ -253,13 +253,77 @@ test("findGlobs", async (t) => {
...
+    stub.resolves({
+      stdout: readFileSync(getLSOutputFixture("ubuntu-18.04.txt")).toString(),
+    });
+
+    const files = await docker.findGlobs([
+      "**/package.json",
+      "**/package-lock.json",
+    ]);
+
+    t.same(files, []);
```
karniwl: is this purposely expected to be empty? perhaps include this in the test name

test/lib/docker.test.ts
mladkau: A sample application see: https://hub.docker.com/_/ghost

test/lib/docker.test.ts
mladkau: yes, I would think so, this is the captured output of "ls -1apR /" from the running container.

test/lib/docker.test.ts
mladkau: oh dear, yes the names are very wrong! Also here I am testing Ubuntu and not Centos!

test/lib/docker.test.ts
mladkau: Yes and yes I will!

test/lib/ls-utils.test.ts
mladkau: That seems to make "prettier" upset :upside_down_face: 

```
ERROR: 219:28  only-arrow-functions  non-arrow functions are forbidden
```

PR #102, opened 2019-09-09
Fix app scan bug with undeterministical manifest files results

jurglic:
In app scanning we had an onerous issue where we would sometimes return empty results for manifest files, even though files were present on image. This was tracked down to be an issue of error handling inside `runSafe()`.
    
Therefore, update `runSafe()` not to nullify stdout/stderr result, even if some command resulted in error code. This shouldn't change existing clients' behaviour, except for the case of listing multiple files (`ls -R /`). In that case some files might become inaccessible during long operation and make the command return error code (alongside valid stdout results). In this we return stdout/stderr to the caller, so the result can be used.

### DIFF
--- a/lib/docker.ts
+++ b/lib/docker.ts
@@ -50,6 +50,10 @@ class Docker {
     this.optionsList = Docker.createOptionsList(options);
   }
 
+  /**
+   * Runs the command, catching any expected errors and returning them as normal
+   * stderr/stdout result.
+   */
   public async runSafe(cmd: string, args: string[] = []) {
     try {
       return await this.run(cmd, args);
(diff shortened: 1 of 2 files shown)


lib/docker.ts
```diff
@@ -50,17 +50,21 @@ class Docker {
...
+      return this.run(cmd, args);
     } catch (error) {
       const stderr: string = error.stderr;
       if (typeof stderr === "string") {
         if (
           stderr.indexOf("No such file") >= 0 ||
           stderr.indexOf("file not found") >= 0
         ) {
-          return { stdout: "", stderr: "" };
+          return { stdout: error.stdout, stderr };
```
mladkau: Nice solution - this code was before in catSafe. As you said the changed code shouldn't affect any current code.

mladkau:
I think you need to fix some unit tests...

PR #103, opened 2019-09-12
Feat/skopeo

ivanstanev:
- [ ] Ready for review

#### What does this PR do?

Allow the plugin to process files saved by `skopeo` in the docker-archive format (equivalent to `docker save`). The format is slightly different than `docker save` so it needs small adjustments.

This change also allows to pass a file path to a docker-archive existing on the filesystem, which allows to:
- skip calling `docker save` in the plugin
- skip the "safe size" checks and perform static scanning anyway

Call .inspect() with the `dockerArchivePath: string` option.

Team Runtime needs this for the `kubernetes-monitor`, we are removing the reliance on the Docker socket. We are using a tool called `skopeo` that works with container registries and is able to pull and save an image to disk in a `docker save`-style format (a.k.a. docker-archive).

With these changes we want to inform the plugin that the file is already present on disk, so it can be opened and inspected statically without using Docker.

- [Jira ticket RUN-450](https://snyksec.atlassian.net/browse/RUN-450)


.gitignore
```diff
@@ -2,6 +2,5 @@
 .idea/
 bin/
 npm-debug.log
-dist/
```
karniwl: is this on purpose? :)

karniwl: also all the files added from `dist/` as part the effect of this can be removed I think?

ivanstanev: @karniwl I added the `dist/` files because I was referencing the snyk-docker-plugin as a git branch inside package.json of kubernetes-monitor 😄For that to work there need to be some published/built assets in order to build our kubernetes-monitor container (I don't know why but that's the reality... npm install locally works but when inside the container/Dockerfile it just fails with missing files)

ivanstanev:
Closing; I had this opened only for visibility/comments 😄

PR #105, opened 2019-09-22
chore: refactor the apk analyzer to decouple docker from the logic

Shesekino:
- [ ] Ready for review

#### What does this PR do?

decouple the way the retrieve the apk data (apk DB file) which is done by running "docker run ..." from the logic that acts on the file's content.
this will allow reusing the logic to analyse a file acquired in a different way (static scanning) while still supporting the existing way.

right now this does absolutely nothing except for showing an initial proposal of how the rest of the analysers might look like.

https://snyksec.atlassian.net/browse/RUN-461

### DIFF
--- a/lib/analyzer/index.ts
+++ b/lib/analyzer/index.ts
@@ -7,7 +7,8 @@ import * as imageInspector from "./image-inspector";
 import * as osReleaseDetector from "./os-release-detector";
 import * as rpmAnalyzer from "./rpm-analyzer";
 
-import apkInput = require("./inputs/apk/docker");
+import apkInputDocker = require("./inputs/apk/docker");
+import apkInputStatic = require("./inputs/apk/static");
 import apkAnalyzer = require("./package-managers/apk");
 
 export { analyze };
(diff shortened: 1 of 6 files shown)


Shesekino:
excellent comment @ivanstanev.
would you please elaborate a bit more on the benefits of the current static scan approach that utilises callbacks (except for "that's how it currently is")?
right now I don't see a reason to keep it as-is, and I would favour a rewrite that would fit this flow - separate "get the data" from "analyse the data".
do you think it's impossible?

ivanstanev:
I see them as completely different approaches that we possibly can't combine (at least at first glance...) With the callback-based approach I think the benefit is that we process the whole image only once and the individual files N times as needed. With the "pull" approach, we are possibly looking into opening the image once for every file we'd like to find/inspect.
However, since we're using _streams_ the "pull"-based approach may be just as good -- we don't really process any files unless we need them! So you're correct and we'd need to refactor that code but ideally I am hoping to keep the great performance benefit of processing the image only once!

As for using the actual code, I fear it will be much more difficult to use or extend the callbacks style approach so I support your suggestion to split the logic into "get" and "process"

karniwl:
@Shesekino I understand this is a draft and I like the general idea of it, but I'm still missing the issue you raised about how to keep dynamic analysis live besides the static analysis. do you have a rough idea of how this refactor takes us there and what's the next step for me to understand the bigger picture?

lib/analyzer/index.ts
```diff
@@ -16,14 +19,23 @@ async function analyze(
   targetImage: string,
   dockerfileAnalysis?: dockerFile.DockerFileAnalysis,
   options?: DockerOptions,
+  analysisType?: string,
 ) {
+
+  let apkDb: string = '';
+  if (analysisType === 'static') {
+    apkDb = await apkInputStatic.getApkDbFileContent(targetImage);
```
Shesekino: here would go a component that encapsulates the "static getters" probably

lib/analyzer/index.ts
```diff
@@ -16,14 +19,23 @@ async function analyze(
...
+  } else { // assuming 'dynamic'
+    apkDb = await apkInputDocker.getApkDbFileContent(targetImage, options);
```
Shesekino: and its "dynamic getters" counterpart

PR #106, opened 2019-09-24
feat: add docker-archive unpacking

ivanstanev:
As part of static image analysis we need to process an image tar on the filesystem.
This PR adds functionality to allow unpacking of what is known as a docker-archive.
For the Runtime team this means processing docker-archive as produced by the Skopeo tool; the format there is slightly different but handled in this PR as well.
For every unpacked file of interest we can run a bunch of actions: useful for key binaries hashing in the future.

Also added unit tests with mocks/fixtures to verify it's working as intended.

- [x] Ready for review

Have a look at the `image-extractor.ts` file: the entry point is `extractFromTar`.

Part of static image analysis for @snyk/runtime 

[Jira ticket RUN-450](https://snyksec.atlassian.net/browse/RUN-450)
[Jira ticket RUN-462](https://snyksec.atlassian.net/browse/RUN-462)


lib/analyzer/image-extractor.ts
```diff
@@ -0,0 +1,198 @@
+import { createReadStream } from "fs";
```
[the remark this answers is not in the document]
Shesekino: this will also be a good opportunity to organise this file a bit and turn it into a dir -
move the types to `types.ts`
things we want to export to `index.ts`
if it grows too big, extract things to dedicated files, etc
what do you think?

lib/stream-utils.ts
```diff
@@ -0,0 +1,43 @@
...
+    });
+  });
+}
+
+/**
+ * Consume the data from the specified stream into a Buffer
+ * @param stream stream to cosume the data from
+ * @returns Buffer with the data consumed from the specified stream
+ */
+async function streamToBuffer(stream: Readable): Promise<Buffer> {
```
mladkau: Why do we have this function? If we calculate hashes then this can be done in chunks. There is no need to load the whole file into memory for this.

lib/analyzer/image-extractor.ts
```diff
@@ -0,0 +1,198 @@
...
+  fileStream: PassThrough,
+  extractActions: ExtractAction[],
+): Promise<Record<string, ProcessedFileContent> | undefined> {
+  const matchedActions = extractActions.filter((action) =>
+    minimatch(fileName, action.fileNamePattern, { dot: true }),
+  );
+
+  if (matchedActions.length > 0) {
+    const buffer = await streamToBuffer(fileStream);
+    return await applyCallbacks(matchedActions, buffer);
```
mladkau: I am not sure we should do this. This means that every file we find must be loaded into memory! These files can be quite big and this is not necessary for all operations. Yes a stream can only be read once but I don't think it is a good reason to load all files into memory.

ivanstanev:
@mladkau I think by answering I'll address both your comments! The current implementation is not yet thinking about key binaries hashing, we most likely would like to pass through the archive twice as there isn't a clean way to work with a stream by consuming it N times. This current approach is for finding individual specific files and applying some sort of "transformation" to it if necessary, something like having an array and applying .map() to it.

The idea is that we will have two phases to processing files: one is to _get_ or locate the file in the archive, the second step is to apply the different analyzers (e.g. apk, apt, rpm) once a file is obtained.

I would like to get the initial implementation here so we can start working with something and when it comes time to worry about key binaries we can adapt the code, for example we can pass the stream directly to the callbacks and the callback handler can pipe to its own processing logic or listen to 'data' events.

As to why we have the `streamToBuffer` function, it's just a default case to return the file as binary. We could have it do `streamToString` by default but not every file is a text file!

lib/analyzer/image-extractor.ts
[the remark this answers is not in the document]
ivanstanev: Just to note why this is resolved: we had a conversation about the organisation of the files and I'll update the PR with the new proposed layout

mladkau:
My main concern is memory consumption here. You might encounter some large files in a file system and loading them into memory is not a good idea. If you really cannot work with streams then do a stream to Temp file.

Shesekino:
@mladkau just out of curiosity, today (with the "dynamic" scanning) - aren't we running into the same memory consumption scenario?

the package manager manifest files would still be ```cat``` in their entirety to a variable, and the binaries would still be executed (which requires loading into memory, I guess?).

basically my question is - with this current implementation, are we simply not improving some aspects of the scan, or are we actually making some aspects worse, to your knowledge?

ivanstanev:
@mladkau I have updated the PR; we can now process the stream directly in the callback! This would help us do things like key binaries hashing efficiently.

lib/stream-utils.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @param stream stream to cosume the data from
+ * @returns Buffer with the data consumed from the specified stream
+ */
+export async function streamToBuffer(stream: Readable): Promise<Buffer> {
+  const chunks: Buffer[] = [];
+  return new Promise((resolve, reject) => {
+    stream.on("end", () => {
+      resolve(Buffer.concat(chunks));
+    });
+    stream.on("error", reject);
```
moshikod: Oops, this one is on me 😬
```suggestion
    stream.on("error", (error) => reject(error));
```

lib/stream-utils.ts
```diff
@@ -0,0 +1,41 @@
...
+export async function streamToString(
+  stream: Readable,
+  encoding: string = "utf8",
+): Promise<string> {
+  const chunks: string[] = [];
+  return new Promise((resolve, reject) => {
+    stream.on("end", () => {
+      resolve(chunks.join(""));
+    });
+    stream.on("error", reject);
```
moshikod: Oops, this one is on me 😬
```suggestion
    stream.on("error", (error) => reject(error));
```

lib/extractor/layer.ts
```diff
@@ -0,0 +1,141 @@
...
+
+function isManifestFile(name: string) {
+  return name === "manifest.json";
+}
+
+function isTarFile(name: string) {
+  // For both "docker save" and "skopeo copy" style archives the
+  // layers are represented as tar archives whose names end in .tar.
+  // For Docker this is "layer.tar", for Skopeo - "<sha256ofLayer>.tar".
+  return basename(name).indexOf(".tar") !== -1;
```
moshikod: Consider the following change
```suggestion
  return basename(name).endsWith(".tar");
```

lib/extractor/layer.ts
```diff
@@ -0,0 +1,141 @@
...
+    const manifest = JSON.parse(manifestFile);
+    return manifest[0].Layers;
+  });
+}
+
+function isManifestFile(name: string) {
+  return name === "manifest.json";
+}
+
+function isTarFile(name: string) {
```
moshikod: Consider the following change
```suggestion
function isTarFile(name: string): boolean {
```

lib/extractor/layer.ts
```diff
@@ -0,0 +1,141 @@
...
+ * Note: consumes the stream.
+ */
+function getLayersNames(stream: Readable): Promise<string[]> {
+  return streamToString(stream).then((manifestFile) => {
+    const manifest = JSON.parse(manifestFile);
+    return manifest[0].Layers;
+  });
+}
+
+function isManifestFile(name: string) {
```
moshikod: Consider the following change
```suggestion
function isManifestFile(name: string): boolean {
```

test/lib/stream-utils.test.ts
```diff
@@ -0,0 +1,20 @@
+import { createReadStream } from "fs";
+import { test } from "tap";
+import { streamToBuffer, streamToString } from "../../lib/stream-utils";
+
+test("stream-utils.streamToString()", async (t) => {
```
moshikod: Well done for adding this required test 👍

test/lib/stream-utils.test.ts
```diff
@@ -0,0 +1,20 @@
...
+  const fileStream = createReadStream(
+    "../fixtures/generic/small-sample-text.txt",
+  );
+  const fileContent = await streamToString(fileStream);
+  t.same(fileContent, "Hello, world!");
+});
+
+test("stream-utils.streamToBuffer()", async (t) => {
```
moshikod: Well done for adding this required test 👍

lib/extractor/layer.ts
```diff
@@ -0,0 +1,141 @@
...
+      stream.resume(); // auto drain the stream
+      next(); // ready for next entry
+    });
+
+    tarExtractor.on("finish", () => {
+      // all layer level entries read
+      resolve(result);
+    });
+
+    tarExtractor.on("error", reject);
```
moshikod: I guess this one is on me 😬
```suggestion
    tarExtractor.on("error", (error) => reject(error));
```

lib/extractor/layer.ts
```diff
@@ -0,0 +1,141 @@
...
+
+      stream.resume(); // auto drain the stream
+      next(); // ready for next entry
+    });
+
+    tarExtractor.on("finish", () => {
+      resolve(getLayersContent(layersNames, layers));
+    });
+
+    tarExtractor.on("error", reject);
```
moshikod: I guess this one is on me 😬
```suggestion
    tarExtractor.on("error", (error) => reject(error));
```

test/lib/stream-utils.test.ts
```diff
@@ -0,0 +1,20 @@
...
+  const fileStream = createReadStream(
+    "../fixtures/generic/small-sample-text.txt",
+  );
+  const fileBytes = await streamToBuffer(fileStream);
+  const expectedBytes = Buffer.from("Hello, world!", "utf8");
+  t.same(fileBytes, expectedBytes);
```
ivanstanev: This should probably use `deepEquals`

mladkau:
@Shesekino You are correct we do run into the same memory consumption errors with the current approach. One reason why we don't do proper key binary lookup in the moment. Getting static scanning to work properly in terms of architecture and performance would be a HUGE step forward!

@ivanstanev  giving the stream directly to the callback avoid loading everything into memory. But we need now to be careful as the stream can only be read once or?

ivanstanev:
@mladkau If we can trust both first two answers here https://stackoverflow.com/questions/51076356/multiple-listeners-reading-from-the-same-stream-in-nodejs then the approach I went with (running the tar file stream through PassThrough streams for every callback) should allow us to clone the stream and process it independently in every callback! If that really doesn't work then we'll have to introduce some code changes but I think it's a problem only for static analysis so far and it'll be on me to make it work 😄

PR #107, opened 2019-09-24
feat: accept static analysis flags

ivanstanev:
Add a mechanism to pass flags for static analysis.
These will be later used both as hints to use static analysis and also as configuration, for example locating the image to statically scan, whether to process key binaries, etc.

- [x] Ready for review

https://github.com/snyk/snyk-docker-plugin/compare/feat/static-scanning-flags?expand=1#diff-13b5b151431c7e7a17f273559ed212d5R28

Try to pass the expected flag, observe it in the debug logs!

To be used by @snyk/runtime when hinting to the plugin to use static analysis for images.

[Jira ticket RUN-442](https://snyksec.atlassian.net/browse/RUN-442)

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -4,6 +4,7 @@ import * as analyzer from "./analyzer";
 import { Docker, DockerOptions } from "./docker";
 import * as dockerFile from "./docker-file";
 import { buildResponse } from "./response-builder";
+import { StaticScanningOptions } from "./types";
 
 export { inspect, dockerFile };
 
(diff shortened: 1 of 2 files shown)


lib/index.ts
```diff
@@ -23,6 +24,16 @@ function inspect(root: string, targetFile?: string, options?: any) {
         manifestExcludeGlobs: options.manifestExcludeGlobs,
       }
     : {};
+
+  const staticScanningOptions: StaticScanningOptions =
+    options && options.staticScanningOptions
+      ? {
+          imagePath: options.staticScanningOptions.imagePath,
```
mladkau: Not sure why these are not part of the dockerOptions?

ivanstanev: Because they are not Docker options  😄 Arguably we've opted to name them "Docker options" whereas they are just any key-value options, and seeing that they're currently used with Docker flags in mind, I opted not to pollute the interface

mladkau: That makes sense. The name `dockerOptions` is a bit of a misnomer here and yes I would agree with @jurglic - let's handle this then in a separate function and not pollute inspect().

ivanstanev:
Thanks for the input @jurglic! As we discussed in our Zoom call, I will next integrate this PR as part of some other improvements I'm planning on opening for review so that you have a real context of how this is going to be used.

lib/index.ts
ivanstanev: Thanks! Will do that as part of a separate PR!

PR #108, opened 2019-09-27
Fix/decouple analyzers

ivanstanev:
- [x] Ready for review

#### What does this PR do?

Decouple the way we retrieve package manager files (e.g. APK DB file), which is currently done with Docker run, from the logic that processes the file content. This will allow reusing the logic of processing a file that was acquired in different ways (e.g. statically or dynamically with Docker) while still supporting the existing functionality.

Additionally introduced static stanning options which serve as hints that the caller wants a static analysis. The code now branches at the entrypoint "inspect()" for the two new paths: static analysis and dynamic (Docker) analysis.

I have decoupled the `analyzer.ts` into a `docker-analyzer.ts` with the idea that now there is also a `static-analyzer.ts`. The entrypoint is now `index.ts` where these two analyzers are exported.

Static scanning is needed by @snyk/runtime for the Kubernetes offering.

[Jira ticket RUN-461](https://snyksec.atlassian.net/browse/RUN-461)


lib/analyzer/package-manager/apk.ts
```diff
@@ -1,19 +1,14 @@
-import { Docker, DockerOptions } from "../docker";
-import { AnalyzerPkg } from "./types";
-export { analyze };
+import { AnalyzerPkg, AnalyzerResult } from "../types";
```
jurglic: Perhaps we could use plural folder form `lib/analyzer/package-managers`, like we used for `lib/inputs`?

lib/analyzer/types.d.ts
```diff
@@ -9,6 +9,12 @@ export interface AnalyzerPkg {
   AutoInstalled?: boolean;
 }
 
+export interface AnalyzerResult {
+  Image: string;
+  AnalyzeType: string;
+  Analysis: AnalyzerPkg[];
```
jurglic: I know names `AnalyzeType` and `Analysis` were here before, but could we improve them now, as we are adding more implementations of this interface?

It really confused me when reading the code. Perhaps we could name them `pkgManager` and `packages`, as they're already called elsewhere?

lib/analyzer/docker-analyzer.ts
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  if (!dockerAnalysis) {
+    return { installedPackages: [] };
+  }
+  const installedPackages = dockerAnalysis.Analysis.map((pkg) => pkg.Name);
+  let pkgManager = dockerAnalysis.AnalyzeType;
+  if (pkgManager) {
+    pkgManager = pkgManager.toLowerCase();
+  }
```
jurglic: Why do we need to lowercase the pkgManager? Can we get rid of this? :)

lib/analyzer/index.ts
```diff
@@ -1,75 +1,4 @@
-import * as Debug from "debug";
-import { DockerOptions } from "../docker";
-import * as dockerFile from "../docker-file";
-import * as apkAnalyzer from "./apk-analyzer";
-import * as aptAnalyzer from "./apt-analyzer";
-import * as binariesAnalyzer from "./binaries-analyzer";
-import * as imageInspector from "./image-inspector";
-import * as osReleaseDetector from "./os-release-detector";
-import * as rpmAnalyzer from "./rpm-analyzer";
+import { analyze as analyzeDynamically } from "./docker-analyzer";
```
jurglic: Maybe we can rename `docker-analyzer` to `dynamic-analyzer`?

lib/analyzer/docker-analyzer.ts
ivanstanev: I think it's used further in the processing logic... I'm just moving code around and I don't want to make this PR too big, would you be happy if I fixed this in a separate PR?

lib/analyzer/index.ts
ivanstanev: I think it would be rather new, right now everything in the dynamic part relates to Docker and I think it may be confusing if half of the things are called Docker and the other half "dynamic"... Maybe it would be better if "analyzeDynamically" got named as "analyzeWithDocker" here instead?

lib/index.ts
```diff
@@ -12,38 +13,78 @@ const debug = Debug("snyk");
...
+    ? {
+        host: options.host,
+        tlsVerify: options.tlsverify,
+        tlsCert: options.tlscert,
+        tlsCaCert: options.tlscacert,
+        tlsKey: options.tlskey,
+        manifestGlobs: options.manifestGlobs,
+        manifestExcludeGlobs: options.manifestExcludeGlobs,
+      }
+    : {};
```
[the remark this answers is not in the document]
ivanstanev: You're right, reverted the typing for now. I can address in another PR with strong type checks

lib/analyzer/docker-analyzer.ts
```diff
@@ -0,0 +1,87 @@
...
+  );
+  let binaries;
+  try {
+    binaries = await binariesAnalyzer.analyze(
+      targetImage,
+      installedPackages,
+      pkgManager,
+      options,
+    );
+  } catch (err) {
```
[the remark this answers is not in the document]
ivanstanev: Done!

lib/analyzer/package-manager/apk.ts
ivanstanev: Done!

lib/analyzer/types.d.ts
ivanstanev: As noted on Slack, I started doing this but it turned out bigger than expected. I am happy to do this in a separate PR!

lib/docker.ts
```diff
@@ -11,6 +11,8 @@ interface DockerOptions {
   tlsCert?: string;
   tlsCaCert?: string;
   tlsKey?: string;
+  manifestGlobs?: string;
+  manifestExcludeGlobs?: string;
```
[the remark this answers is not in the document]
ivanstanev: Removed!

lib/analyzer/index.ts
jurglic: Yep, makes sense 👍

PR #110, opened 2019-09-27
feat: static scanning for apk and apt

ivanstanev:
- [x] Ready for review

#### What does this PR do?

Enable static analysis for APK and APT package managers.
RPM and key binaries aree not yet handled, will be added in future PRs.

Other small improvements:
- unpacking docker-archive files returns them as string, not Buffer, because there is rarely a case where we would need the raw bytes of the file: even for key binaries hashing we would return a hex string
- os-release detector now exports "detect" as "detectDynamically"
- rename the Inspect type to DockerInspectOutput to better reflect it's a result returned by the "docker inspect" command
- when processing callbacks in docker-archive unpacking, await all of them with Promise.all() instead of forEach + individual await
- return the manifest file content when unpacking the docker-archive, so that we can return the image layers in the response

Static scanning is needed by @snyk/runtime for the Kubernetes offering.

[Jira ticket RUN-452](https://snyksec.atlassian.net/browse/RUN-452)


lib/inputs/os-release/static.ts
```diff
@@ -0,0 +1,92 @@
...
+  const osRelease = getContentIfExists(extractedLayers, getOsReleaseAction);
+  return osRelease !== undefined ? (osRelease as string) : "";
+}
+
+export function getLsbRelease(extractedLayers: ExtractedLayers): string {
+  const lsbRelease = getContentIfExists(extractedLayers, getLsbReleaseAction);
+  return lsbRelease !== undefined ? (lsbRelease as string) : "";
+}
+
+export function getDebianVersion(extractedLayers: ExtractedLayers): string {
```
talik077: shouldn't the function name be `getDebianRelease`? as `getAlpineRelease`, `getRedHatRelease`, etc?

lib/inputs/os-release/static.ts
```diff
@@ -0,0 +1,92 @@
...
+  const debianVersion = getContentIfExists(
+    extractedLayers,
+    getDebianVersionAction,
+  );
+  return debianVersion !== undefined ? (debianVersion as string) : "";
+}
+
+export function getAlpineRelease(extractedLayers: ExtractedLayers): string {
```
talik077: all the functions here have the same functionality and receive the same params, I think they only differ by the extract action. can't we have only one function e.g. `getDistributionRelease` and pass the distribution type as a param?

lib/inputs/runtime/docker.ts
```diff
@@ -0,0 +1,15 @@
+import { Docker, DockerOptions } from "../../docker";
+
+export function getRuntime(options: DockerOptions) {
+  return Docker.run(["version"], options)
+    .then((output) => {
```
talik077: why not using `async/await` instead?

lib/index.ts
```diff
@@ -50,12 +51,70 @@ function analyzeDynamically(
...
+      manifestFiles,
+      staticScanningOptions,
+    );
+  } catch (error) {
+    const analysisError = tryGetAnalysisError(error, targetImage);
+    throw analysisError;
+  }
+}
+
+function tryGetAnalysisError(error, targetImage): Error {
```
talik077: In order to reflect specific things that may go wrong, I would like to suggest instead creating a custom `AnalysisError` which inherit from JS Error class. coming from .NET and OOP approach, it sounds a bit more extendable. up to you :)

lib/index.ts
```diff
@@ -50,12 +51,70 @@ function analyzeDynamically(
...
+  if (typeof error === "string") {
```
talik077: I would check if error is an instance of `Error` and return it if it does. otherwise, perform your functionality

lib/inputs/os-release/static.ts
ivanstanev: I've kept the names as they exist right now, this change would need some input from the folks who did it like this the first time 🤔

lib/inputs/runtime/docker.ts
ivanstanev: I only moved the file, I haven't optimised it or changed it to something different 😄

lib/inputs/os-release/static.ts
ivanstanev: It's a very good improvement and it should also affect the `docker` side of this where it also does a chain of calls like these ones. But in order to abstract this I need another layer of indirection, something like a map that can process the input given the DistroType... I'll think about it and see if I can improve it!

lib/inputs/runtime/docker.ts
```diff
@@ -0,0 +1,15 @@
+import { Docker, DockerOptions } from "../../docker";
+
+export function getRuntime(options: DockerOptions) {
```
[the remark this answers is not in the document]
ivanstanev: I've only moved the function here, but I can improve it with a chore commit on top of everything so far

lib/index.ts
ivanstanev: I only wanted to abstract out the repeating piece of code, I also agree it's not ideal with the way the implementation either throws or returns an error for the caller to throw, but I've been trying not to touch/improve on the existing code while adding static analysis, which is the focus of the PR 😄Most of the bigger improvements I'm happy to introduce in a separate PR, there's also one other big task pending from a previous review...

lib/index.ts
ivanstanev: I can include this in a chore commit at the end!

lib/inputs/runtime/docker.ts
talik077: it's not necessary but we can give it some "runtime touch" 😺

lib/index.ts
talik077: since this change was introduced in one `feat` commit, It was a bit hard to detect which part of the code is new and which was a part of the existing code. so I reviewed it as a whole. Usually when I see such cases, if it's easy and quick to fix, I try to provide a small improvement. I'll add a code snippet down below, and if you'll find it useful you could use it as well :)

lib/inputs/os-release/static.ts
talik077: np 👍 let's not resolve it right now and see what @jurglic think :)?

lib/inputs/os-release/static.ts
```diff
@@ -0,0 +1,92 @@
+import { getContentIfExists } from "../../extractor";
+import { ExtractAction, ExtractedLayers } from "../../extractor/types";
+import { streamToString } from "../../stream-utils";
+
+export { getOsReleaseActions, getOsRelease };
+
+const getOsReleaseAction: ExtractAction = {
+  actionName: "os-release",
+  fileNamePattern: "/etc/os-release",
+  callback: (stream) => streamToString(stream),
```
jurglic: Where we set the callback, can we pass the function directly e.g.: `callback: streamToString,`?

lib/inputs/os-release/static.ts
jurglic: Yeah, makes more sense to me to call it `getDebianRelease`.. unfortunatelly don't know why was named like this in first place, maybe just because of rush? :)

ivanstanev: So the file we're inspecting is `/etc/debian_version` so maybe that's why it's called like that? I think I'll keep it so that it's consistent...

lib/inputs/os-release/static.ts
ivanstanev: Done!

lib/inputs/runtime/docker.ts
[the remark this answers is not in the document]
ivanstanev: Done!

lib/inputs/apk/static.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const getApkDbFileContentAction: ExtractAction = {
+  actionName: "apk-db",
+  fileNamePattern: "/lib/apk/db/installed",
+  callback: (stream) => streamToString(stream),
+};
+
+export function getApkDbFileContent(extractedLayers: ExtractedLayers): string {
+  const apkDb = getContentIfExists(extractedLayers, getApkDbFileContentAction);
+  return apkDb !== undefined ? (apkDb as string) : "";
```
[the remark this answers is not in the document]
ivanstanev: Done!

lib/inputs/os-release/static.ts
ivanstanev: I knew it looked suspicious to repeat it like that everywhere 😄Nice obvious fix, thanks; done

lib/index.ts
ivanstanev: I actually don't feel comfortable changing this, it's not a very well-tested path. I looked up some solutions with `instanceof` but I don't understand it really well so for both reasons I'll opt to leave the code as it is!

PR #112, opened 2019-10-08
Feat/rpm scanning

ivanstanev:
- [x] Ready for review

#### What does this PR do?

Implement RPM package scanning in static analysis mode.

Calls into a tool named "rpmdb" which must exist on the system.
The output of the tool is then formatted into what the existing RPM analyzer expects.

This change allows @snyk/runtime to completely stop relying on Docker.

Additional small changes:
- docker-archive extractor now supports reading the file content as Buffer/byte[]: this is necessary to correctly treat and process the RPM DB data -- it is a binary DB
- fix previously added tests to actually run (and fixed a small mistake in stream-utils.test.ts)

[Jira ticket RUN-468](https://snyksec.atlassian.net/browse/RUN-468)

### DIFF
--- a/lib/extractor/index.ts
+++ b/lib/extractor/index.ts
@@ -38,10 +38,34 @@ async function getDockerArchiveLayersAndManifest(
   };
 }
 
-function getContent(
+function isBufferType(type: string | Buffer): type is Buffer {
+  return (type as Buffer).buffer !== undefined;
+}
+
+function isStringType(type: string | Buffer): type is string {
+  return (type as string).substring !== undefined;
+}
+
+function getContentAsBuffer(
+  extractedLayers: ExtractedLayers,
+  extractAction: ExtractAction,
+): Buffer | undefined {
+  const content = getContent(extractedLayers, extractAction);
+  return content !== undefined && isBufferType(content) ? content : undefined;
+}
+
+function getContentAsString(
   extractedLayers: ExtractedLayers,
   extractAction: ExtractAction,
 ): string | undefined {
+  const content = getContent(extractedLayers, extractAction);
+  return content !== undefined && isStringType(content) ? content : undefined;
+}
+
+function getContent(
+  extractedLayers: ExtractedLayers,
+  extractAction: ExtractAction,
+): string | Buffer | undefined {
   const fileName = extractAction.fileNamePattern;
   return fileName in extractedLayers &&
     extractAction.actionName in extractedLayers[fileName]
@@ -49,4 +73,8 @@ function getContent(
     : undefined;
 }
 
-export { getDockerArchiveLayersAndManifest, getContent };
+export {
+  getDockerArchiveLayersAndManifest,
+  getContentAsString,
+  getContentAsBuffer,
+};
--- a/lib/inputs/rpm/static.ts
+++ b/lib/inputs/rpm/static.ts
@@ -1,0 +1,71 @@
+import { unlink, writeFile } from "fs";
+import { tmpdir } from "os";
+import { resolve as resolvePath } from "path";
+import { getContentAsBuffer } from "../../extractor";
+import { ExtractAction, ExtractedLayers } from "../../extractor/types";
+import { streamToBuffer } from "../../stream-utils";
+import { CmdOutput, execute } from "../../sub-process";
+
+export const getRpmDbFileContentAction: ExtractAction = {
+  actionName: "rpm-db",
+  fileNamePattern: "/var/lib/rpm/Packages",
+  callback: streamToBuffer,
+};
+
+export async function getRpmDbFileContent(
+  extractedLayers: ExtractedLayers,
+): Promise<string> {
+  const apkDb = getContentAsBuffer(extractedLayers, getRpmDbFileContentAction);
+  if (!apkDb) {
+    return "";
+  }
+
+  const filePath = generateTempFileName();
+  await writeToFile(filePath, apkDb);
+
+  try {
+    // This is the tool that is expected to be found on the system:
+    // https://github.com/snyk/go-rpmdb
+    const cmdOutput = await execute("rpmdb", ["-f", filePath]).catch(
+      handleError,
+    );
+    return cmdOutput.stdout || "";
+  } finally {
+    await removeFile(filePath);
+  }
+}
+
+function handleError(error): CmdOutput | never {
+  const stderr = error.stderr;
+  if (typeof stderr === "string" && stderr.indexOf("not found") >= 0) {
+    return { stdout: "", stderr: "" };
+  }
+  throw error;
+}
+
+function generateTempFileName(): string {
+  const tmpPath = tmpdir();
+  const randomFileName = Math.random().toString();
+
+  return resolvePath(tmpPath, randomFileName);
+}
+
+function writeToFile(filePath: string, apkDb: Buffer): Promise<void> {
+  return new Promise((resolve, reject) => {
+    writeFile(filePath, apkDb, { encoding: "binary" }, (err) => {
+      if (!err) {
+        resolve();
+      } else {
+        reject(err);
+      }
+    });
+  });
+}
+
+function removeFile(filePath: string): Promise<void> {
+  return new Promise((resolve) => {
+    unlink(filePath, () => {
+      resolve();
+    });
+  });
+}
--- a/lib/stream-utils.ts
+++ b/lib/stream-utils.ts
@@ -21,3 +21,21 @@ export async function streamToString(
     });
   });
 }
+
+/**
+ * Consume the data from the specified stream into a Buffer
+ * @param stream stream to cosume the data from
+ * @returns Buffer with the data consumed from the specified stream
+ */
+export async function streamToBuffer(stream: Readable): Promise<Buffer> {
+  const chunks: Buffer[] = [];
+  return new Promise((resolve, reject) => {
+    stream.on("end", () => {
+      resolve(Buffer.concat(chunks));
+    });
+    stream.on("error", (error) => reject(error));
+    stream.on("data", (chunk) => {
+      chunks.push(Buffer.from(chunk));
+    });
+  });
+}
--- a/lib/analyzer/static-analyzer.ts
+++ b/lib/analyzer/static-analyzer.ts
@@ -10,10 +10,15 @@ import {
   getExtFileContentAction,
 } from "../inputs/apt/static";
 import { getOsReleaseActions } from "../inputs/os-release/static";
+import {
+  getRpmDbFileContent,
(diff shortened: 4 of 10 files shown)


lib/inputs/rpm/static.ts
```diff
@@ -0,0 +1,88 @@
...
+  const apkDb = getContent(extractedLayers, getRpmDbFileContentAction);
+  if (!apkDb) {
+    return "";
+  }
+
+  const filePath = generateTempFileName();
+  await writeToFile(filePath, apkDb as Buffer);
+
+  try {
+    const cmdOutput = await execute("rpmdb", ["-f", filePath]).catch(
```
ivanstanev: Note this is where we call the tool: https://github.com/snyk/go-rpmdb

lib/inputs/rpm/static.ts
```diff
@@ -0,0 +1,88 @@
...
+    unlink(filePath, () => {
+      resolve();
+    });
+  });
+}
+
+/**
+ * Exported for testing
+ */
+export function formatRpmDbOutput(output: CmdOutput): string {
```
ivanstanev: Note how the RPM parser expects the output: https://github.com/snyk/snyk-docker-plugin/blob/master/lib/analyzer/package-managers/rpm.ts#L23
Hence we need to re-format what came out of the `go-rpmdb` tool into this expected format

lib/extractor/index.ts
```diff
@@ -41,7 +41,7 @@ async function getDockerArchiveLayersAndManifest(
 function getContent(
   extractedLayers: ExtractedLayers,
   extractAction: ExtractAction,
-): string | undefined {
+): string | Buffer | undefined {
```
moshikod: @ivanstanev, In which case the return type is of type `Buffer`?

ivanstanev: @moshikod When we read the RPM DB file content: it is a binary file!

lib/inputs/rpm/static.ts
```diff
@@ -0,0 +1,88 @@
...
+  const stderr = error.stderr;
+  if (typeof stderr === "string" && stderr.indexOf("not found") >= 0) {
+    return { stdout: "", stderr: "" };
+  }
+  throw error;
+}
+
+function generateTempFileName(): string {
+  const tmpPath = tmpdir();
+  const randomFileName = Math.random().toString();
```
shaimendel: does it make sense for the file name to somehow indicate the image name and file we try to pull? so if we see a file on the disk that is called rhel8_rpm_db_GUID we can know what is going on there. just a point to think about, not saying we should do it

lib/inputs/rpm/static.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  return resolvePath(tmpPath, randomFileName);
+}
+
+function writeToFile(filePath: string, apkDb: Buffer): Promise<string> {
```
shaimendel: shouldn't all of those functions be async? I know it doesn't matter but just for consistency

lib/inputs/rpm/static.ts
```diff
@@ -0,0 +1,88 @@
...
+      resolve();
+    });
+  });
+}
+
+/**
+ * Exported for testing
+ */
+export function formatRpmDbOutput(output: CmdOutput): string {
+  if (!output.stdout) {
```
shaimendel: shouldn't we throw here?

lib/inputs/rpm/static.ts
```diff
@@ -0,0 +1,88 @@
...
+    return "";
+  }
+
+  const packagesAsJson = output.stdout.split(EOL).filter((pkg) => pkg !== "");
+
+  const formattedPackages = packagesAsJson.map((jsonPackage) => {
+    const pkg: RpmDbPackage = JSON.parse(jsonPackage);
```
shaimendel: just wondering if there is a way for the rpmdb binary to print to stdout already in format we need here, so we don't need the JSON.parse at all (it is blocking the event loop).
Ideally we won't need this function at all, or most of it

test/lib/inputs/rpm/static.test.ts
```diff
@@ -0,0 +1,43 @@
+import { readFile } from "fs";
+import { join } from "path";
+import { test } from "tap";
+import { formatRpmDbOutput } from "../../../../lib/inputs/rpm/static";
+
+function getFixture(fixturePath: string) {
```
shaimendel: can add return type :)

lib/inputs/rpm/static.ts
moshikod: @ivanstanev, you may use the package `tmp` and `fileSync` to create and handle temporary files.

lib/stream-utils.ts
```diff
@@ -21,3 +21,21 @@ export async function streamToString(
...
+ */
+export async function streamToBuffer(stream: Readable): Promise<Buffer> {
+  const chunks: Buffer[] = [];
+  return new Promise((resolve, reject) => {
+    stream.on("end", () => {
+      resolve(Buffer.concat(chunks));
+    });
+    stream.on("error", (error) => reject(error));
+    stream.on("data", (chunk) => {
+      chunks.push(Buffer.from(chunk));
```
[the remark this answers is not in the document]
ivanstanev: The type is provided by TypeScript, in this case chunks is `Buffer[]`

lib/inputs/rpm/static.ts
```diff
@@ -0,0 +1,88 @@
...
+import { resolve as resolvePath } from "path";
+import { getContent } from "../../extractor";
+import { ExtractAction, ExtractedLayers } from "../../extractor/types";
+import { streamToBuffer } from "../../stream-utils";
+import { CmdOutput, execute } from "../../sub-process";
+import { RpmDbPackage } from "./types";
+
+export const getRpmDbFileContentAction: ExtractAction = {
+  actionName: "rpm-db",
+  fileNamePattern: "/var/lib/rpm/Packages",
```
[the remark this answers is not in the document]
ivanstanev: I tried with a variety of images and I believe so, yes.

lib/inputs/rpm/static.ts
```diff
@@ -0,0 +1,88 @@
...
+function handleError(error): CmdOutput | never {
+  const stderr = error.stderr;
+  if (typeof stderr === "string" && stderr.indexOf("not found") >= 0) {
+    return { stdout: "", stderr: "" };
+  }
+  throw error;
+}
+
+function generateTempFileName(): string {
+  const tmpPath = tmpdir();
```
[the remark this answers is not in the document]
ivanstanev: This is resolved specific to the OS, most likely `/tmp`

lib/inputs/rpm/static.ts
ivanstanev: @shaimendel I don't think we care about the package, we only need to extract it somewhere without clashing with an existing file. If something goes wrong we're not going to leave it lying around and consuming disk space

@moshikod That's a pretty big package for something close to what we want here but in 2 lines 😄 I really don't want to add more packages unless we really have to

lib/inputs/rpm/static.ts
ivanstanev: This one shouldn't be async because we directly return the promise, we don't have to await it because the caller of this function decides whether to await

lib/inputs/rpm/static.ts
ivanstanev: No because the caller of "get RPM DB content" decides whether to throw: by returning empty it can examine the other package manager outputs. Only when all are empty (no package manager detected) then the caller throws

lib/extractor/index.ts
```diff
@@ -38,15 +38,43 @@ async function getDockerArchiveLayersAndManifest(
...
+function isStringType(type: string | Buffer): type is string {
+  return (type as string).substring !== undefined;
+}
+
+function getContentAsBuffer(
+  extractedLayers: ExtractedLayers,
+  extractAction: ExtractAction,
+): Buffer | undefined {
+  const content = getContent(extractedLayers, extractAction);
+  return content !== undefined && isBufferType(content) ? content : undefined;
```
shaimendel: why do we return undefined instead of throwing an exception? I don't like returning undefined in case of errors. it looks like C programming :)

lib/inputs/rpm/static.ts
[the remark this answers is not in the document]
shaimendel: do we want to log in case it does not exist? (and it is an RPM image)

lib/inputs/rpm/static.ts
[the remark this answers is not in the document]
shaimendel: don't we want to enforce it is located inside our emptyDIr? that's the whole point of it right?

lib/inputs/rpm/static.ts
shaimendel: this needs to be changed after you get a new release of the rpm tool (and the dockerfile needs to be changed as well to the correct version)

lib/inputs/rpm/static.ts
ivanstanev: They aren't the same I think. To make the function async I have to rewrite the code as such:
```js
async function(): Promise<string> {
  const result: string = await new Promise((resolve) => {
    ... etc. etc.
    resolve(string_result);
  }
  return result;
}
```

Whereas right now it is...
```js
function(): Promise<string> {
  return new Promise((resolve) => {
    ... etc. etc.
  }
}

somewhere else:
const result: string = await function();
```

shaimendel: I think you can just add `async` to your function and it will work, both functions return `Promise<string>`

lib/inputs/rpm/static.ts
```diff
@@ -0,0 +1,71 @@
...
+  const randomFileName = Math.random().toString();
+
+  return resolvePath(tmpPath, randomFileName);
+}
+
+function writeToFile(filePath: string, apkDb: Buffer): Promise<string> {
```
shaimendel: why do you return a string in a write function? you don't use the return value anyway

lib/inputs/rpm/static.ts
```diff
@@ -0,0 +1,71 @@
...
+  const filePath = generateTempFileName();
+  await writeToFile(filePath, apkDb);
+
+  try {
+    // This is the tool that is expected to be found on the system:
+    // https://github.com/snyk/go-rpmdb
+    const cmdOutput = await execute("rpmdb", ["-f", filePath]).catch(
+      handleError,
+    );
+    return cmdOutput.stdout ? cmdOutput.stdout : "";
```
shaimendel: syntactic sugar: `return cmdOutput.stdout || ""`

PR #115, opened 2019-10-16
fix: layers annotation when a package version is specified in instruction

moshikod:
- [ ] Ready for review

#### What does this PR do?

- use the package name w/o version as the dockerfile instruction key
- support apt-get -y options after the command position
- update tests to verify the label hook content

Dockerfile instruction parser

test and monitor an image with a dockerfile that includes a `RUN` line similar to:
`RUN apk add --no-cache --update git=2.18.1-r0`

### DIFF
--- a/lib/instruction-parser.ts
+++ b/lib/instruction-parser.ts
@@ -23,7 +23,7 @@ interface DockerFileLayers {
 
 // Naive regex; see tests for cases
 // tslint:disable-next-line:max-line-length
-const installRegex = /\s*(rpm\s+-i|rpm\s+--install|apk\s+((--update|-u)\s+)*add|apt-get\s+((--assume-yes|--yes|-y)\s+)*install|apt\s+((--assume-yes|--yes|-y)\s+)*install|yum\s+install|aptitude\s+install)\s+/;
+const installRegex = /\s*(rpm\s+-i|rpm\s+--install|apk\s+((--update|-u|--no-cache)\s+)*add(\s+(--update|-u|--no-cache))*|apt-get\s+((--assume-yes|--yes|-y)\s+)*install(\s+(--assume-yes|--yes|-y))*|apt\s+((--assume-yes|--yes|-y)\s+)*install|yum\s+install|aptitude\s+install)\s+/;
 
 /*
  * This is fairly ugly because a single RUN could contain multiple install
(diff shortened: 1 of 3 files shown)


lib/instruction-parser.ts
```diff
@@ -23,7 +23,7 @@ interface DockerFileLayers {
 
 // Naive regex; see tests for cases
 // tslint:disable-next-line:max-line-length
-const installRegex = /\s*(rpm\s+-i|rpm\s+--install|apk\s+((--update|-u)\s+)*add|apt-get\s+((--assume-yes|--yes|-y)\s+)*install|apt\s+((--assume-yes|--yes|-y)\s+)*install|yum\s+install|aptitude\s+install)\s+/;
+const installRegex = /\s*(rpm\s+-i|rpm\s+--install|apk\s+((--update|-u)\s+)*add|apt-get\s+((--assume-yes|--yes|-y)\s+)*install(\s+(--assume-yes|--yes|-y))*|apt\s+((--assume-yes|--yes|-y)\s+)*install|yum\s+install|aptitude\s+install)\s+/;
```
mladkau: Err, what exactly is the change here? That's quite the "Where's Wally" game :smile:

moshikod: apt-get options after command position. for example: `apt-get install -y git`

lib/instruction-parser.ts
```diff
@@ -23,7 +23,7 @@ interface DockerFileLayers {
 
 // Naive regex; see tests for cases
 // tslint:disable-next-line:max-line-length
-const installRegex = /\s*(rpm\s+-i|rpm\s+--install|apk\s+((--update|-u)\s+)*add|apt-get\s+((--assume-yes|--yes|-y)\s+)*install|apt\s+((--assume-yes|--yes|-y)\s+)*install|yum\s+install|aptitude\s+install)\s+/;
+const installRegex = /\s*(rpm\s+-i|rpm\s+--install|apk\s+((--update|-u|--no-cache)\s+)*add(\s+(--update|-u|--no-cache))*|apt-get\s+((--assume-yes|--yes|-y)\s+)*install(\s+(--assume-yes|--yes|-y))*|apt\s+((--assume-yes|--yes|-y)\s+)*install|yum\s+install|aptitude\s+install)\s+/;
```
moshikod: apt-get options after command position. for example: `apt-get install -y git`
apk options after command position. for example `apk add --update --no-cache nodejs`

PR #117, opened 2019-10-28
wip excluding sys files

karniwl:
- [ ] Ready for review

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,4 +1,5 @@
 import * as Debug from "debug";
+import { flatten } from "lodash";
 import * as path from "path";
 import * as analyzer from "./analyzer";
 import { Docker, DockerOptions } from "./docker";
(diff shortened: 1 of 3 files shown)


lib/index.ts
```diff
@@ -215,15 +215,20 @@ async function getManifestFiles(targetImage: string, options?: any) {
...
 
   const globs = options.manifestGlobs as string[];
   const docker = new Docker(targetImage, options);
 
-  let files = await docker.findGlobs(globs, excludeGlobs);
+  let files = await docker.findGlobs(globs, excludeGlobs, "/", false);
+
+  files.forEach(async (dir) => {
+    const nextFiles = await docker.findGlobs(globs, excludeGlobs, dir, true);
+    files.concat(nextFiles);
```
mladkau: this is a NOP: `The concat() method is used to merge two or more arrays. This method does not change the existing arrays, but instead returns a new array.` see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat

karniwl: thanks! silly me :)

lib/index.ts
```diff
@@ -215,15 +216,30 @@ async function getManifestFiles(targetImage: string, options?: any) {
...
   }
 
   const globs = options.manifestGlobs as string[];
   const docker = new Docker(targetImage, options);
 
-  let files = await docker.findGlobs(globs, excludeGlobs);
+  const interFiles = await docker.findGlobs(globs, excludeGlobs, "/", false);
+
+  let files: string[] = flatten(
+    Promise.all(
```
mladkau: Nice one - this should reduce the overhead of running `ls` multiple times!

karniwl: but the test still fails :(

PR #118, opened 2019-11-05
feat: pass tmpdir path for static analysis

ivanstanev:
Do not infer a path where temporary files can be stored: this is problematic for the `kubernetes-monitor` because when the Kubernetes option readOnlyRootFilesystem is set, directories like /tmp become unusable.

Instead, the plugin now receives a path to a directory where it is safe to create temporary files.

- [x] Ready for review

### DIFF
--- a/lib/analyzer/static-analyzer.ts
+++ b/lib/analyzer/static-analyzer.ts
@@ -56,7 +56,7 @@ export async function analyze(
   ] = await Promise.all([
     getApkDbFileContent(archiveLayers),
     getAptDbFileContent(archiveLayers),
-    getRpmDbFileContent(archiveLayers),
+    getRpmDbFileContent(archiveLayers, options.tmpDirPath),
   ]);
 
   const osRelease = await osReleaseDetector
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -1,6 +1,11 @@
 export interface StaticAnalysisOptions {
   imagePath?: string;
   imageType?: ImageType;
+  /**
+   * Provide a path to a directory where the plugin can write temporary files.
+   * If unspecified, defaults to the environment's temporary directory path.
+   */
+  tmpDirPath?: string;
 }
 
 export enum ImageType {
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -126,6 +126,7 @@ function getStaticAnalysisOptions(options?: any): StaticAnalysisOptions {
     ? {
         imagePath: options.staticAnalysisOptions.imagePath,
         imageType: options.staticAnalysisOptions.imageType,
+        tmpDirPath: options.staticAnalysisOptions.tmpDirPath,
       }
     : {};
 }
--- a/lib/inputs/rpm/static.ts
+++ b/lib/inputs/rpm/static.ts
@@ -14,13 +14,14 @@ export const getRpmDbFileContentAction: ExtractAction = {
 
 export async function getRpmDbFileContent(
   extractedLayers: ExtractedLayers,
+  tmpDirPath?: string,
 ): Promise<string> {
   const apkDb = getContentAsBuffer(extractedLayers, getRpmDbFileContentAction);
   if (!apkDb) {
     return "";
   }
 
-  const filePath = generateTempFileName();
+  const filePath = generateTempFileName(tmpDirPath);
   await writeToFile(filePath, apkDb);
 
   try {
@@ -43,8 +44,11 @@ function handleError(error): CmdOutput | never {
   throw error;
 }
 
-function generateTempFileName(): string {
-  const tmpPath = tmpdir();
+/**
+ * Exported for testing
+ */
+export function generateTempFileName(tmpDirPath?: string): string {
+  const tmpPath = tmpDirPath || tmpdir();
   const randomFileName = Math.random().toString();
 
   return resolvePath(tmpPath, randomFileName);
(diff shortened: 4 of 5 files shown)


lib/types.ts
```diff
@@ -1,6 +1,7 @@
 export interface StaticAnalysisOptions {
   imagePath?: string;
   imageType?: ImageType;
+  tmpDirPath?: string;
```
Shesekino: why is this ```undefined```-able?

lib/analyzer/static-analyzer.ts
```diff
@@ -26,7 +26,11 @@ export async function analyze(
   targetImage: string,
   options: StaticAnalysisOptions,
 ) {
-  if (!options.imagePath || options.imageType === undefined) {
+  if (
+    !options.imagePath ||
+    options.imageType === undefined ||
+    options.tmpDirPath === undefined
```
mladkau: I think this is a bit harsh (making tmpDir a mandatory parameter) could we not make tmpdir() the default value and make tmpDirPath really optional?

test/system/static.test.ts
```diff
@@ -13,17 +14,20 @@ const getFixture = (fixturePath) =>
 test("static analysis builds the expected response", async (t) => {
   const thisIsJustAnImageIdentifierInStaticAnalysis = "node:doesnotexist";
   const dockerfile = undefined;
+  const tmpDirPath = tmpdir();
```
mladkau: Does this automatically cleanup the directory? If not we should remove it after the test is done.

lib/analyzer/static-analyzer.ts
Shesekino: I think this is a good separation of concerns for the docker plugin - why burden it with having to decide and manage a temporary dir?
it's a logical component tasked with scanning an image, let its user (in this case, the Kubernetes-Monitor) provide it with storage.
this way it also won't create files in places unknown to its user.
what do you think?

test/system/static.test.ts
ivanstanev: The files created by the plugin are cleaned up once done with them: https://github.com/snyk/snyk-docker-plugin/blob/master/lib/inputs/rpm/static.ts#L33-L35
Note that `tmpdir()` does not create a directory but returns the path to the OS temporary directory (usually /tmp)

lib/analyzer/static-analyzer.ts
ivanstanev: I'd prefer the caller to be explicit about their choice of directory, there's less variations happening internally in the implementation... I'm not strongly objecting having a default but since we're currently the only users of static analysis I want us to be very explicit about this choice since forgetting to set a path can really break our scanning ability

lib/types.ts
ivanstanev: To better capture the intent that when the plugin is called, these options may not actually be set... There is no guarantee that they'll be defined unless we strictly check that

Shesekino: but you strictly check that 😂

Shesekino: this should communicate what's expected (this does not mean we shouldn't strictly check them in addition)

ivanstanev: When I first grab the options none of the values are set, that's why we do the undefined checks. I could better capture the intent by initially marking the type as Partial<StaticAnalysisOptions>, then after the checks we would know it's of type StaticAnalysisOptions, WDYT?

test/system/static.test.ts
mladkau: As long as the test leaves nothing behind it is ok ...

lib/analyzer/static-analyzer.ts
shaimendel: first, this is a "breaking" change, I think it is better to have the default tmp dir in case it is not passed.
additionally it don't think it is reasonable to enforce the users to pass it, because most of them do not care about it (actually, we are the only ones who care)

mladkau: Pushing decisions to the calling code is a common but not always good pattern. If it does have severe consequences for you then yes we could make it explicit. But please do describe then what is actually expected and what do we do with it (i.e. files are left behind in the folder, folder must be empty, etc.)

Shesekino: @mladkau fair enough, makes sense :) I also think we should add the description you suggested regardless of whether or not it's optional.

@shaimendel why not keep it mandatory and let the k8s-monitor upgrade to the latest docker-plugin only when it's ready to pass the new arguments? it's not a living service, you don't have to upgrade your dependency if you don't want to.

> additionally it don't think it is reasonable to enforce the users to pass it, because most of them do not care about it
aren't we the only users for static scanning?

mladkau: @Shesekino In the moment you are the only users but @snyk/magma will very soon follow. Not having static scanning is a big gap for us.

lib/types.ts
Shesekino: I think it's alright?
not too familiar with TS conventions.
I think that as long as ```StaticAnalysisOptions``` is true to its expectations, it's fine

lib/analyzer/static-analyzer.ts
ivanstanev: Thanks for your input! I've made it optional, I think we can ensure we always pass the path from our side

lib/types.ts
ivanstanev: Considering the path is now optional, I've not made any changes. Do you think it's good to go in its current state?

Shesekino: I would raise the issue of the other two parameters being ```?``` but it's not part of this context.
so yes, I think it's great :)

ivanstanev: Noted 😄

PR #123, opened 2019-12-06
chore: typescript 3.7 breaks dockerfile-ast

FauxFaux:
https://github.com/rcjsuen/dockerfile-ast/pull/60

For now, pin us at typescript 3.6.


FauxFaux:
..and apply the new prettier, or the build also fails.

PR #125, opened 2019-12-15
static scanning: support situations where /etc/os-release does not ex…

shaimendel:
support situations where /etc/os-release does not exist or links to /usr/lib/os-release.
according to https://www.freedesktop.org/software/systemd/man/os-release.html:
"The file /etc/os-release takes precedence over /usr/lib/os-release. Applications should check for the former, and exclusively use its data if it exists, and only fall back to /usr/lib/os-release if it is missing."

- [x] Ready for review

#### What does this PR do?
explained in the commit comment


lib/analyzer/os-release/static.ts
```diff
@@ -18,7 +18,14 @@ export async function detect(
     getOsRelease(extractedLayers, OsReleaseFilePath.Linux),
   );
 
-  // First generic fallback
+  // Generic fallback
```
karniwl: both generic fallbacks? maybe edit the comments to differentiate between both? (l21, l28)

PR #130, opened 2020-01-20
fix: Make safeLs more tolerant for expected errors

jurglic:
Recursive listing of files (safeLs) is used for finding manifest files. Make this process more robust by handling expected errors/warning, such as permission denied. This allows for scanning to continue even if some (system) folders are inaccessible.

There's more we could do here for better error handling, but stopping sooner because upcoming switch to static scanning.

### DIFF
--- a/lib/docker.ts
+++ b/lib/docker.ts
@@ -54,16 +54,18 @@ class Docker {
    * Runs the command, catching any expected errors and returning them as normal
    * stderr/stdout result.
    */
-  public async runSafe(cmd: string, args: string[] = []) {
+  public async runSafe(
+    cmd: string,
+    args: string[] = [],
+    // no error is thrown if any of listed errors is found in stderr
+    ignoreErrors: string[] = ["No such file", "file not found"],
+  ) {
     try {
       return await this.run(cmd, args);
     } catch (error) {
       const stderr: string = error.stderr;
       if (typeof stderr === "string") {
-        if (
-          stderr.indexOf("No such file") >= 0 ||
-          stderr.indexOf("file not found") >= 0
-        ) {
+        if (ignoreErrors.some((errMsg) => stderr.indexOf(errMsg) >= 0)) {
           return { stdout: error.stdout, stderr };
         }
       }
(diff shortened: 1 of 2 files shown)


lib/docker.ts
```diff
@@ -54,16 +54,18 @@ class Docker {
...
     try {
       return await this.run(cmd, args);
     } catch (error) {
       const stderr: string = error.stderr;
       if (typeof stderr === "string") {
-        if (
-          stderr.indexOf("No such file") >= 0 ||
-          stderr.indexOf("file not found") >= 0
-        ) {
+        if (ignoreErrors.some((errMsg) => stderr.indexOf(errMsg) >= 0)) {
```
karniwl: any reason not to use `ignoreErrors.includes(errMsg)`?

lib/docker.ts
```diff
@@ -103,7 +105,14 @@ class Docker {
     if (recursive) {
       params += "R";
     }
-    return this.runSafe("ls", [params, path]);
+
+    const ignoreErrors = [
```
moshikod:
```suggestion
    const ignoreErrors: string[] = [
```

lib/docker.ts
```diff
@@ -54,16 +54,18 @@ class Docker {
    * Runs the command, catching any expected errors and returning them as normal
    * stderr/stdout result.
    */
-  public async runSafe(cmd: string, args: string[] = []) {
+  public async runSafe(
+    cmd: string,
+    args: string[] = [],
+    // no error is thrown if any of listed errors is found in stderr
+    ignoreErrors: string[] = ["No such file", "file not found"],
```
moshikod: Consider declaring and using a constant for the default array value

test/lib/docker.test.ts
```diff
@@ -133,6 +133,18 @@ test("safeLs", async (t) => {
     );
   });
 
+  t.test("directory found, some files inaccessable", async (t) => {
```
moshikod: 👏

lib/docker.ts
jurglic: Here I actually prefer the former, because the syntax is less noisy, and hence the code more readable. Other than that they're identical compile and run time.

lib/docker.ts
jurglic: Here I'd err to the side of simpler explicit code at the cost of slight duplication (errors handling logic is a bit surprising), but in any way I hope this code will be ditched soon for static analysis! :)

PR #135, opened 2020-02-04
Feat/pull image if not local

RotemS:
- [x] Ready for review

#### What does this PR do?
Modified to try pulling before inspecting an image when using inspect(). Changed some of the errors to better reflect the new flow. 

Try to inspect an image that doesn't exist locally. If it's possible to pull the image from docker hub, inspect should succeed. 

https://github.com/snyk/snyk/issues/595

### DIFF
--- a/lib/analyzer/docker-analyzer.ts
+++ b/lib/analyzer/docker-analyzer.ts
@@ -19,6 +19,16 @@ export async function analyze(
   dockerfileAnalysis?: dockerFile.DockerFileAnalysis,
   options?: DockerOptions,
 ) {
+  try {
+    await imageInspector.pullIfNotLocal(targetImage, options);
+  } catch (error) {
+    debug(`Error while running analyzer: '${error}'`);
+    throw new Error(
+      "Docker error: image was not found locally and pulling failed: " +
+        targetImage,
+    );
+  }
+
   const [imageInspection, osRelease] = await Promise.all([
     imageInspector.detect(targetImage, options),
     osReleaseDetector.detectDynamically(
@@ -38,31 +48,40 @@ export async function analyze(
     rpmInputDocker.getRpmDbFileContent(targetImage, options),
   ]);
 
-  const results = await Promise.all([
-    apkAnalyzer.analyze(targetImage, apkDbFileContent),
-    aptAnalyzer.analyze(targetImage, aptDbFileContent),
-    rpmAnalyzer.analyze(targetImage, rpmDbFileContent),
-  ]).catch((err) => {
-    debug(`Error while running analyzer: '${err.stderr}'`);
+  let pkgManagerAnalysis;
+  try {
+    pkgManagerAnalysis = await Promise.all([
+      apkAnalyzer.analyze(targetImage, apkDbFileContent),
+      aptAnalyzer.analyze(targetImage, aptDbFileContent),
+      rpmAnalyzer.analyze(targetImage, rpmDbFileContent),
+    ]);
+  } catch (error) {
+    debug(`Error while running analyzer: '${error.stderr}'`);
     throw new Error("Failed to detect installed OS packages");
-  });
+  }
 
   const { installedPackages, pkgManager } = getInstalledPackages(
-    results as any[],
+    pkgManagerAnalysis as any[],
   );
 
-  const binaries = await binariesAnalyzer
-    .analyze(targetImage, installedPackages, pkgManager, options)
-    .catch((err) => {
-      debug(`Error while running binaries analyzer: '${err}'`);
-      throw new Error("Failed to detect binaries versions");
-    });
+  let binarisAnalysis;
+  try {
+    binarisAnalysis = await binariesAnalyzer.analyze(
+      targetImage,
+      installedPackages,
+      pkgManager,
+      options,
+    );
+  } catch (error) {
+    debug(`Error while running binaries analyzer: '${error}'`);
+    throw new Error("Failed to detect binaries versions");
+  }
 
   return {
     imageId: imageInspection.Id,
     osRelease,
-    results,
-    binaries,
+    results: pkgManagerAnalysis,
+    binaries: binarisAnalysis,
     imageLayers: imageInspection.RootFS && imageInspection.RootFS.Layers,
   };
 }
--- a/lib/analyzer/image-inspector.ts
+++ b/lib/analyzer/image-inspector.ts
@@ -1,23 +1,24 @@
 import { Docker, DockerOptions } from "../docker";
 import { DockerInspectOutput } from "./types";
 
-export { detect };
+export { detect, pullIfNotLocal };
 
 async function detect(
   targetImage: string,
   options?: DockerOptions,
 ): Promise<DockerInspectOutput> {
+  const docker = new Docker(targetImage, options);
+  const info = await docker.inspectImage(targetImage);
+  return JSON.parse(info.stdout)[0];
+}
+
+async function pullIfNotLocal(targetImage: string, options?: DockerOptions) {
+  const docker = new Docker(targetImage);
   try {
-    const info = await new Docker(targetImage, options).inspectImage(
-      targetImage,
-    );
-    return JSON.parse(info.stdout)[0];
-  } catch (error) {
-    if (error.stderr.includes("No such object")) {
-      throw new Error(
-        `Docker error: image was not found locally: ${targetImage}`,
-      );
-    }
-    throw new Error(`Docker error: ${error.stderr}`);
+    await docker.inspectImage(targetImage);
+    return;
+  } catch (err) {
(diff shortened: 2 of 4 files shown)


lib/analyzer/docker-analyzer.ts
```diff
@@ -19,14 +19,12 @@ export async function analyze(
...
 ) {
-  const [imageInspection, osRelease] = await Promise.all([
-    imageInspector.detect(targetImage, options),
-    osReleaseDetector.detectDynamically(
-      targetImage,
-      dockerfileAnalysis,
-      options,
-    ),
-  ]);
+  const imageInspection = await imageInspector.detect(targetImage, options);
```
RotemS: Refactored to make sure that imageInspector throws first

lib/analyzer/docker-analyzer.ts
```diff
@@ -19,14 +19,12 @@ export async function analyze(
...
+  const osRelease = await osReleaseDetector.detectDynamically(
+    targetImage,
+    dockerfileAnalysis,
+    options,
+  );
```
RotemS: Removed the promise.all to make sure that imageInspector throws first. osReleaseDetector runs `docker run` which will throw if the image couldn't be pulled. For the sake of throwing an error that reflects that we tried pulling for inspect, changed this.

test/system/system.test.ts
```diff
@@ -401,14 +401,10 @@ test(
...
     const imgName = "redis";
     const imgTag = "3.2.11-alpine";
-    const img = imgName + ":" + imgTag;
     const dockerFileLocation = getDockerfileFixturePath("redis");
     const hostAndImgName = "localhost:5000/" + imgName;
     const hostAndImg = hostAndImgName + ":" + imgTag;
 
-    await dockerPull(t, img);
-    await dockerTag(t, img, hostAndImg);
-    await dockerGetImageId(t, hostAndImg);
```
RotemS: Since inspect() now checks first if the image exists, these were redundant.

lib/docker.ts
```diff
@@ -89,6 +89,33 @@ class Docker {
...
+    } catch (err) {
+      return false;
+    }
+  }
+
+  public async inspectWithPull(targetImage: string) {
+    const isImageLocal = await this.isImageLocal(targetImage);
+    if (!isImageLocal) {
+      try {
+        await this.pull();
```
mladkau: The called function has no parameters and only a single line of code - not sure we gain anything here. Is there a reason why the single line of code should be in a function?

mladkau: ... but thinking about it, it would make sense to move the surrounding try catch block into the function.

lib/docker.ts
```diff
@@ -89,6 +89,33 @@ class Docker {
...
+    try {
+      await this.inspectImage(targetImage);
+      return true;
+    } catch (err) {
+      return false;
+    }
+  }
+
+  public async inspectWithPull(targetImage: string) {
+    const isImageLocal = await this.isImageLocal(targetImage);
```
mladkau: Hmm, also this is calling a function with more a less a single line of code. I think I would write something like:
```
try {
  this.inspectImage(...
} catch() {
  this.pull()
}
return this.inspectImage(...
```
I suggest creating a `isImageLocal` function (with only a single line of actual code) only if there are at least two callers.

lib/analyzer/docker-analyzer.ts
mladkau: It might make sense to add a comment here. This is some important knowledge which is not obvious from the code. Just to prevent someone refactoring the code again into a Promise.all :smile:

test/system/system.test.ts
```diff
@@ -430,6 +430,18 @@ test(
   },
 );
 
+test("inspect image that doesn't exist locally but pullable from docker hub", async (t) => {
+  const img = "alpine:3.10";
```
mladkau: Does this test assume I don't have alpine:3.10 in my local registry?

karniwl:
@RotemS @mladkau let's get a review from @snyk/runtime here

test/system/system.test.ts
```diff
@@ -39,7 +39,7 @@ test("inspect an image that does not exist", (t) => {
   return plugin.inspect("not-here:latest").catch((err) => {
     t.same(
       err.message,
-      "Docker error: image was not found locally: not-here:latest",
+      "Docker error: image not-here:latest was not found locally, failed to pull",
```
shaimendel: just making sure that this error change doesn't break any of your dashbaords

lib/docker.ts
RotemS: The idea was to keep it the same as the rest of the functions in Docker, in terms of returned type etc.
Do you think it's not worth it?

lib/docker.ts
RotemS: Wanted to have a clearer error reflected, but I see your point :)

test/system/system.test.ts
RotemS: If I understood you correctly, it runs `docker rmi` hoping that it would indeed remove it locally

lib/analyzer/docker-analyzer.ts
```diff
@@ -19,6 +19,15 @@ export async function analyze(
   dockerfileAnalysis?: dockerFile.DockerFileAnalysis,
   options?: DockerOptions,
 ) {
+  await imageInspector.pullIfNotLocal(targetImage, options).catch((err) => {
```
shaimendel: why do you use `.catch` instead of wrapping with regular try-catch? (in the original PR the error was `No such object`)

lib/analyzer/docker-analyzer.ts
```diff
@@ -19,6 +19,15 @@ export async function analyze(
...
+    if (err.stderr.includes("pull access denied")) {
```
shaimendel: are you sure that this is the only case you want to throw at? sounds weird

lib/analyzer/image-inspector.ts
```diff
@@ -1,23 +1,22 @@
...
-  } catch (error) {
-    if (error.stderr.includes("No such object")) {
-      throw new Error(
-        `Docker error: image was not found locally: ${targetImage}`,
-      );
-    }
-    throw new Error(`Docker error: ${error.stderr}`);
+    await docker.inspectImage(targetImage);
+  } catch (err) {
+    await docker.pull(targetImage);
```
shaimendel: not critical, a pure matter of style: I don't like actual logic in `catch` phrases. I would rather add a `return` in the end of the `try` and leave the `catch` empty, like:
```
try {
    await docker.inspectImage(targetImage);
    return;
} catch (err) { }

await docker.pull(targetImage);
```

shaimendel: and also, if you don't use err so don't define it :)

lib/analyzer/docker-analyzer.ts
RotemS: Actually only did that because that's used in the rest of the file

lib/analyzer/docker-analyzer.ts
```diff
@@ -19,6 +19,18 @@ export async function analyze(
   dockerfileAnalysis?: dockerFile.DockerFileAnalysis,
   options?: DockerOptions,
 ) {
+  try {
+    await imageInspector.pullIfNotLocal(targetImage, options);
+  } catch (error) {
+    if (error.stderr.includes("pull access denied")) {
+      throw new Error(
+        "Docker error: image was not found locally and pulling failed: " +
```
shaimendel: why does the `pull access denied` error gets "special treatment" and you actually convert the error to this string, and the other cases you throw as is? I assume that either you always throw the error as-is or throw this "Docker error" instead

lib/analyzer/image-inspector.ts
RotemS: 👍 
About the not declaring, is that possible? I'm getting an error if I try to.

dkontorovskyy: Just put a comment in `catch` block. Smth like `// do nothing`

lib/analyzer/image-inspector.ts
```diff
@@ -1,23 +1,24 @@
...
-        `Docker error: image was not found locally: ${targetImage}`,
-      );
-    }
-    throw new Error(`Docker error: ${error.stderr}`);
+    await docker.inspectImage(targetImage);
+    return;
+  } catch (err) {
+    // image doesn't exist locally
   }
+  await docker.pull(targetImage);
```
mladkau: Can this not be pulled into the catch block while removing the final return in the try block?

RotemS: I'm not sure I understand, do you mean removing the `return;`? 
I think maybe only if it was separated into another function that catches the error from this one or used directly in the calling function?

PR #137, opened 2020-02-19
Chore/types

Shesekino:
- [x] Ready for review

#### What does this PR do?

some cosmetic changes to prepare improvements in the image static scanning capabilities.
each commit has plenty of information but the gist is:
1. bump typescript version
2. refactor a single function to ```async```/```await```
3. add some types where possible and simple. it's a first step, still relying on many ```any```'s, but it only surfaces these ```any``` without introducing new ones.

each commit separately.

as the changes are only cosmetic, I'm hoping our automated tests suffice here.
open to suggestions on how to manually test it!

https://snyk.slack.com/archives/CS76EM7UM/p1582131635005500

https://snyksec.atlassian.net/browse/RUN-697

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -5,7 +5,7 @@ import { Docker, DockerOptions } from "./docker";
 import * as dockerFile from "./docker-file";
 import { getRuntime } from "./inputs/runtime/docker";
 import { buildResponse } from "./response-builder";
-import { StaticAnalysisOptions } from "./types";
+import { ManifestFile, PluginResponse, StaticAnalysisOptions } from "./types";
 
 export { inspect, dockerFile };
 
@@ -35,7 +35,7 @@ async function analyzeDynamically(
   targetImage: string,
   dockerfileAnalysis: dockerFile.DockerFileAnalysis | undefined,
   analysisOptions: any,
-) {
+): Promise<PluginResponse> {
   const [runtime, dependencies, manifestFiles] = await Promise.all([
     getRuntime(analysisOptions),
     getDependencies(targetImage, dockerfileAnalysis, analysisOptions),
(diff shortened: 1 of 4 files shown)


lib/index.ts
```diff
@@ -31,27 +31,30 @@ function inspect(root: string, targetFile?: string, options?: any) {
...
-      analysisOptions,
-    );
-  });
+  ]);
+
+  return buildResponse(
+    runtime,
+    dependencies,
+    dockerfileAnalysis,
+    manifestFiles!, // bug in typescript wrongly adds `undefined`
```
Shesekino: https://github.com/microsoft/TypeScript/issues/33752

lib/types.ts
```diff
@@ -21,3 +21,23 @@ export enum OsReleaseFilePath {
...
+export interface ManifestFile {
+  name: string;
+  path: string;
+  contents: string;
+}
+
+export interface PluginMetadata {
+  name: string;
+  runtime: string | undefined;
+  packageManager: any; // TODO
```
mladkau: Does the TODO add anything here? I would remove it and make the type more concrete when it becomes necessary. If you think the TODO should be there then we should at least describe what exactly needs doing :smiley:

Shesekino: it makes it a bit more obvious that we're unhappy with the ```any``` type here, and we intend fixing it soon.
I find it self explanatory that it refers to the ```any```, because it really sticks out to me.
what do you think?

mladkau: I don't mind really but have a dislike for TODO's in the code which don't exactly say what needs doing. For you in the moment it is clear that they refer to the any type but it could also mean that some types are obsolete or need further differentiation or maybe the name is just not ideal in the moment. The meaning becomes ambiguous over time.

Instead of TODOs in the code I would track this as a technical debt task in Jira. This is a good example of technical debt which needs addressing at some point. In magma we have the "Tech-debt & maintenance" epic for this.

Shesekino: a great explanation, I'm removing the TODOs. thanks!

PR #139, opened 2020-02-23
Chore/binaries

Shesekino:
- [x] Ready for review

#### What does this PR do?

enhance the static scanning capabilities with calculating hashes for files and returning them as part of the key-binaries results.

each commit separately.

https://snyksec.atlassian.net/browse/RUN-702
https://snyksec.atlassian.net/browse/RUN-703
https://snyksec.atlassian.net/browse/RUN-704


lib/stream-utils.ts
```diff
@@ -1,5 +1,9 @@
+import * as crypto from "crypto";
 import { Readable } from "stream";
 
+const HASH_ALGORITHM = "sha256"; // TODO algorithm?
```
Shesekino: the algorithm used here needs to match the one used by the key-binary lookup service, but I would rather both of them to be ```sha1``` instead of ```sha256```` probably

lib/inputs/binaries/static/index.ts
```diff
@@ -1 +1,21 @@
...
+};
+
+export function getBinariesHashes(extractedLayers: ExtractedLayers): string[] {
+  const hashes: string[] = [];
+  for (const fileName of Object.keys(extractedLayers)) {
+    for (const actionName of Object.keys(extractedLayers[fileName])) {
+      if (actionName !== getNodeBinariesFileContentAction.actionName) {
+        continue;
+      }
+      hashes[fileName] = extractedLayers[fileName][actionName];
```
shaimendel: I assume that's the exact expected format :)

lib/inputs/binaries/static/index.ts
```diff
@@ -1 +1,21 @@
...
+import { streamToHash } from "../../../stream-utils";
+
+export const getNodeBinariesFileContentAction: ExtractAction = {
+  actionName: "node",
+  fileNamePattern: "**/*/node",
+  callback: streamToHash,
+};
+
+export function getBinariesHashes(extractedLayers: ExtractedLayers): string[] {
+  const hashes: string[] = [];
```
shaimendel: is it an array or a map? looks like you use it as a map

lib/inputs/binaries/static/index.ts
Shesekino: what expected format?

lib/inputs/binaries/static/index.ts
Shesekino: it is an array, being used like a map - that's a bug :)

lib/inputs/binaries/static/index.ts
shaimendel: the format that registry expects in the end?

Shesekino: if you're talking about Registry's expectations in the k8s flow - these are not set yet, and still need to be defined. I would rely on the k8s-agent to make the necessary adjustments.
if you're talking about Registry's expectations in the CLI flow - these are not used yet, and I'm more than happy to make adjustments there as we go instead of being fixated on what exists at the moment

lib/inputs/binaries/static/index.ts
```diff
@@ -1 +1,24 @@
...
+  for (const fileName of Object.keys(extractedLayers)) {
+    for (const actionName of Object.keys(extractedLayers[fileName])) {
+      if (actionName !== getNodeBinariesFileContentAction.actionName) {
+        continue;
+      }
+      if (!(typeof extractedLayers[fileName][actionName] === "string")) {
+        throw new Error("expected string");
+      }
+      hashes.push(extractedLayers[fileName][actionName] as string);
```
shaimendel: what happens if the same file in the same path occurs in multiple layers?

Shesekino: the extraction logic always provides the file in the path from the most recent (relevant) layer,
so theoretically we don't expect duplications here.
that said, I don't mind putting it explicitly using a ```set```.
also enhanced the test to check for it.

lib/inputs/binaries/static/index.ts
```diff
@@ -0,0 +1,24 @@
...
+  const hashes: Set<string> = new Set<string>();
+  for (const fileName of Object.keys(extractedLayers)) {
+    for (const actionName of Object.keys(extractedLayers[fileName])) {
+      if (actionName !== getNodeBinariesFileContentAction.actionName) {
+        continue;
+      }
+      if (!(typeof extractedLayers[fileName][actionName] === "string")) {
+        throw new Error("expected string");
+      }
+      hashes.add(extractedLayers[fileName][actionName] as string);
```
shaimendel: will this solve the problem of files in the same locations? looks like it will just dedup in cases the hash repeats twice (and not if the path returns twice)

lib/stream-utils.ts
shaimendel: let's do it then

Shesekino: we will, when it's the time and when the PR will be around it. now is neither the time nor the PR. I'm looking to ship incremental value here, not solve everything :)

lib/inputs/binaries/static/index.ts
Shesekino: not sure what this means, please read my response on the previous comment and let me know if something doesn't make sense https://github.com/snyk/snyk-docker-plugin/pull/139#discussion_r383237088

PR #153, opened 2020-03-08
Fix/types2

Shesekino:
- [x] Ready for review

#### What does this PR do?

improve the typing around the internal parseAnalysisResults() function.
it handles inputs from both static and dynamic analysis, only in the former flow there's some implicit behaviour resulting in ```undefined``` values around key-binary data.
this behaviour is actually okay, so I'm just making it explicit.

https://snyksec.atlassian.net/browse/RUN-761

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,7 +1,13 @@
 import * as Debug from "debug";
 import * as path from "path";
 import * as analyzer from "./analyzer";
-import { AnalysisType } from "./analyzer/types";
+import {
+  AnalysisType,
+  AnalyzedPackage,
+  Binary,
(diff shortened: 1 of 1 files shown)


lib/index.ts
```diff
@@ -274,7 +280,10 @@ async function getManifestFiles(
     .filter((i) => i.contents !== "");
 }
 
-function parseAnalysisResults(targetImage, analysis) {
+function parseAnalysisResults(
+  targetImage,
```
talik077: what is the type of `targetImage`? is it string?

Shesekino: I don't know, it's untyped

PR #155, opened 2020-03-13
fix: Exlcude system directories when calling findGlobs on the root path

mladkau:
- [x ] Ready for review
- [ x] Follows CONTRIBUTING rules
- [x ] Reviewed by Snyk internal team

#### What does this PR do?

This PR adds a special case to findGlobs to make it more efficient. The function now excludes system directories when it is called on the root directory.

https://snyksec.atlassian.net/browse/DC-382

### DIFF
--- a/lib/docker.ts
+++ b/lib/docker.ts
@@ -14,6 +14,8 @@ interface DockerOptions {
   tlsKey?: string;
 }
 
+const SystemDirectories = ["dev", "proc", "sys"];
+
 class Docker {
   public static run(args: string[], options?: DockerOptions) {
     return subProcess.execute("docker", [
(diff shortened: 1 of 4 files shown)


lib/docker.ts
```diff
@@ -14,6 +14,8 @@ interface DockerOptions {
   tlsKey?: string;
 }
 
+const SystemDirectories = ["dev", "proc", "sys"];
```
mladkau: When running findGlobs on the root of a docker image we want to exclude these system directories as they will definitely not contain any manifest files for container applications.

lib/docker.ts
```diff
@@ -128,15 +130,43 @@ class Docker {
...
 
-    if (eventLoopSpinner.isStarving()) {
-      await eventLoopSpinner.spin();
-    }
+    if (recursive && path === "/") {
+      // When scanning from the root of a docker image we need to
+      // exclude system files e.g. /proc, /sys, etc. to make the
+      // operation less expensive.
+
+      const outputRoot = await this.lsSafe("/", false);
```
mladkau: We do a non-recursive ls on the root and then for each subdirectory we do a recursive one except on the excluded system directories. We then merge the results together.

lib/ls-utils.ts
```diff
@@ -51,27 +53,35 @@ function parseLsOutput(output: string): DiscoveredDirectory {
...
+        currDir = getSubDir(res, currPath.substring(1).split("/"));
+      } else {
+        currDir = res;
+      }
+    } else if (i !== "" && !i.endsWith("/") && i !== "." && i !== "..") {
+      currDir.files.push({
+        name: i,
+        path: currPath,
+      });
+    } else if (
```
mladkau: Only this condition was added. It parses in the non recursive case also the subdirectories.

test/lib/docker.test.ts
```diff
@@ -208,7 +208,7 @@ test("findGlobs", async (t) => {
         "\n./\n../\ndir11/\nfile3.json\nfile4.txt\n\n/app/dir1/dir11:\n./\n../\n" +
         "file5.txt\n\n/app/dir2:\n./\n../\n\n/app/dir3:\n./\n../\nfile6.json\n",
     });
-    const files = await docker.findGlobs(["**/file?.json"]);
+    const files = await docker.findGlobs(["**/file?.json"], [], "/foo");
```
mladkau: To the existing tests I've added a pseudo subdirectory so we don't trigger the new special case. In these tests we want to make sure that the inclusion/exclusion mechanism works as expected.

test/lib/ls-utils.test.ts
```diff
@@ -215,14 +228,3 @@ test("parse ls output", async (t) => {
     });
   }
 });
-
-const getLSOutputFixture = (file: string) =>
```
mladkau: This was some testing code which I mistakenly checked in some time ago ... Removing this should speed up the tests :smile:

Shesekino: would be great to have this kind of things in a separate commit, at least for next time :) thanks for the heads up!

lib/ls-utils.ts
```diff
@@ -41,6 +41,8 @@ async function iterateFiles(
  * directory sequence.
  */
 function parseLsOutput(output: string): DiscoveredDirectory {
+  const outputLines = output.trim().split("\n");
+  const isRecursiveOutput = outputLines[0].endsWith(":");
```
Shesekino: won't this mistaken a file name that ends with ```:``` as a false indication of recursive scan?

mladkau: It should be fine for any files which end in `:` as the first line should always be `./` so:
```
 > ls -1ap
./
../
.:
.dockerenv
:
app/
bin/
dev/
etc/
foo:
home/
```
and (here the recursive case):
```
> ls -1apR
.:
./
../
.:
.dockerenv
:
app/
bin/
dev/
etc/
foo:
home/
```

PR #156, opened 2020-03-18
chore: start using release-notes-preview to add visibility to release-notes

Shesekino:
- [x] Ready for review

#### What does this PR do?

will post an estimation of the release notes to each PR opened against the master branch, if a release should occur as a result of merging it.
https://github.com/snyk/release-notes-preview

### DIFF
--- a/.github/workflows/release-notes.yaml
+++ b/.github/workflows/release-notes.yaml
@@ -1,0 +1,21 @@
+name: Release-Notes-Preview
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  preview:
+    runs-on: ubuntu-latest
+    steps:
+    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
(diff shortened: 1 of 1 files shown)


.github/workflows/release-notes.yaml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+    - uses: actions/checkout@v2
+    - run: |
+        git fetch --prune --unshallow --tags
+    - uses: snyk/release-notes-preview@v1.3.4
+      with:
+        releaseBranch: master
+        githubPosterId: 18642669
```
shaimendel: on idea what this means :)

Shesekino: responded on Slack
> the ID is the (public) ID of the snyk-deployer user, it's used to decide which comments to delete (so we don't spam more and more comments).

PR #159, opened 2020-03-20
Feat/container experimental

Shesekino:
- [x] Ready for review

#### What does this PR do?

- statically scanning Distroless APT-installed files
- assumption: image magically present in the local Docker daemon
- somewhat inefficient scanning : first load all apt package files into the memory, then process all of them.

the test file.

https://snyksec.atlassian.net/browse/RUN-797

image


lib/experimental.ts
```diff
@@ -1,15 +1,27 @@
...
   // assumption #2: the `docker` binary is available locally
-  throw new Error("not implemented");
+  // TODO: docker-save the image from the daemon to an archive
+  // TODO: generate options for static analysis based on that
+
+  const scanningOptions = {
+    staticAnalysisOptions: {
+      imagePath: "./egg.tar", // TODO
+      imageType: ImageType.DockerArchive,
+      tmpDirPath: "", // TODO only for RPM though?
```
ivanstanev: Correct! This will soon be removed, because we can feed the RPM file directly into our `rpm-parser` instead of extracting it to file so that some tool can read it from the filesystem 🙏

Shesekino: :fanmob:

test/system/static.test.ts
```diff
@@ -222,3 +223,54 @@ test("static analysis works for scratch images", async (t) => {
     "operating system for scratch image is unknown",
   );
 });
+
+test("static analysis for distroless images", async (t) => {
+  const imageNameAndTag = "gcr.io/distroless/base:latest";
```
ivanstanev: Using latest may cause flakes when the latest image changes one day

lib/analyzer/types.ts
```diff
@@ -29,6 +29,7 @@ export enum AnalysisType {
   Rpm = "Rpm",
   Binaries = "binaries",
   Linux = "linux", // default/unknown/tech-debt
+  Distroless = "distroless",
```
Shesekino: need to see if/how this might affect our backend

lib/inputs/distroless/static.ts
```diff
@@ -0,0 +1,21 @@
+import { ExtractAction, ExtractedLayers } from "../../extractor/types";
+import { streamToString } from "../../stream-utils";
+
+export const getDpkgPackageFileContentAction: ExtractAction = {
+  actionName: "dpkg",
+  fileNamePattern: "/var/lib/dpkg/status.d/*",
+  callback: streamToString, // TODO replace with a parser for apt data extractor
```
Shesekino: this will make the difference between loading into our memory all the files in ```/var/lib/dpkg/status.d/*``` (in addition to all the files we normally load) and only afterwards extracting the little that we need, or loading one file at a time and storing what we need as we go along.
should be significant proportionally to the number of installed packages.

lib/analyzer/static-analyzer.ts
```diff
@@ -44,6 +51,10 @@ export async function analyze(
     getNodeBinariesFileContentAction,
   ];
 
+  if (options.distroless) {
```
Shesekino: something of a hack, especially as it appears twice in this file.
then again, the whole way we determine the package manager and dependencies is a bit hackish so I'm not feeling too bad with it right now.

lib/analyzer/types.ts
Shesekino: removed this bit; we'll use Apt instead.

lib/analyzer/package-managers/apt.ts
```diff
@@ -22,6 +22,24 @@ export function analyze(
   });
 }
 
+export function analyzeDistroless(
+  targetImage: string,
+  aptFiles: string[],
```
shaimendel: should the name actually be `dpkgFiles`? in general in this pr wherever it says apt we might should say dpkg?

lib/analyzer/package-managers/apt.ts
```diff
@@ -22,6 +22,24 @@ export function analyze(
   });
 }
 
+export function analyzeDistroless(
```
shaimendel: I assume that's a material for another PR, but I find it weird that this function is not `async` and we return a `Promise.resolve`.. we not just having it `async` and return the value without the `promise.resolve`

lib/experimental.ts
```diff
@@ -1,15 +1,54 @@
...
   // assume Distroless scanning
-  return distroless(options);
+  return distroless(targetImage);
 }
 
 // experimental flow expected to be merged with the static analysis when ready
-export async function distroless(options: any): Promise<PluginResponse> {
-  // assumption #1: the image is present in the local Docker daemon
+export async function distroless(targetImage: string): Promise<PluginResponse> {
+  const archiveDir = path.join(os.tmpdir(), "snyk-image-archives");
```
shaimendel: this should be validated with the customer

lib/experimental.ts
```diff
@@ -1,15 +1,54 @@
...
+        imageType: ImageType.DockerArchive,
+        // TODO only for RPM, may be removed once we get rid of bdb dep
+        tmpDirPath: "",
+        distroless: true,
+      },
+    };
+
+    return staticModule.analyzeStatically(targetImage, scanningOptions);
+  } finally {
+    fs.unlinkSync(archiveFullPath);
```
shaimendel: this means that we might "leak" folders, in case we create them in `createTempDirIfMissing`

lib/experimental.ts
```diff
@@ -1,15 +1,54 @@
...
+  }
+}
+
+function createTempDirIfMissing(archiveDir: string): void {
+  try {
+    fs.mkdirSync(archiveDir);
```
shaimendel: why not checking if the folder exists instead of this?

lib/analyzer/package-managers/apt.ts
Shesekino: if we want to make a more broad change, then yes.
otherwise I think we should be consistent with the current codebase.
happy to consider a minor refactor/reword effort after we get rid of the dynamic scanning code

lib/analyzer/package-managers/apt.ts
Shesekino: we're not ```await```ing on anything, so we don't need it to be ```async```.
this function returns a Promise as all other analyzer functions return promises, because the base code ```await``` on a list of functions, so we want all of them to be promises.
I'm more than happy to have a bigger refactor in these areas after we remove the dynamic scanning bits, I'm very unhappy with how the code looks right now.

lib/experimental.ts
Shesekino: what?

lib/experimental.ts
Shesekino: a will leak one directory under the OS's temp directory.
I think this is fine :)

lib/experimental.ts
Shesekino: because this code is simpler.

lib/inputs/distroless/static.ts
ivanstanev: Interesting! So the callback would be a function that processes the stream in place and returns only the data that we need from the file? (Instead of reading the content as a string, and only later going over the string to figure out what we need)

Shesekino: thinking of a function that first collects the entire file as a string, then processes it, and stores only the interesting bits.
that is, instead of resulting with a list of files with their entire content, just do the analysis on each file after reading it.
haven't thought it 100% through but the idea is just not storing all the files in memory all the time.

lib/experimental.ts
shaimendel: I'm really not in favour of throwing exception on a "normal" flow, exceptions should be used to errors :) 
what is not simple with checking with `fs.existsSync` before?

Shesekino: you would still need to create it and check for errors, and you would probably want to also check for permissions.
throwing and catching exceptions is an okay pattern when it's controlled well. we're not writing real-time c++ code where a few milliseconds would be critical

Shesekino: this specific bit was part of a snippet I found online to make sure a dir is created on older Node versions (on newer versions than the Plugin uses, you can pass a ```{ recursive: true }``` flag and just similarly to ```mkdir -p``` it would just work

lib/analyzer/static-analyzer.ts
```diff
@@ -61,6 +72,11 @@ export async function analyze(
     getRpmDbFileContent(archiveLayers, options.tmpDirPath),
   ]);
 
+  let distrolessAptFiles: string[] = [];
+  if (options.distroless) {
+    distrolessAptFiles = getAptFiles(archiveLayers);
```
shaimendel: does this has to be wrapped in an if statement?

lib/experimental.ts
shaimendel: that this is the place they want us to save the temporary archives.. for example - maybe the file system is readonly in the tmp folder so this will fail (in their CI/CD process)

Shesekino: I'd be happy to make it configurable - but right now I'm using the standard place to store tmp files as dictated by the OS. I prefer keeping things simple and standard, once we have it working we can extend it.

lib/analyzer/static-analyzer.ts
Shesekino: if we want this change to be rolled in a very controlled environment, then yes in my opinion

lib/experimental.ts
shaimendel: we have only one customer that is supposed to use it, I don't say we shouldn't merge it now, but I say that we need to let them know of that so we can approve that that's good for them

PR #163, opened 2020-03-22
Feat/distroless os info

Shesekino:
- [x] Ready for review

#### What does this PR do?

when analysing ```/etc/os-release``` for OS information, look for the ```PRETTY_NAME```.
if it's there, return it.
otherwise, return an empty string.

regardless, the OS Release data structure will now start containing a ```prettyName``` value.

https://github.com/GoogleContainerTools/distroless/issues/487

https://snyksec.atlassian.net/browse/RUN-805

### DIFF
--- a/lib/analyzer/os-release/release-analyzer.ts
+++ b/lib/analyzer/os-release/release-analyzer.ts
@@ -16,7 +16,13 @@ export async function tryOSRelease(text: string): Promise<OSRelease | null> {
     version = version.split(".")[0];
   }
 
-  return { name, version };
+  let prettyName: string = "";
+  const prettyNameRes = text.match(/^PRETTY_NAME=(.+)$/m);
+  if (prettyNameRes) {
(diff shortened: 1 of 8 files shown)


lib/analyzer/os-release/release-analyzer.ts
```diff
@@ -16,7 +16,16 @@ export async function tryOSRelease(text: string): Promise<OSRelease | null> {
     version = version.split(".")[0];
   }
 
-  return { name, version };
+  let distroless = false;
+  const prettyNameRes = text.match(/^PRETTY_NAME=(.+)$/m);
```
shaimendel: what do you think about always returning the entire `PRETTY_NAME`? just in case we'll want that to other use cases as well

Shesekino: I think I wish I had thought about it before changing 3 test files 😅

PR #164, opened 2020-03-23
Fix/static scan image

Shesekino:
- [x] Ready for review

#### What does this PR do?

fix ```image ID``` sent by the static scanning to stop showing the image name, and instead pick up the image ID (actually image digest) from the archive manifest file.

commit by commit

https://snyk.slack.com/archives/CVDMVLMU3/p1584953582004100

### DIFF
--- a/test/system/static.test.ts
+++ b/test/system/static.test.ts
@@ -401,3 +401,44 @@ test("experimental static analysis for debian images", async (t) => {
     "static experimental flag does not save the image",
   );
 });
+
+test("static and dynamic scanning results are aligned", async (t) => {
+  const imageNameAndTag = "debian:10";
+  const dockerfile = undefined;
+
+  await subProcess.execute("docker", ["image", "pull", imageNameAndTag]);
+  const pluginResultDynamic = await plugin.inspect(imageNameAndTag);
+
+  // static scan doesn't handle creating the image archive yet
+  const archivePath = path.join(os.tmpdir(), "debian-10.tar");
+  await subProcess.execute("docker", [
+    "save",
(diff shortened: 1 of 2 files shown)


test/system/static.test.ts
```diff
@@ -401,3 +401,44 @@ test("experimental static analysis for debian images", async (t) => {
...
+
+  t.equals(
+    JSON.stringify(pluginResultDynamic.package.dependencies),
+    JSON.stringify(pluginResultStatic.package.dependencies),
+    "identical dependencies for regular Debian images between dynamic and static scans",
+  );
+
+  t.equals(
+    pluginResultDynamic.plugin.dockerImageId,
+    `sha256:${pluginResultStatic.plugin.dockerImageId}`, // TODO: how to handle that?
```
ivanstanev: Do we not want the `imageIdFromArchiveManifest()` to return `sha256:<hash>`, similarly to Docker inspect?

Shesekino: that's a good question.
the thing is, ```docker inspect``` just returns it with the prefix, while the static scanning would add it itself?
and what if it's not ```sha256``` but some other hashing algorithm? I'm just not sure 🙈

Shesekino: it looks like nothing really processes this value, I think, except for our UI
https://github.com/snyk/registry/blob/develop/frontend/src/components/ProjectManifestMetaTable.vue#L216-L217
which splits by ```:```, takes the one-before-last part and trims it to 12 characters.
so the result in this case should be identical.
I'm now somewhat comfortable leaving it as is...
wdyt?

PR #165, opened 2020-03-24
Feat/scan openjdk binaries

agatakrajewska:
- [x] Ready for review

#### What does this PR do?

We're adding support for openjdk binaries scanning. 

Plus a bonus typo fix ;-)


lib/inputs/binaries/static/index.ts
```diff
@@ -1,23 +1,36 @@
...
   for (const fileName of Object.keys(extractedLayers)) {
     for (const actionName of Object.keys(extractedLayers[fileName])) {
-      if (actionName !== getNodeBinariesFileContentAction.actionName) {
-        continue;
-      }
-      if (!(typeof extractedLayers[fileName][actionName] === "string")) {
-        throw new Error("expected string");
-      }
-      hashes.add(extractedLayers[fileName][actionName] as string);
+      binariesExtractActions.forEach((action) => {
```
Shesekino: ```for (x of Y)``` is usually more recommended than ```forEach```, it's more readable and easier to understand with async actions

lib/inputs/binaries/static/index.ts
```diff
@@ -1,23 +1,36 @@
 import { ExtractAction, ExtractedLayers } from "../../../extractor/types";
 import { streamToHash } from "../../../stream-utils";
 
+export const getOpenJDKBinariesFileContentAction: ExtractAction = {
+  actionName: "java",
+  fileNamePattern: "**/java",
```
Shesekino: there's several different ```java``` binaries often, for example in ```.../bin/java``` but also ```.../bin/jre/java```.
do we need/want to collect both? worth checking

test/system/static.test.ts
```diff
@@ -185,6 +185,33 @@ test("static analysis provides hashes for key binaries", async (t) => {
...
+      imagePath: getFixture("docker-save/openjdk.tar"),
+      imageType: ImageType.DockerArchive,
+    },
+  };
+
+  const pluginResult = (await plugin.inspect(
+    thisIsJustAnImageIdentifierInStaticAnalysis,
+    dockerfile,
+    pluginOptions,
+  )) as PluginResponseStatic;
```
Shesekino: not sure we need the ```as PluginResponseStatic``` here

test/system/static.test.ts
```diff
@@ -185,6 +185,33 @@ test("static analysis provides hashes for key binaries", async (t) => {
...
+
+  t.equals(pluginResult.hashes.length, 1, "found one openjdk key binary");
+  const expectedHashes = [
+    "004182a1acb5aad313f4554cbafe474a9bdc143260576ac3fa4ab388c3f40476",
```
Shesekino: does this example suffice or should we have one with more hashes in it? that is, more files in different places?

test/system/static.test.ts
agatakrajewska: looks like we do as otherwise it treats it as 'PluginResponse' and in that case it doesn't recognise the hashes

lib/inputs/binaries/static/index.ts
agatakrajewska: as per chat, we're collecting all the java binaries

PR #168, opened 2020-03-27
feat: scanning and hashing of binary files inside docker

firecode9:
- [X] Ready for review
- [X] Reviewed by Snyk internal team

#### What does this PR do?
This PR adds support for binary files scanning inside docker


lib/docker.ts
```diff
@@ -135,14 +139,17 @@ class Docker {
    * Find files on a docker image according to a given list of glob expressions.
    */
   public async findGlobs(
-    globs: string[],
+    globs: any,
```
shaimendel: why?

lib/docker.ts
```diff
@@ -135,14 +139,17 @@ class Docker {
...
     exclusionGlobs: string[] = [],
     path: string = "/",
     recursive: boolean = true,
     excludeRootDirectories: string[] = SystemDirectories,
-  ) {
+  ): Promise<any> {
```
shaimendel: can you please add a type here?

lib/docker.ts
```diff
@@ -186,9 +193,19 @@ class Docker {
         }
       }
       if (!exclude) {
-        for (const g of globs) {
-          if (minimatch(filepath, g)) {
-            res.push(filepath);
+        if (globs.manifestGlobs) {
+          for (const g of globs.manifestGlobs) {
```
shaimendel: is there a way that a file can match more than one glob (either manifest or binary)? I think we should stop the for loops if we find one matching glob, since it is super super expensive in CPU

lib/index.ts
```diff
@@ -146,3 +186,22 @@ async function getManifestFiles(
...
+
+async function calcHashOfBinFiles(
+  docker: Docker,
+  files: string[],
+): Promise<BinHashFile[]> {
+  const contents = await Promise.all(files.map((f) => docker.catSafeB64(f)));
+
+  return files
+    .map((g, i) => {
+      const sha1 = crypto.createHash('sha1');
```
shaimendel: can we use any other modern hash? `sha256` as a minimum (depends what maven central supports)

shaimendel (changes requested):
Since we don't want to make the gap between dynamic and static scanning bigger, please implement also the static counter part of this logic :)

lib/docker.ts
```diff
@@ -116,6 +116,10 @@ class Docker {
     return this.runSafe("cat", [filename]);
   }
 
+  public async catSafeB64(filename: string) {
+    return this.runSafe("cat", [filename, " | base64"]);
```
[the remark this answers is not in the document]
firecode9: This is a temp solution, I'm not going to use it, but I agree :-)

lib/docker.ts
firecode9: globs is an object now of:
manifestGlobs && binariesGlobs

I'll make it into a type

lib/docker.ts
firecode9: yep

lib/index.ts
firecode9: I tend to agree, however I still dont know what mvn central/plugin supports

firecode9: I added an option to define the type of hash to use (and defaults to sha1 in case it is not given)

lib/docker.ts
```diff
@@ -91,6 +101,25 @@ class Docker {
...
 
+  public runAsStream(
+    cmd: string,
+    args: string[] = [],
+    cb: subProcess.ExecuteAsStreamCallback,
+  ) {
+    return subProcess.executeAsStream("docker", cb, [
+      ...this.optionsList,
+      "run",
+      "--rm",
```
shaimendel: I assume we don't want that, we will delete the docker image after we finish :)
even worse, we will re-pull the image on every call to `runAsStream`

lib/docker.ts
```diff
@@ -186,14 +226,27 @@ class Docker {
...
       if (!exclude) {
-        for (const g of globs) {
-          if (minimatch(filepath, g)) {
-            res.push(filepath);
+        let foundInManifestGlobs;
+
+        if (globs.manifestGlobs) {
+          for (const g of globs.manifestGlobs) {
+            if (minimatch(filepath, g)) {
+              foundInManifestGlobs = true;
```
shaimendel: don't you want to `break` on that case? also in the loop below

lib/index.ts
```diff
@@ -146,3 +182,39 @@ async function getManifestFiles(
...
+  const resultArr: BinHashFile[] = [];
+
+  for (const file of files) {
+    let hash: crypto.Hash;
+    if (options && options.hash) {
+      hash = crypto.createHash(options.hash);
+    } else {
+      // defaults to sha1
+      hash = crypto.createHash("sha1");
+    }
```
ivanstanev: Is this used by some external service? SHA1 is deprecated and shouldn't really be used

lib/docker.ts
firecode9: yes, fixed

lib/index.ts
firecode9: afaik, sha1 is used by maven central, so I must use it. if that's not the case I will change it to sha-256 (or sha-512)

lib/types.ts
```diff
@@ -41,6 +52,7 @@ export interface PluginResponse {
   plugin: PluginMetadata;
   package: any;
   manifestFiles: ManifestFile[];
+  binHashFiles: BinHashFile[];
```
Shesekino: note this affects the static scanning as well

lib/docker.ts
```diff
@@ -14,6 +14,16 @@ interface DockerOptions {
   tlsKey?: string;
 }
 
+interface Globs {
+  manifestGlobs?: string[];
+  binariesGlobs?: string[];
```
Shesekino: since we're using singular form for ```manifest``` maybe we should also use it for ```binary```? ```binaryGlobs```? does it make sense? or more likely is my English just bad? 😬

lib/types.ts
Shesekino: also I think this should be some how merged with the ```hashes``` that the static scanning response has?

Shesekino (changes requested):
very nice work, thanks 👏
I would absolutely love to see a few more things, besides the comments I've written:
1. it would be very helpful for the commit message to be more descriptive. ```feat: jar scanning docker``` is great, but I think we can write something slightly more detailed for our users and readers to understand.
a few more details in the commit message's body would also be great.
2. I think a system test would do wonders - show an end-to-end usage of the new functionality and how it affects the plugin's response.
3. waiting for implementation for the static scanning bit - I think it should be simpler than the Docker part 🤞

lib/docker.ts
Shesekino: don't mind me resolving this conversation please, ```--rm``` tells Docker to delete the container, not the image, after running, so it's actually cleaner and we like it ❤️

lib/index.ts
Shesekino: the reasoning here makes sense, but we might want to run this by the internal-security channel at least as a heads up, apparently using sha1 for any reason causes questions

firecode9: image

As you can see here, mvn central api is looking for sha1, so I dont think we have an option

firecode9:
> very nice work, thanks 👏
> I would absolutely love to see a few more things, besides the comments I've written:
> 
> 1. it would be very helpful for the commit message to be more descriptive. `feat: jar scanning docker` is great, but I think we can write something slightly more detailed for our users and readers to understand. 
>    a few more details in the commit message's body would also be great.
> 2. I think a system test would do wonders - show an end-to-end usage of the new functionality and how it affects the plugin's response.
> 3. waiting for implementation for the static scanning bit - I think it should be simpler than the Docker part 🤞

1. Yes, I agree
2. Yes, tests are still missing for the main that I wanted to do it in steps, step 1 is making sure what we have now doesnt break with my changes
3. I think its better to have the static scanning part in a different PR, so this PR wont be too big

lib/docker.ts
```diff
@@ -14,6 +14,16 @@ interface DockerOptions {
   tlsKey?: string;
 }
 
+interface Globs {
+  manifestGlobs?: string[];
```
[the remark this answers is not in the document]
firecode9: ok, fixing

lib/docker.ts
```diff
@@ -116,6 +145,13 @@ class Docker {
     return this.runSafe("cat", [filename]);
   }
 
+  public async catBinarySafe(
+    filename: string,
+    cb: subProcess.ExecuteAsStreamCallback,
+  ) {
```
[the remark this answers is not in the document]
firecode9: fixed

lib/docker.ts
```diff
@@ -14,6 +14,16 @@ interface DockerOptions {
...
 }
 
+interface Globs {
+  manifestGlobs?: string[];
+  binariesGlobs?: string[];
+}
+
+interface FindGlobsResult {
+  manifestFiles: string[];
+  binFiles: string[];
```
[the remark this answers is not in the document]
firecode9: fixed

Shesekino:
> 2\. Yes, tests are still missing for the main reason that I wanted to do it in steps, step 1 is making sure what we have now doesnt break with my changes

fair enough, but before this can be merged I really think we should have some kind of system/end-to-end test. in my opinion, small steps should be about small scope, not compromising about tests. makes sense?

lib/docker.ts
```diff
@@ -186,14 +226,29 @@ class Docker {
...
+              result.binFiles.push(filepath);
+              break;
+            }
           }
         }
       }
     });
 
-    return res;
+    return result;
```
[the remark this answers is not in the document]
firecode9: ok, fixed

lib/index.ts
```diff
@@ -146,3 +182,39 @@ async function getManifestFiles(
     })
     .filter((i) => i.contents !== "");
 }
+
+async function calcHashOfBinFiles(
+  docker: Docker,
+  files: string[],
+  options?: any,
```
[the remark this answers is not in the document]
firecode9: the options are given by the caller to this plugin, so I think it can be many things

lib/sub-process.ts
```diff
@@ -37,3 +51,32 @@ function execute(
     });
   }) as Promise<CmdOutput>;
 }
+
+function executeAsStream(
+  command: string,
+  cb: ExecuteAsStreamCallback,
+  args?: string[],
+  options?,
```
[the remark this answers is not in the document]
firecode9: > the options are given by the caller to this plugin, so I think it can be many things

same thing

lib/sub-process.ts
```diff
@@ -37,3 +51,32 @@ function execute(
...
+): Promise<any> {
```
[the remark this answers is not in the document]
firecode9: here we return just an empty promise, so I think any is fine here, no ?

lib/types.ts
firecode9: not sure what is `hashes` in static, I will check

lib/docker.ts
firecode9: yes. fixed

firecode9:
> > 2. Yes, tests are still missing for the main reason that I wanted to do it in steps, step 1 is making sure what we have now doesnt break with my changes
> 
> fair enough, but before this can be merged I really think we should have some kind of system/end-to-end test. in my opinion, small steps should be about small scope, not compromising about tests. makes sense?

yes i agree, tests should and will be written *before* this merge is ready

lib/index.ts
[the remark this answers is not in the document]
Shesekino: it definitely can - but I still expect ```options``` to be an object (maybe an empty object, but an object all the same). I don't expect it to ever be a string or an Error etc.
I also prefer it to be not undefined-able - that is, if no options are provided, we would just have an empty object, that's fine, no need for it to be undefined

lib/sub-process.ts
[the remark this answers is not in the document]
Shesekino: no, ```any``` means this function may return anything.
if this function returns a Promise that contains no data, then ```Promise<void>``` is what you're looking for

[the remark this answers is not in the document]
firecode9: oh right. fixed

lib/types.ts
firecode9: agreed to merge both dynamic/static data structures

lib/index.ts
```diff
@@ -2,15 +2,22 @@ import * as path from "path";
...
 import * as staticUtil from "./static";
-import { ManifestFile, PluginResponse } from "./types";
+import {
+  BinHashFile,
+  FindFilesResult,
+  ManifestFile,
+  PluginResponse,
+} from "./types";
+
+import * as crypto from "crypto";
```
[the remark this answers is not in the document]
firecode9: fixed

lib/index.ts
```diff
@@ -47,17 +54,21 @@ async function analyzeDynamically(
...
 ): Promise<PluginResponse> {
-  const [runtime, dependencies, manifestFiles] = await Promise.all([
+  const [runtime, dependencies, findFilesResult] = await Promise.all([
     getRuntime(analysisOptions),
     getDependencies(targetImage, dockerfileAnalysis, analysisOptions),
-    getManifestFiles(targetImage, analysisOptions),
+    getFiles(targetImage, analysisOptions),
   ]);
 
+  const manifestFiles = findFilesResult ? findFilesResult.manifestFiles : [];
```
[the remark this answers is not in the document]
firecode9: yes

lib/index.ts
```diff
@@ -109,24 +120,49 @@ async function getDependencies(
...
+    globs.binariesGlobs = options.binariesGlobs as string[];
+  }
 
+  const { manifestFiles, binFiles } = await docker.findGlobs(
+    globs,
+    excludeGlobs,
+  );
+
+  return {
+    manifestFiles: await handleManifestFiles(docker, manifestFiles),
```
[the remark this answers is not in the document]
firecode9: logic tells me it's sequential. I'll fix it.

lib/index.ts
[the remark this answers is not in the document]
firecode9: I dont understand something...
These options are coming from here:
image

and it's defined as `options?: any`, so I'm not sure how I can change it

lib/sub-process.ts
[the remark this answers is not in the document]
firecode9: options is not needed here so I'm removing it.

lib/index.ts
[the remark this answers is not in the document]
Shesekino: it's defined as ```any``` because it was untyped to begin with.
a few weeks ago when our team resumed work on the docker-plugin I started adding types wherever I could, without going down rabbit holes too much. that is, where things were too difficult to type, I added ```any``` to mark them as "TODO" and moved on to things easier to type, incrementally improving things as I went.

the ```options``` object is one of the things I haven't gotten to covering yet, but basically we know it's an object that may or may not have things inside it.

so, what I mean is that if we're adding more code here, we shouldn't stick to the technical debt we have around typing in the plugin, but rather try to improve it, even just slightly every time.

makes sense?
if we can find a way to improve it, it would be great. if it's too much work right now then I guess we can disregard it... whatever makes sense to you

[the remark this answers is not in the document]
firecode9: as agreed, this change would be done later on as part of a bigger refactor to this codebase

lib/index.ts
```diff
@@ -146,3 +189,37 @@ async function getManifestFiles(
     })
     .filter((i) => i.contents !== "");
 }
+
+async function calcHashOfBinaryFiles(
```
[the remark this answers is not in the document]
firecode9: I've moved the function into docker class

lib/index.ts
```diff
@@ -146,3 +189,37 @@ async function getManifestFiles(
...
+  docker: Docker,
```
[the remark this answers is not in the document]
firecode9: I've moved the function into docker class

Shesekino (commented):
nice and clean 💅 
for the 2nd commit, let's please rephrase the commit message and not call it ```docker``` since it's a container-image not necessarily based on Docker, and we added specifically jar files.
so 
```feat: scanning and hashing of Java archives inside an image (static)``` is more accurate

I'm also missing some tests related to the new functionality:
1. showing the static scanning works for Java archives
2. showing the static scanning is aligned with the dynamic scanning (at least in terms of hashes found around Java archives)

firecode9:
> nice and clean 💅
> for the 2nd commit, let's please rephrase the commit message and not call it `docker` since it's a container-image not necessarily based on Docker, and we added specifically jar files.
> so
> `feat: scanning and hashing of Java archives inside an image (static)` is more accurate
> 
> I'm also missing some tests related to the new functionality:
> 
> 1. showing the static scanning works for Java archives
> 2. showing the static scanning is aligned with the dynamic scanning (at least in terms of hashes found around Java archives)

the 2nd commit is still on going, sorry but forgot to mention it and to flag that it's not ready for merge yet.

ivanstanev (commented):
Left some comments but I'm not sure I fully understand the feature to give an in-depth review!

One request for future PRs is to split the changes into small, scoped commits. I find it impossible to understand the Dynamic scanning changes, as too many things happen in a single commit.
Also if something goes wrong, it will be incredibly hard to pinpoint _what_ in the tens of changes caused the problem. A list of sample commit changes for the PR could look like this:
- renaming some interface
- introducing a new function (and its tests), with a good commit description what's the intended usage (here I'm thinking of `catSafe`)
- replacing the return type for a function with some new interface
- introducing some other new function (here I'm thinking of `calcHashOfBinaryFiles`)
- making currently private vars/funcs now public/exported
- only now modifying the code to start using all of the above changes, and adjusting the tests if needed

I find that this is much easier to follow and see what your intentions are, it's incrementally improving the code rather than dumping everything in one big change. It also gives enough context how the code should improve gradually. And if some commit introduced a hard-to-fix problem, we could revert just that commit and still benefit from the rest of the improvements, instead of reverting a single massive commit and losing on all of the changes.

Shesekino:
@firecode9 should we close this one?

PR #172, opened 2020-03-31
feat: use @snyk/rpm-parser to parse RPM packages

ivanstanev:
Removes the use of an external tool to process RPM packages.

- [x] Ready for review

#### More information

Added some questions for discussion as review comments below


lib/inputs/rpm/static.ts
```diff
@@ -14,62 +14,21 @@ export const getRpmDbFileContentAction: ExtractAction = {
...
+    const parserResponse = await getPackages(rpmDb);
+    if (parserResponse.error !== undefined) {
+      throw parserResponse.error;
+    }
+    return parserResponse.response;
+  } catch (error) {
+    debug("An error occurred while analysing RPM packages");
+    debug(error);
+    return "";
   }
```
ivanstanev: Notice that regardless of the error type, we prefer to log for now.
Do you think we should allow the error to propagate up? I think I prefer to fail hard instead of silently returning an empty response as if we didn't process anything...

ivanstanev:
Thinking of also adding a container image fixture with a tiny RPM database to have a sanity check for the parser.

lib/inputs/rpm/static.ts
Shesekino: I haven't gone too much into the details here but I propose:
1. in terms of error throwing, make no changes from the existing state
2. in terms of logging and propagating **debug data**, do whatever you think is good for us to improve things

ivanstanev: I think the old flow basically never threw, we haven't had any cases where it did, hence the current change of not throwing should be OK as well!

lib/inputs/rpm/static.ts
```diff
@@ -14,62 +14,21 @@ export const getRpmDbFileContentAction: ExtractAction = {
...
-  } finally {
-    await removeFile(filePath);
-  }
-}
-
-function handleError(error): CmdOutput | never {
-  const stderr = error.stderr;
-  if (typeof stderr === "string" && stderr.indexOf("not found") >= 0) {
-    return { stdout: "", stderr: "" };
+    const parserResponse = await getPackages(rpmDb);
```
Shesekino, line 24: why do we need the try/catch here if we're just printing the error and returning an empty string?
in case ```getPackages``` throws even though theoretically it shouldn't?

Shesekino:
> Thinking of also adding a container image fixture with a tiny RPM database to have a sanity check for the parser.

either that or a test that begins with pulling an image, saving it to an archive, then you're a bit more flexible about the size of the image as it's not checked in the repo

lib/inputs/rpm/static.ts
ivanstanev: Correct, it's in case`getPackages` throws, if the DB file is bad/corrupted or some of our assumptions about the DB have changed.

ivanstanev:
> > Thinking of also adding a container image fixture with a tiny RPM database to have a sanity check for the parser.
> 
> either that or a test that begins with pulling an image, saving it to an archive, then you're a bit more flexible about the size of the image as it's not checked in the repo

I think it will slow down tests significantly depending on your connection, all these RPM-based images are in the hundreds of megabytes so I preferred to check in a ~5MB fixture instead...

PR #177, opened 2020-04-07
feat: return scan results array

Shesekino:
- [x] Ready for review

#### What does this PR do?

the ecosystem hints array contains unstructured data for different ecosystems encountered while scanning an image.
the `data` is expected to be consistent for each `type`+`version` couple.
```
Plugin Result: {
  plugin: PluginMetadata; // probably irrelevant RIGHT NOW
  package: {
    dependencies: { [key: string]: any; }; // OS dep graphs
    docker: { [key: string]: any; }; // not even god knows what’s in here
    name: string; // image name
    version: string; // image tag
    packageFormatVersion: string; // “deb:0.0.1”, used in Vuln
    targetOS: {
      name: string; // “debian”
      prettyName: string; // "Debian GNU/Linux 10 (buster)", “Distroless”
      version: string; “10”
     };
  };
   // ******** ADDING NOW ********
  scanResults: [
    {
      "type": "node",
      "version": "0.0.1",
      "data": TBD,
    {
      "type": "node",
      "version": "0.0.1",
      "data": TBD,
    },
    {
      "type": "java",
      "version": "0.0.2",
      "data": TBD,
    },
    {
      "type": "debian",
      "version": "0.0.1",
      "data": TBD,
    },
  ],
};
```

### DIFF
--- a/lib/response-builder.ts
+++ b/lib/response-builder.ts
@@ -29,6 +29,7 @@ function buildResponse(
     plugin,
     package: pkg,
     manifestFiles,
+    scanResults: [], // TODO
   };
 }
 
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -32,6 +32,13 @@ export interface PluginMetadata {
   imageLayers: string[];
 }
 
+interface ScanResult {
+  type: string;
+  version: string;
+  data: any;
+}
+
+// should be renamed and organised
 export interface Package {
   dependencies: {
     [key: string]: any; // TODO
@@ -51,7 +58,8 @@ export interface Package {
 
 export interface PluginResponse {
   plugin: PluginMetadata;
-  package: Package;
+  package: Package; // under deprecation, Package is one type of scanResult
+  scanResults: ScanResult[]; // to replace package
   manifestFiles: ManifestFile[];
 }
 
--- a/test/fixtures/responses/all-deps.json
+++ b/test/fixtures/responses/all-deps.json
@@ -1115,5 +1115,6 @@
       },
       "binaries": []
     }
-  }
+  },
+  "scanResults": []
 }


lib/types.ts
```diff
@@ -32,6 +32,12 @@ export interface PluginMetadata {
   imageLayers: string[];
 }
 
+interface EcosystemHints {
```
firecode9, line 35: I think an export is missing here

Shesekino: it's not used anywhere in the code, but let's please focus more on the message (the name and the placement in the data structure) and less on the implementation detail :) it's a draft

firecode9:
👍 looks good

michael-go:
note the `package` is supposed to be a single dependency-tree. With time, extra stuff was attached to it ... stuff that kinda made half-sense as it provided needed metadata (like `tragetOS` needed to match to correct vulns snapshot ...),
but here we you adding data about multiple scan artifacts, feels very wrong to go under the sigular `package` alongside `.dependencies`.
I suggest this should instead be exposed via a new interface (not current `inspect()`), which will return multiple scan results, where the os-packages are just one of the results in an array. I imagine a top level object with things that describe the entire image like: OS, layers, etc and then an array with all the different artifacts: os-packages-dep-graph, key-binaries-array-or-flat-dep-graph-TBD, node-app-dep-graph-or-manifests-TBD, python-app-dep-graph-or-manifsts-TBS, maven-jars-thing-TBD

Shesekino:
@michael-go why a new interface?
👍 on the suggested for an array with different artifacts

michael-go:
maybe it doesn't have to be a new interface actually, but it 100% IMO shouldn't be under `.package`.
today `inspect()` returns `{plugin, package}`- so to support multiple artifacts it should at the very least be a new element with a new interface.

lib/types.ts
```diff
@@ -32,6 +32,13 @@ export interface PluginMetadata {
...
```
Shesekino, line 35: can be renamed later on, I just don't want it mixed up with the existing ```package```

lib/types.ts
```diff
@@ -51,7 +58,8 @@ export interface Package {
 
 export interface PluginResponse {
   plugin: PluginMetadata;
-  package: Package;
+  package: Package; // under deprecation, Package is one type of ecosystem hint
+  packages: EcosystemHints[]; // to replace package
   manifestFiles: ManifestFile[];
```
firecode9, line 63: I know it's not related to this change, but we might want to remove this field going forward (or now), as the new EcosystemHints[] will provide these.
@mladkau ?

Shesekino: I think we would remove it when we restructure how ```packages``` look like exactly, but let's keep the discussion and changes focused please 🙏

lib/response-builder.ts
```diff
@@ -29,6 +29,7 @@ function buildResponse(
     plugin,
     package: pkg,
     manifestFiles,
+    scanResults: [], // TODO
```
shaimendel, line 32: what is the todo here?

Shesekino: ```scanResults``` are expected to actually have things in them eventually

PR #178, opened 2020-04-07
Feat/allow complex file matching

ivanstanev:
- [x] Ready for review

#### What does this PR do?

Fixes two issues with the current extraction logic, namely:
- Complex file pattern matching was not possible (e.g. if we want to exclude certain paths), so now this logic is handled by a function in every ExtractAction
- A bug in `getContent()` that treats match patterns as the file names, so files extracted by an ExtractAction that used patterns could never be read (we didn't use those, but would be a bug if we did).

Commit by commit as usual 🙏 

[Jira ticket RUN-859](https://snyksec.atlassian.net/browse/RUN-859)
[Jira ticket RUN-860](https://snyksec.atlassian.net/browse/RUN-860)

### DIFF
--- a/test/lib/extractor/image.test.ts
+++ b/test/lib/extractor/image.test.ts
@@ -1,3 +1,7 @@
+#!/usr/bin/env node_modules/.bin/ts-node
+// Shebang is required, and file *has* to be executable: chmod +x file.test.js
+// See: https://github.com/tapjs/node-tap/issues/313#issuecomment-250067741
+
 import * as path from "path";
 import { test } from "tap";
 import { getDockerArchiveLayersAndManifest } from "../../../lib/extractor";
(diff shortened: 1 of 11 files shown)


test/lib/extractor/image.test.ts
```diff
@@ -1,3 +1,7 @@
+#!/usr/bin/env node_modules/.bin/ts-node
+// Shebang is required, and file *has* to be executable: chmod +x file.test.js
+// See: https://github.com/tapjs/node-tap/issues/313#issuecomment-250067741
+
```
Shesekino, line 4: 🤦‍♂ we would want to address this silliness at some point... maybe move to Jest while at it?
awesome find 👍

ivanstanev: I dream about moving to Jest every night

PR #179, opened 2020-04-08
fix: match file paths using string comparison instead of using minimatch

ivanstanev:
minimatch is a heavy library that we don't need to call often. Instead we could use a simple string comparison to achieve the same thing in static scanning.

- [x] Ready for review

[Jira ticket RUN-861](https://snyksec.atlassian.net/browse/RUN-861)

### DIFF
--- a/lib/inputs/binaries/static/index.ts
+++ b/lib/inputs/binaries/static/index.ts
@@ -1,17 +1,15 @@
-import * as minimatch from "minimatch";
-
 import { ExtractAction, ExtractedLayers } from "../../../extractor/types";
 import { streamToHash } from "../../../stream-utils";
 
 export const getOpenJDKBinariesFileContentAction: ExtractAction = {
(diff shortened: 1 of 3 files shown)


lib/inputs/binaries/static/index.ts
```diff
@@ -1,17 +1,15 @@
-import * as minimatch from "minimatch";
-
 import { ExtractAction, ExtractedLayers } from "../../../extractor/types";
 import { streamToHash } from "../../../stream-utils";
 
 export const getOpenJDKBinariesFileContentAction: ExtractAction = {
   actionName: "java",
-  filePathMatches: (filePath) => minimatch(filePath, "**/java", { dot: true }),
+  filePathMatches: (filePath) => filePath.endsWith("java"),
```
agatakrajewska, line 6: amazing 😄

PR #180, opened 2020-04-09
fix: refactor code to add sha1 future support

firecode9:
- [x] Ready for review

#### What does this PR do?
In order to support sha1 hashing as well in the future, I've refactored the code a bit to be able to select the type of hash we are going to do.

sha1 will be used going forward for jar scanning, as maven central stores only sha1 in the query api

### DIFF
--- a/lib/stream-utils.ts
+++ b/lib/stream-utils.ts
@@ -1,8 +1,9 @@
 import * as crypto from "crypto";
 import { Readable } from "stream";
 
-const HASH_ALGORITHM = "sha256"; // TODO algorithm?
-const HASH_ENCODING = "hex";
+export const HASH_ALGORITHM_SHA256 = "sha256"; // TODO algorithm?
+export const HASH_ALGORITHM_SHA1 = "sha1";
+export const HASH_ENCODING = "hex";
 
 export async function streamToString(
   stream: Readable,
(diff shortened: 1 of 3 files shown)


lib/stream-utils.ts
```diff
@@ -33,10 +34,22 @@ export async function streamToBuffer(stream: Readable): Promise<Buffer> {
...
+}
+
+export async function streamToHashSHA256(stream: Readable): Promise<string> {
+  return doStreamToHash(stream, HASH_ALGORITHM_SHA256, HASH_ENCODING);
+}
+
+export async function doStreamToHash(
+  stream: Readable,
+  hashAlgo: string,
+  hashEnc: string,
```
ivanstanev, line 48: IMO it's better to be explicit rather than shorten var names: `hashAlgorithm` and `hashEncoding`; it's just better for readability

lib/stream-utils.ts
```diff
@@ -33,10 +34,22 @@ export async function streamToBuffer(stream: Readable): Promise<Buffer> {
...
 }
 
-export async function streamToHash(stream: Readable): Promise<string> {
+export async function streamToHashSHA1(stream: Readable): Promise<string> {
+  return doStreamToHash(stream, HASH_ALGORITHM_SHA1, HASH_ENCODING);
+}
+
+export async function streamToHashSHA256(stream: Readable): Promise<string> {
+  return doStreamToHash(stream, HASH_ALGORITHM_SHA256, HASH_ENCODING);
+}
```
ivanstanev, line 43: Perhaps a small unit test for each to validate them?

ivanstanev (changes requested):
Left some minor comments!

Also the commit should be a `fix` because we touch the code. Chores are usually for changes like renaming a variable (this doesn't change the functionality) or updating docs or updating our CI config (again, not a functional change!). Could you also add a short explanation in the commit message why we need the SHA1 support? (I know why, but the next person who reviews the changes won't 😄)

lib/stream-utils.ts
firecode9: Agreed .fixed

lib/stream-utils.ts
firecode9: yes. on the way..

Shesekino:
👋 I'm not sure that's the exact direction we should take this refactor.
let's please wait until this kind of hashing is needed, then let team Runtime handle the refactor and implementation there.

PR #187, opened 2020-04-12
[DRAFT] node deps

Shesekino:
some (incomplete) code and some (incomplete) thoughts on extracting node dependency trees from manifest files


lib/analyzer/applications/node.ts
```diff
@@ -0,0 +1,116 @@
...
+const TYPE = "node";
+const VERSION = "1.0.0";
+
+export interface NodeScanResult extends types.ScanResult {
+  type: string;
+  version: string;
+  data: NodeData;
+}
+
+interface NodeData {
```
shaimendel, line 15: we should consider somehow having a map somewhere in the plugin between the app type, app version and the interface itself, so maybe Registry can use it as well? instead of defining things in 2 places

lib/analyzer/applications/node.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export async function nodeLockFilesToData(
+  filePathToContent: FilePathToContent,
+): Promise<NodeScanResult[]> {
+  const scanResults: NodeScanResult[] = [];
+
+  const filePairs = findManifestLockPairsInSameDirectory(filePathToContent);
+  for (const thing of filePairs) {
+    // TODO: consider spinning the event loop
+    const parserResult = await lockFileParser.buildDepTree(
```
shaimendel, line 39: what happens if for some reason `buildDepTree` fails? does it throw? just wondering if we need to wrap in try-catch to continue to other files

Shesekino: no idea, but good point

lib/analyzer/applications/node.ts
Shesekino: https://github.com/snyk/scan-schemas is expected to host that

PR #190, opened 2020-04-22
feat: allow scanning existing Docker archives on the filesystem

ivanstanev:
As part of the experimental option, allow the plugin to directly unpack and analyse a Docker archive that exists on the filesystem.
This uses the existing static analysis flow.

- [ ] Ready for review

[Jira ticket RUN-897](https://snyksec.atlassian.net/browse/RUN-897)
[Jira ticket RUN-877](https://snyksec.atlassian.net/browse/RUN-877)

### DIFF
--- a/lib/experimental.ts
+++ b/lib/experimental.ts
@@ -4,15 +4,41 @@ import * as path from "path";
 
 import { pullIfNotLocal } from "./analyzer/image-inspector";
 import { Docker } from "./docker";
+import { getDockerArchivePath, getImageTransport } from "./image-transport";
 import * as staticModule from "./static";
-import { ImageType, PluginResponse } from "./types";
+import {
+  ImageTransport,
+  ImageType,
+  PluginResponse,
+  StaticAnalysisOptions,
+} from "./types";
 
 export async function experimentalAnalysis(
   targetImage: string,
   options: any,
 ): Promise<PluginResponse> {
   // assume Distroless scanning
-  return distroless(targetImage, options);
+  const imageTransport = getImageTransport(targetImage);
+  switch (imageTransport) {
+    case ImageTransport.DockerArchive:
+      // TODO: maybe all the steps in this case can be wrapped in a function
+      const imagePath = getDockerArchivePath(targetImage);
+      // TODO: validate the file in "imagePath" exists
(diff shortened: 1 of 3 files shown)


lib/experimental.ts
```diff
@@ -4,15 +4,41 @@ import * as path from "path";
...
 export async function experimentalAnalysis(
   targetImage: string,
   options: any,
 ): Promise<PluginResponse> {
   // assume Distroless scanning
-  return distroless(targetImage, options);
+  const imageTransport = getImageTransport(targetImage);
+  switch (imageTransport) {
+    case ImageTransport.DockerArchive:
+      // TODO: maybe all the steps in this case can be wrapped in a function
```
shaimendel, line 24: yup, we should have a function that gets a tar file path on the disk and returns the scan. that way the distroless flow (which should be called in a different name I guess now, something like containerRegsitryAnalysis) can call it after it calls docker save

lib/types.ts
```diff
@@ -4,6 +4,11 @@ export interface StaticAnalysisOptions {
   distroless: boolean;
 }
 
+export enum ImageTransport {
```
Shesekino, line 7: what does image transport mean?

lib/image-transport.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export function getImageTransport(targetImage: string): ImageTransport {
+  if (targetImage.startsWith("docker-archive:")) {
+    return ImageTransport.DockerArchive;
+  }
+
+  return ImageTransport.ContainerRegistry;
+}
+
+export function getDockerArchivePath(targetImage: string): string {
```
Shesekino, line 11: does it have to be a Docker archive? or any image archive?

ivanstanev: I think so, the format of this tar is specifically named "docker archive" and "targetImage" is going to contain the prefix `docker-archive:`

lib/types.ts
ivanstanev: You can obtain an image from a variety of sources, e.g. container registry, docker socket, filesystem path, etc. I think `skopeo` is calling it "repository type":
https://github.com/containers/skopeo
> Skopeo operates on the following image and repository types:

Maybe it's not a bad idea to use that terminology 😄

PR #191, opened 2020-04-22
feat/allow scanning existing Docker archives on the filesystem

ivanstanev:
- [x] Ready for review

#### What does this PR do?

Allows to scan Docker archives as long as:
- the `experimental` flag is used
- the image is present on the filesystem

The image should be provided as `docker-archive:/path/to/archive.tar`.

[Jira ticket RUN-897](https://snyksec.atlassian.net/browse/RUN-897)
[Jira ticket RUN-877](https://snyksec.atlassian.net/browse/RUN-877)

### DIFF
--- a/lib/dependency-tree/index.ts
+++ b/lib/dependency-tree/index.ts
@@ -25,6 +25,10 @@ export function buildTree(
     imageVersion = targetImage.slice(versionSeparator + 1);
   }
 
+  if (imageName.endsWith(".tar")) {
+    imageVersion = "";
+  }
+
   const shaString = "@sha256";
 
   if (imageName.endsWith(shaString)) {
--- a/lib/experimental.ts
+++ b/lib/experimental.ts
@@ -4,6 +4,7 @@ import * as path from "path";
 
 import { pullIfNotLocal } from "./analyzer/image-inspector";
 import { Docker } from "./docker";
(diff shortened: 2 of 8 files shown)


lib/experimental.ts
```diff
@@ -46,20 +46,27 @@ export async function distroless(
...
-    return await staticModule.analyzeStatically(targetImage, scanningOptions);
+    return await getStaticAnalysisResult(targetImage, archiveFullPath);
   } finally {
     fs.unlinkSync(archiveFullPath);
   }
 }
 
+async function getStaticAnalysisResult(
+  targetImage: string,
+  imagePath: string,
```
Shesekino, line 71: image path === archive path?

lib/dependency-tree/index.ts
```diff
@@ -25,6 +25,10 @@ export function buildTree(
     imageVersion = targetImage.slice(versionSeparator + 1);
   }
 
+  if (imageName.endsWith(".tar")) {
```
shaimendel, line 28: shouldn't we check if it begins with `docker-archive` like in the second place we do it?

ivanstanev: We strip that from the `targetImage` so there is no prefix here

PR #192, opened 2020-04-23
feat: return multiple scan results to support application dependencies

Shesekino:
- [x] Ready for review

#### What does this PR do?

BREAKING CHANGE: this commit changes the return type of the main `inspect` function to adhere to Snyk's CLI's newer format that supports multiple scan results, sometimes called projects.
To use the new format, iterate over the new `scannedProjects` property of the return value. Each item has a `depTree` with dependencies (similar to the old `Package`) and a `packageManager`.

```snyk test```
```snyk monitor```

we're adapting to the CLI's new format that supports multiple scan results.

https://snyksec.atlassian.net/browse/RUN-893

### CODE REVIEW

File: test/system/system.test.ts [hunk 1]
```diff
@@ -595,18 +590,6 @@ test("inspect centos", (t) => {
...
-        manifest,
-        [
-          {
-            name: "redhat-release",
-            path: "/etc",
-            contents: "Q2VudE9TIExpbnV4IHJlbGVhc2UgNy40LjE3MDggKENvcmUpIAo=",
-          },
-        ],
-        "manifest files",
-      );
```
ivanstanev, line 609: Is this removed unintentionally?

test/system/static.test.ts
```diff
@@ -88,8 +85,8 @@ test("static analysis builds the expected response", async (t) => {
   );
 
   t.deepEqual(
-    pluginResultWithSkopeoCopy.package.dependencies,
-    pluginResultWithDockerSave.package.dependencies,
+    pluginResultWithSkopeoCopy.scannedProjects[0].depTree.dependencies,
```
ivanstanev, line 88: In every test that we now use `scannedProjects`, could we also introduce an assertion that the length === 1? It will help us catch if there are more returned results! (At which point we can decide if that is expected)

lib/types.ts
```diff
@@ -32,29 +32,84 @@ export interface PluginMetadata {
...
   };
-  docker: {
-    [key: string]: any; // TODO
+  labels?: {
+    [key: string]: string;
   };
-  name: string;
+}
+
+export interface DepTree extends DepTreeDep {
```
ivanstanev, line 64: Are these two interfaces present in some shared library we could use? They seem oddly familiar!

lib/types.ts
```diff
@@ -32,29 +32,84 @@ export interface PluginMetadata {
...
+  docker: {
+    [key: string]: any; // TODO
+  };
+  files?: any;
 }
 
-export interface PluginResponseStatic extends PluginResponse {
-  hashes: string[];
-}
+// export type SupportedPackageManagers =
```
ivanstanev, line 81: Useful as a reference but not sure it should be part of the code! Maybe as part of an internal doc?

test/system/system.test.ts
Shesekino: yes, I'm not sure if the multiple response object gets processed in the CLI to even preserve this list, and even if it does I believe having them there is incorrect

test/system/static.test.ts
Shesekino: that's a bit like pouring cement over our code, because:
1. right now it's hard coded to return a single item
2. I'm starting to work on returning additional results
3. I have this assertion in a single place, duplicating it now means rewriting it in a few days.

when this area is stabilised a bit I'm happy to write more thorough tests

lib/types.ts
```diff
@@ -32,29 +32,84 @@ export interface PluginMetadata {
...
+}
+
+// NEW STUFF
+
+export interface PluginResponse {
+  plugin: PluginMetadata;
+  scannedProjects: ScannedProjectCustom[];
+}
+
+export interface ScannedProjectCustom {
```
firecode9, line 46: Maybe let's drop the custom ? In the cli the custom is:

image

firecode9: Another option is to move ScannedProjectCustom to cli-interface maybe and use it there ?

lib/types.ts
Shesekino: they are written and rewritten everywhere 😭and the current advice is not to use a shared library.
you can find them in http://github.com/snyk/snyk-cli-interface, https://github.com/snyk/snyk, and some of our backend services I believe in different forms.

firecode9: image

They exists on the cli-interface module, but I think @Shesekino doesnt want to have cli-interface import here

lib/types.ts
Shesekino: it's only somewhat useful if it's in this specific place for the next few weeks. afterwards I definitely want it cleaned.

lib/types.ts
Shesekino: that's what we have, except for the optional ```callGraph``` which I don't want to include yet, and for that here it's flattened.

test/system/static.test.ts
ivanstanev: > I have this assertion in a single place 

If this gives us enough confidence that it would cover the other cases then I'm happy with it! What I want to prevent here is for some unexpected reason we start creating more projects than we should 🙏

Shesekino: that's a good concern and thanks for raising it :) but right now it's not needed.
hopefully the next piece of work (returning node dependencies) will solidify some of the tests

shaimendel (dismissed):

PR #195, opened 2020-04-28
feat: scan Node application dependencies in static scanning

ivanstanev:
- [x] Ready for review

#### What does this PR do?

Introduces Node application dependencies scanning as part of static scanning.

Review commit by commit 🙏 

[Jira ticket RUN-880](https://snyksec.atlassian.net/browse/RUN-880)

### DIFF
--- a/lib/analyzer/applications/node.ts
+++ b/lib/analyzer/applications/node.ts
@@ -1,0 +1,118 @@
+import * as path from "path";
+import * as lockFileParser from "snyk-nodejs-lockfile-parser";
+
+import { ScannedProjectCustom } from "../../types";
+import { FilePathToContent } from "./types";
+
+interface ManifestLockPathPair {
+  manifest: string;
+  lock: string;
+  lockType: lockFileParser.LockfileType;
+}
+
+export async function nodeFilesToScannedProjects(
+  filePathToContent: FilePathToContent,
+): Promise<ScannedProjectCustom[]> {
+  const scanResults: ScannedProjectCustom[] = [];
+
+  const filePairs = findManifestLockPairsInSameDirectory(filePathToContent);
+
+  const shouldIncludeDevDependencies = false;
+  const shouldBeStrictForManifestAndLockfileOutOfSync = false;
+
+  for (const pathPair of filePairs) {
+    // TODO: initially generate as DepGraph
+    const parserResult = await lockFileParser.buildDepTree(
+      filePathToContent[pathPair.manifest],
+      filePathToContent[pathPair.lock],
+      shouldIncludeDevDependencies,
+      pathPair.lockType,
+      shouldBeStrictForManifestAndLockfileOutOfSync,
+      // Don't provide a default manifest file name, prefer the parser to infer it.
+    );
+
+    const strippedLabelsParserResult = stripUndefinedLabels(parserResult);
+
+    scanResults.push({
+      depTree: strippedLabelsParserResult,
+      packageManager: pathPair.lockType,
+      // The targetFile ensures project uniqueness; we choose the manifest file as a target.
+      targetFile: pathPair.manifest,
+    });
+  }
+
+  return scanResults;
+}
+
+function findManifestLockPairsInSameDirectory(
+  filePathToContent: FilePathToContent,
+): ManifestLockPathPair[] {
+  const fileNamesGroupedByDirectory = groupFilesByDirectory(filePathToContent);
+  const manifestLockPathPairs: ManifestLockPathPair[] = [];
+
+  for (const directoryPath of Object.keys(fileNamesGroupedByDirectory)) {
+    const filesInDirectory = fileNamesGroupedByDirectory[directoryPath];
+    if (filesInDirectory.length !== 2) {
+      // either a missing file or too many files, ignore
+      continue;
+    }
+
+    const hasPackageJson = filesInDirectory.includes("package.json");
+    const hasPackageLockJson = filesInDirectory.includes("package-lock.json");
+    const hasYarnLock = filesInDirectory.includes("yarn.lock");
+
+    if (hasPackageJson && hasPackageLockJson) {
+      manifestLockPathPairs.push({
+        manifest: path.join(directoryPath, "package.json"),
+        lock: path.join(directoryPath, "package-lock.json"),
+        lockType: lockFileParser.LockfileType.npm,
+      });
+      continue;
+    }
+
+    if (hasPackageJson && hasYarnLock) {
+      manifestLockPathPairs.push({
+        manifest: path.join(directoryPath, "package.json"),
+        lock: path.join(directoryPath, "yarn.lock"),
+        lockType: lockFileParser.LockfileType.yarn,
+      });
+      continue;
+    }
+  }
+
+  return manifestLockPathPairs;
+}
+
+// assumption: we only care about manifest+lock files if they are in the same directory
+function groupFilesByDirectory(
+  filePathToContent: FilePathToContent,
+): { [directoryName: string]: string[] } {
+  const fileNamesGroupedByDirectory: { [directoryName: string]: string[] } = {};
+  for (const filePath of Object.keys(filePathToContent)) {
+    const directory = path.dirname(filePath);
+    const fileName = path.basename(filePath);
+    if (!fileNamesGroupedByDirectory[directory]) {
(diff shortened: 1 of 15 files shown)


lib/types.ts
```diff
@@ -52,9 +52,9 @@ export interface ScannedProjectCustom {
 }
 
 export interface DepTreeDep {
-  name: string;
-  version: string;
-  dependencies: {
+  name?: string;
```
Shesekino, line 55: the CLI doesn't know what it's getting or what plugin it's calling necessarily, so it doesn't mark these fields as mandatory.
the Plugin does know what it aims to return, so marking them as optional would be a mistake. I think we should definitely remove this commit

package.json
```diff
@@ -26,6 +26,7 @@
     "dockerfile-ast": "0.0.19",
     "event-loop-spinner": "^1.1.0",
     "semver": "^6.1.0",
+    "snyk-nodejs-lockfile-parser": "^1.21.0",
```
Shesekino, line 29: might be worth pinning to a specific version

lib/analyzer/applications/node.ts
```diff
@@ -0,0 +1,117 @@
...
+export async function nodeFilesToScannedProjects(
+  filePathToContent: FilePathToContent,
+): Promise<ScannedProjectCustom[]> {
+  const scanResults: ScannedProjectCustom[] = [];
+
+  const filePairs = findManifestLockPairsInSameDirectory(filePathToContent);
+
+  const shouldIncludeDevDependencies = false;
+  const shouldBeStrictForManifestAndLockfileOutOfSync = false;
+  const defaultManifestFileName = undefined;
```
shaimendel, line 22: are we ok with that? will it maybe have UI implication in some scenario?

lib/analyzer/applications/node.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  for (const pathPair of filePairs) {
+    // TODO: initially generate as DepGraph
+    const parserResult = await lockFileParser.buildDepTree(
+      filePathToContent[pathPair.manifest],
+      filePathToContent[pathPair.lock],
+      shouldIncludeDevDependencies,
+      pathPair.lockType,
```
shaimendel, line 30: this can be npm or yarn?

lib/types.ts
ivanstanev: But how does scanning a Node app produce `packageFormatVersion` or a `targetOS` object? How is `docker` also required for such results?

ivanstanev: I'm also following the interfaces in the `lockfile-parser`, otherwise we get a significant mismatch

lib/analyzer/applications/node.ts
ivanstanev: Yes
```js
export declare enum LockfileType {
    npm = "npm",
    yarn = "yarn"
}
```

lib/types.ts
Shesekino: it's not, we need to distinguish the type for the OS scan and the "app" scan.
the OS scan must have these fields, the "app" scan should have... whatever it has to have (for you to discover 😛 )
at least I think it's better to distinguish these two types...

lib/types.ts
```diff
@@ -46,7 +48,7 @@ export interface PluginResponse {
 
 export interface ScannedProjectCustom {
   packageManager: string; // actually SupportedPackageManagers; in the CLI
-  depTree: DepTree;
+  depTree: DepTree | PkgTree;
```
Shesekino, line 51: could we have the type here always be just ```DepTree```? if ```PkgTree``` has discrepancies then adapt them to this format? better to have the return type deterministic if possible, I think?

ivanstanev: The PkgTree is pretty much DepTree but with all fields optional... (And missing `docker`, `targetOS`...) I find it really hard to adapt it unless I force-cast and skip all compiler checks

lib/analyzer/applications/node.ts
```diff
@@ -0,0 +1,118 @@
+import * as path from "path";
+import * as lockFileParser from "snyk-nodejs-lockfile-parser";
+
+import { ScannedProjectCustom } from "../../types";
+import { FilePathToContent } from "./types";
+
+interface ManifestLockPathPair {
```
firecode9, line 7: ManifestLockPathPair has 3 items, so maybe call it ManifestLockPathTuple?

lib/types.ts
firecode9: is depTree a must? I mean going forward if we start adding non-dep-tree stuff the depTree wont have a lot of meaning

Shesekino: I think it's a must from the CLI's perspective?

lib/analyzer/applications/node.ts
Shesekino: the ```pair``` here refers to the manifest + lock files, not to the items in the interface, so tuple doesn't quite work here.

lib/analyzer/static-analyzer.ts
```diff
@@ -19,12 +19,21 @@ import {
   getAptFiles,
   getDpkgPackageFileContentAction,
 } from "../inputs/distroless/static";
+import {
```
Shesekino, line 22: regarding the commit message:
```feat: scan Node applications in static scanning```
-->
```feat: scan Node application dependencies in static scanning```

File: lib/types.ts (hunk 5 above)
Comment (firecode9) [reply lib/types.ts:51]: true. I think we would need to change it to be optional in the cli-interface as well. but maybe not part of this commit

File: lib/analyzer/applications/node.ts [hunk 8]
```diff
@@ -0,0 +1,118 @@
...
+      fileNamesGroupedByDirectory[directory] = [];
+    }
+    fileNamesGroupedByDirectory[directory].push(fileName);
+  }
+  return fileNamesGroupedByDirectory;
+}
+
+function stripUndefinedLabels(
+  parserResult: lockFileParser.PkgTree,
+): lockFileParser.PkgTree & { labels: Record<string, string> } {
```
Shesekino, line 104: the kind of types is usually a code smell I think... can we use one type?

lib/analyzer/static-analyzer.ts
```diff
@@ -105,12 +115,20 @@ export async function analyze(
 
   const binaries = getBinariesHashes(archiveLayers);
 
+  const applicationScanResults: ScannedProjectCustom[] = [];
```
Shesekino, line 118: consider renaming to ```applicationDependenciesScanResults``` or basically anything that helps us stop using "app" because we're not scanning applications, we're scanning application dependencies

lib/response-builder.ts
```diff
@@ -25,17 +25,44 @@ function buildResponse(
...
-        packageManager: plugin.packageManager,
-        depTree: pkg,
-      },
-    ],
+    scannedProjects: groupedScannedProjects,
   };
 }
 
+/**
+ * Group the projects for the App+OS feature.
```
Shesekino, line 51: maybe better to dereference the specific feature

lib/types.ts
Shesekino: I'm still troubled by this ☹️
maybe we should merge these two types? or is this insane? I have no idea.
maybe even just add a comment explaining what each of these two things mean for a later refactor to be made easier with more context?

lib/response-builder.ts
```diff
@@ -28,18 +28,41 @@ function buildResponse(
...
+ * By sharing the same fields in the meta object, projects can be treated as related.
+ */
+function groupScannedProjects(
+  imageName: string,
+  scannedProjects: types.ScannedProjectCustom[],
+): types.ScannedProjectCustom[] {
+  return scannedProjects.map((project) => {
+    project.meta = {
```
firecode9, line 59: I think it's better to check if meta is already defined, just in case we dont override anything

lib/response-builder.ts
```diff
@@ -28,18 +28,41 @@ function buildResponse(
...
 
+  const scannedProjects = [
+    {
+      packageManager: plugin.packageManager,
+      depTree: pkg,
+    },
+    ...applicationScans,
+  ];
+
+  const groupedScannedProjects = groupScannedProjects(
```
firecode9, line 39: I think groupScannedProjects is confusing, because we dont group them here, we just adding meta part, so maybe something like addMetaToScannedProjects?

PR #197, opened 2020-04-30
fix: more precise typing for analyseDockerfile()

Shesekino:
- [x] Ready for review

#### What does this PR do?

more precise typing for analyseDockerfile()
this function may only return a DockerFileAnalysis type, so there's no need for declaring it to be DockerFileAnalysis | undefined.

https://snyk.slack.com/archives/CGSNWAE76/p1588257666067100

### DIFF
--- a/lib/docker-file.ts
+++ b/lib/docker-file.ts
@@ -29,7 +29,7 @@ async function readDockerfileAndAnalyse(
 
 async function analyseDockerfile(
   contents: string,
-): Promise<DockerFileAnalysis | undefined> {
+): Promise<DockerFileAnalysis> {
   const dockerfile = DockerfileParser.parse(contents);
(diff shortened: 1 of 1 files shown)


ivanstanev:
If this is used for dynamic scanning, version 3.0.0+ doesn't work well with dynamic scanning (the app+OS part) 😬

Shesekino:
@ivanstanev not sure I'm following. is there a problem with merging this type change? it seems the function only returns that kind of object, never ```undefined```.

ivanstanev:
Just noting it down, in case we want to bump the plugin in other services so that they stop complaining about the undefined type 😛

Shesekino:
yes, it's definitely useless (at the moment) for that reason - but it's still nice to have, might as well merge that now :) thanks!

PR #198, opened 2020-05-03
[RUN-660] feat: support dockerfile analysis during static scan

jmickey:
- [x] Ready for review

#### What does this PR do?
Adds support for Dockerfile analysis when completing a static docker archive scan. This was a feature that was available previously when dynamically scanning the docker image.

The main advantage of this is that we can now retreive the base image used to build the container as long as the user provides us with the Dockerfile that was used to build it.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -18,30 +18,34 @@ export { inspect, dockerFile };
 const MAX_MANIFEST_FILES = 5;
 const debug = Debug("snyk");
 
-function inspect(
+async function inspect(
   root: string,
   targetFile?: string,
   options?: any,
 ): Promise<PluginResponse> {
   const targetImage = root;
 
+  const dockerfileAnalysis = await dockerFile.readDockerfileAndAnalyse(
+    targetFile,
+  );
+
   if (options && options.experimental) {
-    return experimentalAnalysis(targetImage, options);
+    return await experimentalAnalysis(targetImage, dockerfileAnalysis, options);
   }
 
   if (staticUtil.isRequestingStaticAnalysis(options)) {
-    return staticUtil.analyzeStatically(targetImage, options);
+    return await staticUtil.analyzeStatically(
+      targetImage,
+      dockerfileAnalysis,
+      options,
+    );
   }
 
-  return dockerFile
-    .readDockerfileAndAnalyse(targetFile)
-    .then((dockerfileAnalysis) => {
-      return analyzeDynamically(
-        targetImage,
-        dockerfileAnalysis,
-        getDynamicAnalysisOptions(options),
-      );
-    });
+  return await analyzeDynamically(
+    targetImage,
+    dockerfileAnalysis,
+    getDynamicAnalysisOptions(options),
+  );
 }
 
 async function analyzeDynamically(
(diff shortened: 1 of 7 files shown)


jmickey:
TODO: Add addition tests

lib/response-builder.ts
```diff
@@ -9,7 +10,7 @@ export { buildResponse };
 function buildResponse(
   runtime: string | undefined,
   depsAnalysis,
-  dockerfileAnalysis,
+  dockerfileAnalysis: DockerFileAnalysis | undefined,
```
Shesekino, line 13: nice!

test/system/static.test.ts
```diff
@@ -495,3 +495,26 @@ test("static and dynamic scanning results are aligned", async (t) => {
...
+    dockerfilePath,
+    pluginOptionsWithDockerSave,
+  )
+
+  t.equals(
+    pluginResultStatic.scannedProjects[0].depTree.docker.baseImage,
+    "debian:stretch-slim",
+    "base image matches expected",
+  );
+});
```
Shesekino, line 520: newline 😛

lib/index.ts
```diff
@@ -18,30 +18,28 @@ export { inspect, dockerFile };
...
   targetFile?: string,
   options?: any,
 ): Promise<PluginResponse> {
   const targetImage = root;
 
+  const dockerfileAnalysis = await dockerFile.readDockerfileAndAnalyse(targetFile);
+
   if (options && options.experimental) {
-    return experimentalAnalysis(targetImage, options);
+    return experimentalAnalysis(targetImage, dockerfileAnalysis, options);
```
Shesekino, line 31: how do you feel about adding an ```await``` for each ```return``` statement here?

is idea is having a stricter standards, even when unnecessary, so that future changes will be a bit safer to make (let's say tomorrow someone wants to wrap one of the calls in try/catch clauses and they forget adding ```await``` etc.)
more light reading I just enjoyed https://jakearchibald.com/2017/await-vs-return-vs-return-await/

test/system/static.test.ts
jmickey: BAH. I've configured my vs-code to fix this automatically from now on so shouldn't be an issue in the future.

test/system/static.test.ts
```diff
@@ -495,3 +495,48 @@ test("static and dynamic scanning results are aligned", async (t) => {
...
+    },
+  };
+
+  const pluginResultStatic = await plugin.inspect(
+    thisIsJustAnImageIdentifierInStaticAnalysis,
+    dockerfilePath,
+    pluginOptionsWithDockerSave,
+  );
+
+  const results = pluginResultStatic.scannedProjects[0].depTree.docker;
```
Shesekino, line 518: how about adding a test & assertion: "Dockerfile analysis results are identical between static and dynamic scans".
is that reasonable or completely redundant? from an API point of view

jmickey: Both dynamic and static scanning use the same calls to build the Dockerfile analysis and to build the response, so tbh I am not sure if there is value in comparing them?

PR #200, opened 2020-05-05
Feat/manifest files

Shesekino:
- [x] Ready for review

#### What does this PR do?

include manifestFiles found with globs on the dynamic image scans

### DIFF
--- a/lib/response-builder.ts
+++ b/lib/response-builder.ts
@@ -37,6 +37,15 @@ function buildResponse(
     ...applicationDependenciesScanResults,
   ];
 
+  if (manifestFiles.length > 0) {
+    scannedProjects.push({
+      scanType: types.ScanType.ManifestFiles,
+      data: manifestFiles,
+      packageManager: "PLEASE DON'T USE THIS",
+      depTree: { dependencies: {} },
+    } as types.ScannedProjectManifestFiles);
+  }
+
   const scannedProjectsWithImageName = assignImageNameToScannedProjectMeta(
     pkg.name,
     scannedProjects,
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -55,7 +55,7 @@ export interface ScannedProjectCustom {
    * a future change should be mindful but find a way to unify them if possible.
    */
   depTree: DepTree | PkgTree;
-  targetFile?: string;
+  targetFile?: string; // currently used for application-dependencies scans
   meta?: any; // not to pollute with actual data; reserved for actual metadata
 }
 
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -39,8 +39,6 @@ export interface PluginResponseStatic extends PluginResponse {
   hashes: string[];
 }
 
-// NEW STUFF
-
 export interface PluginResponse {
   plugin: PluginMetadata;
   scannedProjects: ScannedProjectCustom[];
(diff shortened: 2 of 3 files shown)


lib/types.ts
```diff
@@ -55,8 +53,25 @@ export interface ScannedProjectCustom {
    * a future change should be mindful but find a way to unify them if possible.
    */
   depTree: DepTree | PkgTree;
-  targetFile?: string;
-  meta?: any;
+  targetFile?: string; // currently used for application-dependencies scans
+  meta?: any; // not to pollute with actual data; reserved for actual metadata
```
ivanstanev, line 57: 👍  🙂 👍

lib/types.ts
```diff
@@ -55,8 +53,25 @@ export interface ScannedProjectCustom {
...
+}
+
+enum ScanType {
+  DependencyTree,
+  DependencyGraph,
+  ManifestFiles,
+}
+
+export interface ScannedProjectExtended extends ScannedProjectCustom {
```
ivanstanev, line 66: WDYT of naming `ScannedProjectCustom` just `ScannedProject`? And then this new type `ScannedProjectExtended` could become `ScannedProjectCustom` or keep it `ScannedProjectExtended`

lib/types.ts
```diff
@@ -55,8 +53,25 @@ export interface ScannedProjectCustom {
...
+  scanType: ScanType;
+  // unknowingly structured data; determined by `scanType`
+  data: any;
```
ivanstanev, line 69: What about typing this as `unknown`? This disallows accessing properties of `data`, unless you assert its type or explicitly cast it (this way the consumer of the plugin response must actively check what they are receiving)

lib/types.ts
Shesekino: ```ScannedProjectCustom``` is used to reflect exactly what the CLI wants right now, I'd rather keep it this way
happy to rename ```ScannedProjectExtended``` to whatever makes sense to you

lib/types.ts
Shesekino: sounds brilliant I think :)

test/system/system.test.ts
```diff
@@ -357,7 +362,12 @@ test("inspect redis:3.2.11-alpine", (t) => {
     .then((imageId) => {
       expectedImageId = imageId;
       return plugin.inspect(img, dockerFileLocation, {
-        manifestGlobs: ["/etc/redhat-release*", "/etc/foo", "/nonexist/bar"],
+        manifestGlobs: [
+          "/etc/redhat-release*",
```
mladkau, line 366: We should also test globs which are not for a specific directory. So the DRA uses for example `"**/package.json"` or the exclusion of `**/node_modules/**`. Unfortunately, when it comes to manifest files, we can't assume in which path we will find them.

test/system/system.test.ts
```diff
@@ -411,6 +421,37 @@ test("inspect redis:3.2.11-alpine", (t) => {
         },
         "deps",
       );
+
+      const manifestFilesResult = res.scannedProjects.find((res) => {
+        if (!("scanType" in res)) {
+          return false;
+        }
+        const scannedProjectExtended = res as ScannedProjectExtended;
+        if (scannedProjectExtended.scanType === ScanType.ManifestFiles) {
```
mladkau, line 430: Just a simple return statement is fine here...
```suggestion
        return scannedProjectExtended.scanType === ScanType.ManifestFiles
```

test/system/system.test.ts
mladkau: The current list of globs which is used in production can be found here: https://github.com/snyk/docker-registry-agent/blob/develop/src/routes/discover.ts#L160

lib/response-builder.ts
```diff
@@ -37,6 +37,15 @@ function buildResponse(
     ...applicationDependenciesScanResults,
   ];
 
+  if (manifestFiles.length > 0) {
+    scannedProjects.push({
+      scanType: types.ScanType.ManifestFiles,
+      data: manifestFiles,
+      packageManager: "PLEASE DON'T USE THIS",
```
shaimendel, line 44: not sure what I think about that :)

test/system/system.test.ts
```diff
@@ -357,7 +362,13 @@ test("inspect redis:3.2.11-alpine", (t) => {
     .then((imageId) => {
       expectedImageId = imageId;
       return plugin.inspect(img, dockerFileLocation, {
-        manifestGlobs: ["/etc/redhat-release*", "/etc/foo", "/nonexist/bar"],
+        manifestGlobs: [
```
shaimendel, line 365: I think there are also globs for directories to ignore right?

PR #201, opened 2020-05-07
support retrieving arbitrary files for static scans

Shesekino:
- [x] Ready for review

#### What does this PR do?

first proposal for collecting arbitrary files (right now called manifestGlobs/manifestGlobsExclude) for the static scan.

https://snyksec.atlassian.net/browse/RUN-933

### DIFF
--- a/lib/analyzer/static-analyzer.ts
+++ b/lib/analyzer/static-analyzer.ts
@@ -20,6 +20,7 @@ import {
   getAptFiles,
   getDpkgPackageFileContentAction,
 } from "../inputs/distroless/static";
+import * as filePatternStatic from "../inputs/file-pattern/static";
 import {
   getNodeAppFileContent,
   getNodeAppFileContentAction,
@@ -31,6 +32,7 @@ import {
 } from "../inputs/rpm/static";
 import {
   ImageType,
+  ManifestFile,
   ScannedProjectCustom,
   StaticAnalysisOptions,
 } from "../types";
@@ -70,6 +72,16 @@ export async function analyze(
     staticAnalysisActions.push(getDpkgPackageFileContentAction);
   }
 
+  const checkForGlobs = shouldCheckForGlobs(options);
+  if (checkForGlobs) {
+    staticAnalysisActions.push(
+      filePatternStatic.generateExtractAction(
+        options.globsToFind.include,
+        options.globsToFind.exclude,
+      ),
+    );
+  }
+
   const dockerArchive = await getDockerArchiveLayersAndManifest(
     options.imagePath,
     staticAnalysisActions,
@@ -90,6 +102,12 @@ export async function analyze(
   let distrolessAptFiles: string[] = [];
   if (options.distroless) {
     distrolessAptFiles = getAptFiles(archiveLayers);
+  }
+
+  const manifestFiles: ManifestFile[] = [];
+  if (checkForGlobs) {
+    const matchingFiles = filePatternStatic.getMatchingFiles(archiveLayers);
+    manifestFiles.push(...matchingFiles);
   }
 
   let osRelease: OSRelease;
@@ -133,6 +151,7 @@ export async function analyze(
     binaries,
     imageLayers: dockerArchive.manifest.Layers,
     applicationDependenciesScanResults,
+    manifestFiles,
   };
 }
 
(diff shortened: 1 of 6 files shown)


lib/analyzer/static-analyzer.ts
```diff
@@ -70,13 +72,35 @@ export async function analyze(
     staticAnalysisActions.push(getDpkgPackageFileContentAction);
   }
 
+  // TODO AMIR need stricter checks for exclude globs
```
Shesekino, line 75: this is some input validation that should probably happen earlier

lib/inputs/file-pattern/static.ts
```diff
@@ -0,0 +1,64 @@
...
+import { ExtractAction, ExtractedLayers } from "../../extractor/types";
+import { streamToString } from "../../stream-utils";
+import { ManifestFile } from "../../types";
+
+function generatePathMatcher(
+  globsInclude: string[],
+  globsExclude: string[],
+): (filePath: string) => boolean {
+  return (filePath: string): boolean => {
+    let exclude = false;
```
Shesekino, line 13: logic copied from the Docker class that already has it, consider extracting common things?

lib/inputs/file-pattern/static.ts
```diff
@@ -0,0 +1,64 @@
...
+  return {
+    actionName: "find-files-by-pattern",
+    filePathMatches: generatePathMatcher(globsInclude, globsExclude),
+    callback: streamToString,
+  };
+}
+
+export function getMatchingFiles(
+  extractedLayers: ExtractedLayers,
+): ManifestFile[] {
```
Shesekino, line 43: it's totally unclear if I'm going for a generic retrieve file mechanism or specific to manifest files

test/system/static.test.ts
```diff
@@ -467,16 +486,26 @@ test("static and dynamic scanning results are aligned", async (t) => {
...
+  );
+  const osDepsStatic = pluginResultStatic.scannedProjects.find(
+    (res) => res.packageManager === "deb",
+  );
+  t.ok(osDepsStatic !== undefined);
+
   t.equals(
-    JSON.stringify(pluginResultDynamic.scannedProjects[0].depTree.dependencies),
-    JSON.stringify(pluginResultStatic.scannedProjects[0].depTree.dependencies),
+    JSON.stringify(osDepsDynamic!.depTree.dependencies),
```
Shesekino, line 507: can probably skip the stringify and use t.same instead

lib/inputs/file-pattern/static.ts
Shesekino: deciding against it, we intend to delete the Docker code at some point so it would be a redundant abstraction

lib/analyzer/static-analyzer.ts
```diff
@@ -70,6 +72,21 @@ export async function analyze(
     staticAnalysisActions.push(getDpkgPackageFileContentAction);
   }
 
+  const checkForGlobs =
```
shaimendel, line 75: might worths a function?

lib/analyzer/static-analyzer.ts
```diff
@@ -70,6 +72,21 @@ export async function analyze(
...
+    options &&
+    options.globsToFind &&
+    options.globsToFind.include &&
+    Array.isArray(options.globsToFind.include) &&
+    options.globsToFind.include.length > 0;
+  if (checkForGlobs) {
+    staticAnalysisActions.push(
+      filePatternStatic.generateExtractAction(
+        options.globsToFind.include,
+        options.globsToFind.exclude,
```
ivanstanev, line 85: Should we also assert that `exclude` is a `string[]`? Looks like we pass it all the way to `generatePathMatcher` and assume it's a `string[]`

Shesekino: I'm not sure.
I mean, yes, we should, but I don't want this area to be responsible for input validation.
the plugin gets plenty of options, but I don't think it asserts their correctness at any point... it may be a valid approach ("don't misuse the options, or prepare to catch your own exceptions") but if we're opting to protect the library's caller then I want to invest less here and more in a more centralised module to check all options once.

lib/analyzer/static-analyzer.ts
Shesekino: sure

PR #210, opened 2020-05-22
Plugin supports more container archives for static scanning

mladkau:
- [x] Ready for review

#### What does this PR do?

Make the static scanning code a bit more robust so it accepts a wider range of image archives for static scanning.

Pull request container 2 distinct commits. First makes the code more robust the second one makes sure Manifest files are kept in Buffers rather than strings.

A new test fixture was added containing a container archive from lib pull. Use this for manual testing.


lib/extractor/layer.ts
```diff
@@ -29,9 +31,13 @@ export async function extractDockerArchive(
 
     tarExtractor.on("entry", async (header, stream, next) => {
       if (header.type === "file") {
+        const normalizedHeaderName = path.normalize(header.name);
```
mladkau, line 34: Normalizing paths here so `./foo/bar` and `foo/bar` do not mismatch.

lib/extractor/layer.ts
```diff
@@ -46,7 +52,9 @@ export async function extractDockerArchive(
 
     tarExtractor.on("error", (error) => reject(error));
 
-    createReadStream(dockerArchiveFilesystemPath).pipe(tarExtractor);
+    createReadStream(dockerArchiveFilesystemPath)
+      .pipe(gunzip())
```
mladkau, line 56: We now support gzip compressed layers.

lib/inputs/file-pattern/static.ts
```diff
@@ -34,7 +34,7 @@ export function generateExtractAction(
   return {
     actionName: "find-files-by-pattern",
     filePathMatches: generatePathMatcher(globsInclude, globsExclude),
-    callback: streamToString,
+    callback: streamToBuffer,
```
mladkau, line 37: The content of manifest files is not streamed into a buffer.

test/system/static.test.ts
```diff
@@ -35,6 +35,13 @@ test("static analysis builds the expected response", async (t) => {
     },
   };
 
+  const pluginOptionsWithLibPull = {
+    staticAnalysisOptions: {
+      imagePath: getFixture("lib-pull/nginx.tar"),
```
mladkau, line 40: I've added a new fixture for a lib-pull archive.

lib/inputs/file-pattern/static.ts
agatakrajewska: how come we don't want the manifest files as base64 encoded str any more?

mladkau: This is based on this comment from @Shesekino in slack: https://snyk.slack.com/archives/CBXS33GJY/p1589820258095800?thread_ts=1589385083.016900

... if you care about the context and want to go down the :rabbit: :hole: have a look at the whole thread - but be warned it is huge :sweat_smile:

lib/extractor/layer.ts
agatakrajewska: Is this going to ungzip gzipped stream and do nothing if it's not gzip??

mladkau: You are correct! Hence the interesting name of the library `gunzip-maybe` (see https://github.com/mafintosh/gunzip-maybe). 

From the project description: `Transform stream that gunzips its input if it is gzipped and just echoes it if not.`

lib/index.ts
```diff
@@ -165,8 +165,8 @@ async function getManifestFiles(
       return {
         name: path.basename(g),
         path: path.dirname(g),
-        contents: Buffer.from(contents[i].stdout).toString("base64"),
+        contents: Buffer.from(contents[i].stdout),
       };
     })
-    .filter((i) => i.contents !== "");
+    .filter((i) => i.contents.toString() !== "");
```
shaimendel, line 171: is there a more Buffer-y way to do it? for example `i.contents.length > 0`

lib/inputs/file-pattern/static.ts
```diff
@@ -48,14 +48,15 @@ export function getMatchingFiles(
       if (actionName !== "find-files-by-pattern") {
         continue;
       }
-      if (!(typeof extractedLayers[filePath][actionName] === "string")) {
-        throw new Error("expected string");
+
+      if (!Buffer.isBuffer(extractedLayers[filePath][actionName])) {
+        throw new Error("expected a buffer");
```
shaimendel, line 53: please continue and not throw :) not sure if this check is needed at all

lib/inputs/file-pattern/static.ts
```diff
@@ -48,14 +48,15 @@ export function getMatchingFiles(
...
       }
 
       manifestFiles.push({
         name: path.basename(filePath),
         path: path.dirname(filePath),
-        contents: extractedLayers[filePath][actionName] as string,
+        contents: extractedLayers[filePath][actionName] as Buffer,
```
shaimendel, line 59: why do you need the `as Buffer`?

shaimendel (commented):
Great job!

shaimendel:
also (not critical for now but for future work) we would appreciate the commits messages to explain the why and not the what :)

lib/index.ts
mladkau: Not sure what a "Buffer-y way" is. For this case both constructs are fine. I have no preference either way. I can change it.

lib/inputs/file-pattern/static.ts
mladkau: Could you please elaborate why the behaviour should be changed in line with this PR. Why now continue and not throw - which of my changes needs this change in behaviour?

lib/inputs/file-pattern/static.ts
mladkau: The values of extractedLayers are either string or Buffer - manifestFiles needs a Buffer for the contents field (see my change in types.ts) - as before, we need to be explicit what we are assigning here. This is the same as before - before you had to say it is a string.

mladkau:
Sure. I suggest to fix your "Contributing" rules then. The first point is about commit messages and the example does not reflect what you are saying. It should probably have a sentence about this.

lib/index.ts
shaimendel: I prefer to not call toString if not necessary, if it's a Buffer then let's use it as a Buffer :)

lib/inputs/file-pattern/static.ts
shaimendel: you're correct 👍 I thought you added the throw

mladkau:
Closing this PR for now as we couldn't agree on some details. See also: https://snyk.slack.com/archives/CGSNWAE76/p1590147147160200

I'll add new PR's with the details we did agree on.

PR #211, opened 2020-05-24
[RUN-587] feat: utilise docker daemon to save image to temp directory

jmickey:
- [x] Ready for review

#### What does this PR do?

Repaces the existing Docker save implementation with an alternative implementation that no longer relies on the presense of the Docker binary.

This PR utilises the [docker-modem](https://github.com/apocas/docker-modem) project to make calls directly the UNIX Docker socket API, generally located at `/var/run/docker.sock`. The docker-modem API returns a stream that we can pipe directly into a file write stream.

From what I can tell the previous implementation of the `Docker.save()` was being utilised within the experimental workflow, which is accessed via the `--experimental` flag within the CLI. When the image is not prefixed with `docker-archive:` then it branches to the "distroless" workflow - This workflow seem poorly named as it doesn't actually seem to matter if the image is distroless or not. I could be missing something though.

I haven't altered the existing (non-experimental) static analysis workflow as we haven't pushed this through to replace the dynamic scanning flow.

Start with the [docker.ts](https://github.com/snyk/snyk-docker-plugin/pull/211/files#diff-1a778f82722b9f884412efd52d3c46d9) file. Additional test have been added into the [docker.test.ts](https://github.com/snyk/snyk-docker-plugin/pull/211/files#diff-36abe35061e760a3bbe9c05b3d215723) file.

[RUN-587](https://snyksec.atlassian.net/browse/RUN-587)


jmickey:
Tests are failing because CircleCI doesn't provide access to the Docker socket with their containers. I doubt there is a way around this if we want to continue using Docker executors. We may need to switch to a machine executor instead. Alternatively I can mock the call, but that seems counter intuitive.

lib/docker.ts
```diff
@@ -96,6 +107,50 @@ class Docker {
   }
 
   public async save(targetImage: string, destination: string) {
+    const request = {
+      path: `/images/${targetImage}/get?`,
+      method: "GET",
+      isStream: true,
+      statusCodes: {
+        200: true,
```
Shesekino, line 115: why do we need these status codes? how does Modem react to them?

lib/docker.ts
```diff
@@ -96,6 +107,50 @@ class Docker {
...
+        stream.on("end", () => {
+          writeStream.end();
+        });
+
+        stream.pipe(writeStream);
+      });
+    });
+  }
+
+  public async saveCli(targetImage: string, destination: string) {
```
Shesekino, line 153: is anyone using the old implementation?

package.json
```diff
@@ -23,6 +23,7 @@
   "dependencies": {
     "@snyk/rpm-parser": "^1.1.0",
     "debug": "^4.1.1",
+    "docker-modem": "^2.1.3",
```
Shesekino, line 26: what do you think about moving to a fixed version?

test/lib/docker.test.ts
```diff
@@ -66,6 +67,55 @@ test("docker run", async (t) => {
...
+    const destination = path.join(os.tmpdir(), "image.tar");
+
+    t.tearDown(() => {
+      if (existsSync(destination)) {
+        unlinkSync(destination);
+      }
+    });
+
+    await docker.save(targetImage, destination);
+    t.true(existsSync(destination));
```
Shesekino, line 93: would it be sensible to assert anything else here beyond the file existing? for example the file size from the fixtures' dir is the same as the new file?

test/lib/docker.test.ts
```diff
@@ -66,6 +67,55 @@ test("docker run", async (t) => {
...
+  });
+
+  t.test("promise rejects when image doesn't exist", async (t) => {
+    const image = "someImage:latest";
+    await t.rejects(
+      docker.save(image, "tmp/image.tar"),
+      "server error",
+      "rejects with expected error",
+    );
```
Shesekino, line 102: how about an assertion no new file is created?

test/lib/docker.test.ts
```diff
@@ -66,6 +67,55 @@ test("docker run", async (t) => {
...
+
+  t.test(
+    "promise rejects when image cannot be written to destination",
+    async (t) => {
+      const image = "openjdk-test:latest";
+      const destination = "/somefolder/image.tar";
+      await t.rejects(
+        docker.save(image, destination),
+        { code: "ENOENT" },
+        "promise rejects with failed file open",
```
Shesekino, line 113: same suggestion - how about asserting no new file is created?

test/lib/docker.test.ts
```diff
@@ -66,6 +67,55 @@ test("docker run", async (t) => {
...
+    const image = "someImage:latest";
+    await t.rejects(
+      docker.save(image, "tmp/image.tar"),
+      "server error",
+      "rejects with expected error",
+    );
+  });
+
+  t.test(
+    "promise rejects when image cannot be written to destination",
```
Shesekino, line 106: great test case!

lib/docker.ts
jmickey: Status codes are utilised here: https://github.com/apocas/docker-modem/blob/master/lib/modem.js#L299. Basically if the code isn't true it will return an error that incorporates the string provided.

lib/docker.ts
jmickey: No

package.json
jmickey: Happy to move to a fixed version.

test/lib/docker.test.ts
jmickey: Yes, I was was considering doing a checksum comparison between the original tar and the new one.

test/lib/docker.test.ts
jmickey: Sure!

test/lib/docker.test.ts
jmickey: Sure!

jmickey:
> considering we're looking to remove anything that tightly couples us with docker - why did you choose to add to the Docker instead of creating a different module? (not implying that's better, probably what you did is better - just asking about your mind set)

@Shesekino - The implementation of this, unfortunately, still tightly couples us to Docker. The difference is that in this case we're relying on the existance of the daemon rather than access to the full Docker binary - so still an improvement. The socket path (`/var/run/docker.sock`) and API calls are both specific to Docker.

lib/docker.ts
Shesekino: maybe worth deleting it altogether then?

test/lib/docker.test.ts
```diff
@@ -1,11 +1,32 @@
...
 import * as path from "path";
 import * as sinon from "sinon";
 import { test } from "tap";
 import * as subProcess from "../../lib/sub-process";
 
 import { Docker } from "../../lib/docker";
 
+function getChecksum(path: string): Promise<string> {
+  return new Promise((res, rej) => {
+    const hash = crypto.createHash("sha1");
```
Shesekino, line 13: let's use sha256 here please, I'll elaborate on slack

test/lib/docker.test.ts
```diff
@@ -1,11 +1,32 @@
...
+import * as os from "os";
 import * as path from "path";
 import * as sinon from "sinon";
 import { test } from "tap";
 import * as subProcess from "../../lib/sub-process";
 
 import { Docker } from "../../lib/docker";
 
+function getChecksum(path: string): Promise<string> {
+  return new Promise((res, rej) => {
```
Shesekino, line 12: let's go with resolve/reject, I think it's more readable
also "res" is some times used to abbreviate "response"

PR #212, opened 2020-05-25
fix: allow one more type of rpm failures in dynamic scanning

shaimendel:
- [x] Ready for review

#### What does this PR do?
allow one more type of rpm failures in dynamic scanning

### DIFF
--- a/lib/inputs/rpm/docker.ts
+++ b/lib/inputs/rpm/docker.ts
@@ -18,6 +18,13 @@ export function getRpmDbFileContent(
       if (typeof stderr === "string" && stderr.indexOf("not found") >= 0) {
         return { stdout: "", stderr: "" };
       }
+
+      if (
+        typeof stderr === "string" &&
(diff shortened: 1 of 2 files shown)


lib/inputs/rpm/docker.ts
```diff
@@ -18,6 +18,13 @@ export function getRpmDbFileContent(
       if (typeof stderr === "string" && stderr.indexOf("not found") >= 0) {
         return { stdout: "", stderr: "" };
       }
+
+      if (
+        typeof stderr === "string" &&
+        stderr.toLowerCase().indexOf("no such") >= 0
```
Shesekino, line 24: how about adding to the string, if we know it's not just ```no such``` but actually ```No such file or directory```?
so longer string and without ```.toLowerCase()```

shaimendel: I personally prefer to remove the entire check and never throw :) so I preferred to make it more generic here

PR #216, opened 2020-05-28
fix: Support container archives with compressed layer tar archives

mladkau:
- [x] Ready for review

#### What does this PR do?

Adding support for compressed layer.tar files when scanning container archives. The decompression code is only triggered if the layers are actually compressed. Otherwise, the byte stream is just passed on.


mladkau:
Unfortunately, the test wouldn't work here as the tar file uses relative paths with `./` which will cause in this version an exception:
```
[2020-05-28T13:02:51.844Z] ERROR: docker-registry-agent-main/20963 on pauk: UNCAUGHT EXCEPTION!
    TypeError: Cannot read property 'Layers' of undefined
```
We could maybe modify an existing fixture and just manually compress the layers there?

ivanstanev:
Ohh that's interesting, wonder why it produced it with relative paths...
But then that probably means that this version cannot be used within libpull/dockerpull? Is there a follow up PR to this or something in libpull to account for it?

ivanstanev:
But to answer your question, I would really prefer we have a test for this so if you can do it then gzipping the layers of an existing fixture should work!

mladkau:
The reason for the exception is that the plugin uses not normalized paths. See the discussion linked in #210 - https://snyk.slack.com/archives/CGSNWAE76/p1590147147160200

If in the manifest.json you find `foo/bar.tar` and the tar file has the entry `./foo/bar.tar` then you have a missmatch which results in the exception.

mladkau:
I've added a fixture with compressed layers. Let me know if you want different names, paths, etc ...

PR #219, opened 2020-06-02
Feat/parse oci images

agatakrajewska:
- [x] Ready for review

#### What does this PR do?
We're adding support to`inspect` OCI images.

Commit by commit please :-) 

This PR is adding the code to parse the new format, not doing with it anything yet. 

[RUN-951](https://snyksec.atlassian.net/browse/RUN-951)


lib/extractor/layer.ts
```diff
@@ -53,6 +55,70 @@ export async function extractDockerArchive(
...
+          const promises = [
+            streamToJson(streamCopy1).catch(() => undefined),
+            extractImageLayer(streamCopy2, extractActions).catch(
+              () => undefined,
+            ),
+          ];
+          const [manifest, layer] = await Promise.all(promises);
+
+          const headerParts = header.name.split("/");
+          const headerName = `sha256:${headerParts[headerParts.length - 1]}`;
```
agatakrajewska, line 91: `header.name` in OCI archives is something like: `blob/sha256/xxxxxxxx`, so I'm extracting here `sha256:xxxxxxx`, which makes it super easy then to compare it to the layer identifier in manifests, which is of this exact format, wdyt??

lib/extractor/layer.ts
```diff
@@ -53,6 +55,70 @@ export async function extractDockerArchive(
...
+}> {
+  return new Promise((resolve, reject) => {
+    const tarExtractor: Extract = extract();
+
+    const layers: Record<string, ExtractedLayers> = {};
+    const manifests: OciArchiveManifest[] = [];
+    let imageIndex;
+
+    tarExtractor.on("entry", async (header, stream, next) => {
+      if (header.type === "file") {
```
shaimendel, line 73: not critical by any mean, just personal preference: instead of having 
```
if (header.type === "file") {
lots of lots of code
}
```
I prefer to have
```
if (header.type !== "file") {
    return;
}
rest of code
```

lib/extractor/layer.ts
```diff
@@ -53,6 +55,70 @@ export async function extractDockerArchive(
...
+        if (isImageIndexFile(header.name)) {
+          imageIndex = await streamToJson<OciImageIndex>(stream);
```
shaimendel, line 75: in same style: I prefer to `return` after this line and not have an `else` section

lib/extractor/layer.ts
```diff
@@ -53,6 +55,70 @@ export async function extractDockerArchive(
...
+        } else {
+          const streamCopy1 = new PassThrough();
+          const streamCopy2 = new PassThrough();
+          stream.pipe(streamCopy1);
+          stream.pipe(streamCopy2);
+
+          const promises = [
+            streamToJson(streamCopy1).catch(() => undefined),
```
shaimendel, line 83: why do we have the `.catch` instead of having a try-catch around the `Promise.all`?

lib/extractor/layer.ts
shaimendel: is it always sha256? or can be other hashes?

lib/extractor/layer.ts
```diff
@@ -53,6 +55,70 @@ export async function extractDockerArchive(
...
+            extractImageLayer(streamCopy2, extractActions).catch(
+              () => undefined,
+            ),
+          ];
+          const [manifest, layer] = await Promise.all(promises);
+
+          const headerParts = header.name.split("/");
+          const headerName = `sha256:${headerParts[headerParts.length - 1]}`;
+          if (manifest !== undefined) {
```
shaimendel, line 92: not sure what's my opinion about that you try to parse the file as two options, one throws and one not. why not going with the other option of going by the actual format and understanding which files are layers and which not?

lib/extractor/layer.ts
```diff
@@ -135,6 +201,48 @@ function getLayersContentAndArchiveManifest(
...
+  imageIndex: OciImageIndex,
+  manifestCollection: OciArchiveManifest[],
+): OciArchiveManifest {
+  if (imageIndex.manifests.length === 1) {
+    return manifestCollection[0];
+  }
+
+  const manifest = imageIndex.manifests.find(
+    (item) =>
+      item.platform?.architecture === "amd64" && item.platform.os === "linux",
```
shaimendel, line 236: why theres `?` on the left part and not also in the right part? is it intentional?

lib/extractor/layer.ts
```diff
@@ -135,6 +201,48 @@ function getLayersContentAndArchiveManifest(
...
+    manifest,
+  };
+}
+
+function getOciManifest(
+  imageIndex: OciImageIndex,
+  manifestCollection: OciArchiveManifest[],
+): OciArchiveManifest {
+  if (imageIndex.manifests.length === 1) {
+    return manifestCollection[0];
```
shaimendel, line 231: what happens if there is only one manifest and it is not linux amd64?

lib/extractor/layer.ts
ivanstanev: I think the two `consts` can be in the `if (layer !== undefined) {` flow

lib/extractor/layer.ts
agatakrajewska: 💯

lib/extractor/layer.ts
ivanstanev: >going by the actual format and understanding which files are layers and which not?

Do you mean looking at the manifest to determine this? I think because we have to process the archive 2 times - one to extract the index.json, and another to extract the files we want!

lib/extractor/layer.ts
```diff
@@ -135,6 +201,48 @@ function getLayersContentAndArchiveManifest(
...
+  // filter empty layers
+  // get the layers content without the name
+  // reverse layers order from last to first
+
+  // get manifest file first
+  const manifest = getOciManifest(imageIndex, manifestCollection);
+  const filteredLayers = manifest.layers
+    .filter((layer) => Object.keys(layers[layer.digest]).length !== 0)
+    .map((layer) => layers[layer.digest])
+    .reverse();
```
ivanstanev, line 218: Do we know if the layers are in order of how they appeared in the "Dockerfile"?

lib/extractor/layer.ts
ivanstanev: Because you want every copy to either succeed or fail, having a try-catch on the list of promises will fail all of them if one of them fails

lib/extractor/layer.ts
agatakrajewska: I've tried a few images and it was always `sha256`, BUT looking now its'a a digest and according to the spec: https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests it can be different hashes too. Hmmm, ok I think I can extract which hash it is from the path, which is `/blob/sha256/xxxxxx` in this case 🤔

lib/extractor/layer.ts
agatakrajewska: And when we catch an error we return `undefined`, which means that noting was read

lib/extractor/layer.ts
agatakrajewska: I mean, how we've tried it Shai, I couldn't get it working, I don't think you can in any way store streams, you have to process them immediately when you get the event. Maybe there is a way to do it in a similar way like I'm doing it now, just pass callbacks somehow. 
@ivanstanev how we were trying to approach it didn't mean scanning it twice, it was something like that: 
- create an object for streams
- on every tar layer event add stream chunk to the object, with the `header.name` identifier, also we get index.json
- now (further down in the code) we've got `index.json` so we know the manifests and I've tried to loop through the streams and call either manifest extract if it's manifest or layer extract if not. 
Of course that did not work cause streams were finished by that time. But maybe could do something like that with promises 🤔

lib/extractor/layer.ts
agatakrajewska: I think so from reading the docs: "The array MUST have the base layer at index 0. Subsequent layers MUST then follow in stack order"

lib/extractor/layer.ts
shaimendel: hmmm the downside here is that we parse layers of archs we don't really need. so we parse all the layers, even the ones that are not linux and amd64, and only filter them out in the end

agatakrajewska: But in the other case, we store those streams in the memory, so we could have 2GB worth of streams in the memory, is that not worse?

lib/extractor/layer.ts
agatakrajewska: The `&&` expression is evaluated left to right, so if platform is `null` or `undefined`, it's returned as `false` and if it's not `false` it means that platform is not `null`, so we don't need the `?`, I hope I'm right?! 😆

lib/extractor/layer.ts
```diff
@@ -135,6 +201,48 @@ function getLayersContentAndArchiveManifest(
...
+    return manifestCollection[0];
+  }
+
+  const manifest = imageIndex.manifests.find(
+    (item) =>
+      item.platform?.architecture === "amd64" && item.platform.os === "linux",
+  );
+
+  if (manifest === undefined) {
+    throw new Error("Unsupported type of CPU architecture or operating system");
```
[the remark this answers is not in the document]
agatakrajewska: 🤔 I will test it!

lib/extractor/layer.ts
agatakrajewska: 😱That should have been removed when I've added the `error`.

lib/extractor/layer.ts
agatakrajewska: Based on above documentation I've extracted format: `hash_name:hash_value`, let me know what you think!

agatakrajewska:
> It looks great overall! But I have two questions:
> 
> * Is there any reason we don't include tests? What do you think to include some tests?
> * I saw that `extractOciArchive` you've created was not called in anywhere. Is it supposed to be it? If it is, could you provide a context where it would be called, please? :)
> 
> Anyway, it's a great job! 🎉

Thanks!!! :-) 
It is actually not called anywhere, you are right! This is coming in the second PR, basically it meant a lot of refactoring, so it was better to deliver unused path first. 
Regarding tests, I had another task to add tests but I think it would be good to add them here maybe!

lib/extractor/layer.ts
agatakrajewska: Ok, so I've updated it (and tested) so if we've got a platorm, return undefined if it doesn't match and if we don't have the platform, return the first one (platform is optional and only present when multiple manifests)

lib/extractor/layer.ts
agatakrajewska: Actually I can't really do that! 😬 `return` leaves the block so we don't get any more events and when I tried to use `next()` I was getting some VERY unexpected results!

ivanstanev: Ah yes, instead of `return` you have to consume the stream! Like it was done here: https://github.com/snyk/snyk-docker-plugin/blob/50264f304f1fd4d2362db11813ab1617ade4bfe9/lib/extractor/layer.ts#L40-L41

I'm OK with leaving it as is anyway

lib/extractor/layer.ts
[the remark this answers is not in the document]
ivanstanev: I think we need to ensure all functions called in the `.on()` handlers are exception safe. If they do throw we can do `catch (error) { reject(error); }` so that we can properly return the exception to the caller of `await scanOciArchive()` (or whatever the name was 😛)

lib/extractor/layer.ts
ivanstanev: >and only filter them out in the end

You're right but I think it is still better performance-wise than the worst case (when the index.json is located right at the end of the tar-stream) of passing through the stream twice 😕 

Let's look at an example:
Untitled drawing (5)

In the current case we process:
- 3GB of data, reading every relevant file of every layer, regardless if it's the wrong OS. The majority of files are _not_ relevant (we have around 8 types that we care about)

In the "use index.json to understand the relevant layers" case we process:
- Almost 6GB of data: almost 3 GB to reach the index.json in order to parse it, and almost 3GB where we skip the first 2GB and process only 1GB for relevant files/layers

lib/extractor/layer.ts
```diff
@@ -53,6 +55,73 @@ export async function extractDockerArchive(
...
+): Promise<{
+  layers: ExtractedLayers[];
+  manifest: OciArchiveManifest;
+}> {
+  return new Promise((resolve, reject) => {
+    const tarExtractor: Extract = extract();
+
+    const layers: Record<string, ExtractedLayers> = {};
+    const manifests: Record<string, OciArchiveManifest> = {};
+    let imageIndex;
```
ivanstanev, line 70: This is a bit misleading because it has type `any`, it should be `OciImageIndex | undefined` and you will find that there is a potential undefined check that we don't handle! And it can result in an error where we call `imageIndex.manifests.find()`

Actually according to the spec the index.json is _optional_ https://github.com/opencontainers/image-spec/blob/master/image-index.md

Also the `manifests[]` may be of size 0. Maybe it's a bit too early to handle these cases but they are possible according to the spec and we should think how to act when it happens!

lib/extractor/layer.ts
shaimendel: yes :) great explanation! I would still put it in both just for the case someone will change the order of the condition or something like that, but really not critical

lib/extractor/layer.ts
shaimendel: should be verified with @garethr :)

lib/extractor/layer.ts
ivanstanev: OK we discussed an alternative that makes more sense, and we can definitely look to improve this in the future so we don't process irrelevant layers

lib/extractor/layer.ts
```diff
@@ -53,6 +55,73 @@ export async function extractDockerArchive(
...
+
+    tarExtractor.on("entry", async (header, stream, next) => {
+      if (header.type === "file") {
+        if (isImageIndexFile(header.name)) {
+          imageIndex = await streamToJson<OciImageIndex>(stream);
+        } else {
+          const streamCopy1 = new PassThrough();
```
shaimendel, line 77: not critical at all - I prefer to change `streamCopy1` to `jsonStream` and `streamCopy2` to `layerStream`

lib/extractor/layer.ts
agatakrajewska: I think I definitely need to handle image index being optional at this point though, wdyt? It is optional, so not doing it would be just not handling a case. 
In case of manifests array being size of 0, I'm wondering if it would be enough to handle it when we see that case, possibly add some logging here.

ivanstanev: I guess a manifest.length of 0 should mean we return no analysis? But if index.json is missing then probably we can't process the image either so we could throw a user error? Not sure, could check with Gareth as well

lib/extractor/layer.ts
agatakrajewska: done!

lib/extractor/layer.ts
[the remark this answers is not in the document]
agatakrajewska: So I actually have tried that before and result is exactly the same! I am getting an unhandled rejection, there's two ways you reject the promise right, one is by calling `reject()` and one when an error is thrown, so I think this is ok 🤔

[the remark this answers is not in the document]
agatakrajewska: So that code above gives exactly the same results as: 

```
      try {
      resolve(
        getOciLayersContentAndArchiveManifest(imageIndex, manifests, layers),
      );
      } catch (error) {
        reject(error);
      }
```

lib/extractor/layer.ts
agatakrajewska: I posted on slack and checked with Gareth :-)

lib/stream-utils.ts
```diff
@@ -48,3 +48,9 @@ export async function streamToHash(stream: Readable): Promise<string> {
     stream.pipe(hash);
   });
 }
+
+export async function streamToJson<T>(stream: Readable): Promise<T> {
+  return streamToString(stream).then((file) => {
```
Shesekino, line 53: async/await > promise.then 🙏

lib/extractor/layer.ts
```diff
@@ -53,6 +56,73 @@ export async function extractDockerArchive(
...
+        if (isImageIndexFile(header.name)) {
+          imageIndex = await streamToJson<OciImageIndex>(stream);
+        } else {
+          const jsonStream = new PassThrough();
+          const layerStream = new PassThrough();
+          stream.pipe(jsonStream);
+          stream.pipe(layerStream);
+
+          const promises = [
+            streamToJson(jsonStream).catch(() => undefined),
```
Shesekino, line 84: why are we okay with disregarding errors here?

lib/extractor/layer.ts
```diff
@@ -53,6 +56,73 @@ export async function extractDockerArchive(
   });
 }
 
+export async function extractOciArchive(
```
Shesekino, line 59: feels to me like this should either be some how merged with ```extractDockerArchive```
or we want this module/library refactored so there's different types of extractors, one for Docker and one for OCI images.
either way, let's try to reduce the cost of change here, taking into that the layer scan and extraction process may change in the future to support layer scanning & caching.
wdyt?

lib/extractor/layer.ts
```diff
@@ -148,6 +263,10 @@ function isManifestFile(name: string): boolean {
   return name === "manifest.json";
 }
 
+function isImageIndexFile(name: string): boolean {
```
Shesekino, line 266: what's the ```Image``` in the function name here?

agatakrajewska: This is referring to the `image index` file (documentation reference), which is a root manifest: https://github.com/opencontainers/image-spec/blob/master/image-index.md

lib/extractor/layer.ts
agatakrajewska: yes, exactly, if we get an error, we were parsing a layer, so we don't care about the error

agatakrajewska: I think I didn't say error enough times!

lib/stream-utils.ts
agatakrajewska: done

lib/extractor/layer.ts
agatakrajewska: That's actually a good idea, to have separate extractors. I'm following up this PR with another task, which is doing a fair bit of refactoring (because it will be using this code) so it makes sense to make the changes there. I will definitely address that.

lib/extractor/layer.ts
```diff
@@ -53,6 +56,73 @@ export async function extractDockerArchive(
...
+
+          const promises = [
+            streamToJson(jsonStream).catch(() => undefined),
+            extractImageLayer(layerStream, extractActions).catch(
+              () => undefined,
+            ),
+          ];
+          const [manifest, layer] = await Promise.all(promises);
+
+          // header format is /blobs/hash_name/hash_value
```
ivanstanev, line 91: Actually, any chance to assert that this is true and bail early if not? E.g.
```js
const headerParts = header.name.split("/");
if (!(headerParts.length === 3 && headerParts[0] === 'blobs')) {
    return; // or break; or however we need to continue
}
```

Is there a chance we could get an error if we received a file that is neither a layer nor a manifest?

lib/extractor/layer.ts
```diff
@@ -53,6 +56,73 @@ export async function extractDockerArchive(
...
+            layers[digest] = layer as ExtractedLayers;
+          }
+        }
+      }
+
+      stream.resume(); // auto drain the stream
+      next(); // ready for next entry
+    });
+
+    tarExtractor.on("finish", async () => {
```
ivanstanev, line 110: Wondering if this should be async, it doesn't seem to await anything

lib/extractor/layer.ts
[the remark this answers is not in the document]
agatakrajewska: It is working as expected after properly testing :-)

lib/extractor/layer.ts
agatakrajewska: I think you're right!

lib/extractor/layer.ts
```diff
@@ -135,12 +205,56 @@ function getLayersContentAndArchiveManifest(
...
+    item.platform
+      ? item.platform.architecture === "amd64" && item.platform.os === "linux"
+      : item,
+  );
+
+  if (manifestInfo === undefined) {
+    throw new Error("Unsupported type of CPU architecture or operating system");
+  }
+
+  return manifestCollection[manifestInfo.digest];
```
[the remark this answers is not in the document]
agatakrajewska: No, it's an empty object to start with! 👍

lib/extractor/layer.ts
```diff
@@ -53,6 +56,73 @@ export async function extractDockerArchive(
...
+          const [manifest, layer] = await Promise.all(promises);
+
+          // header format is /blobs/hash_name/hash_value
+          // we're extracting hash_name:hash_value format to match manifest digest
+          const headerParts = header.name.split("/");
+          const hashName = headerParts[1];
+          const hashValue = headerParts[headerParts.length - 1];
+          const digest = `${hashName}:${hashValue}`;
+          if (manifest !== undefined) {
+            manifests[digest] = manifest as OciArchiveManifest;
```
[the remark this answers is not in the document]
agatakrajewska: I mean, something like this would be better, so I don't put files that don't care about into that object at all: 
```
          if (manifest !== undefined && manifest.layers) {
            manifests[digest] = manifest as OciArchiveManifest;
          }
```
But it errors at `manifest.layers` because the type is unknown at this point, I'm having a look

lib/extractor/layer.ts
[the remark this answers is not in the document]
ivanstanev: I mean the return result, e.g. 
```js
const manifestCollection = {};
...
const result = manifestCollection['digestThatIsNotInCollection']; // undefined!
return result;
```

lib/extractor/layer.ts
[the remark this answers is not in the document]
ivanstanev: Yup I think your idea is right! You can use the TS compiler to help you as well and create a type guard like this:
```js
function isOciArchiveManifest(manifest: any): manifest is OciArchiveManifest {
  return manifest !== undefined && manifest.layers && manifest.layers.length >= 0;
}
```
and then use it something like:
```js
if (isOciArchiveManifest(manifest)) {
  manifests[digest] = manifest; // as OciArchiveManifest; you don't need the cast, TS now knows it's of type OciArchiveManifest!
}
```

PR #222, opened 2020-06-07
[RUN-994] feat: pull images directly from docker registry API

jmickey:
## Do not merge - need to make some dependencies public before

- [x] Ready for review

#### What does this PR do?

We want to remove the inherent reliance on the docker daemon when pulling container images from remote Docker v2 compliant repositories. This PR utilises the `snyk-docker-pull` lib to pull and save images to the local file system for scanning via the static scanning flow.

[RUN-994](https://snyksec.atlassian.net/browse/RUN-994)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,3 +10,13 @@ Snyk helps you find, fix and monitor for known vulnerabilities in your dependenc
 ## Snyk Docker CLI Plugin
 
 This plugin provides dependency metadata for Docker images. 
+
+## Running Tests
+
+To run tests the following environment variables need to be set:
+
+`DOCKER_HUB_PRIVATE_IMAGE`
+`DOCKER_HUB_USERNAME`
+`DOCKER_HUB_PASSWORD`
+
+`DOCKER_HUB_PRIVATE_IMAGE` should refer to an image that is hosted on Docker Hub but not available publicly. During CI test this is set to `snykgoof/dockergoof:alpine`.
--- a/lib/analyzer/image-inspector.ts
+++ b/lib/analyzer/image-inspector.ts
@@ -1,7 +1,12 @@
+import * as Debug from "debug";
+import * as path from "path";
+import * as tmp from "tmp";
 import { Docker, DockerOptions } from "../docker";
-import { DockerInspectOutput } from "./types";
+import { ArchiveResult, DockerInspectOutput, ImageDetails } from "./types";
 
-export { detect, pullIfNotLocal };
+export { detect, getImageArchive, extractImageDetails, pullIfNotLocal };
+
+const debug = Debug("snyk");
 
 async function detect(
   targetImage: string,
@@ -12,6 +17,124 @@ async function detect(
   return JSON.parse(info.stdout)[0];
 }
 
+/**
+ * In the case that an `ImageType.Identifier` is detected we need to produce
+ * an image archive, either by saving the image if it's already loaded into
+ * the local docker daemon, or by pulling the image from a remote registry and
+ * saving it to the filesystem directly.
+ *
+ * Users may also provide us with a URL to an image in a Docker compatible
+ * remote registry.
+ *
+ * @param {string} targetImage - The image to test, this could be in one of
+ *    the following forms:
+ *      * [registry/]<repo>/<image>[:tag]
+ *      * <repo>/<image>[:tag]
+ *      * <image>[:tag]
+ *    In the case that a registry is not provided, the plugin will default
+ *    this to Docker Hub. If a tag is not provided this will default to
+ *    `latest`.
+ * @param {string} [username] - Optional username for private repo auth.
+ * @param {string} [password] - Optional password for private repo auth.
+ */
+async function getImageArchive(
+  targetImage: string,
+  username?: string,
+  password?: string,
+): Promise<ArchiveResult> {
+  const docker = new Docker(targetImage);
+  const destination = tmp.dirSync({ unsafeCleanup: true });
+  const saveLocation: string = path.join(destination.name, "image.tar");
+
+  try {
+    // TODO: Eventually need to provide support for specifying a user-defined
+    // temporary directory.
+    await docker.save(targetImage, saveLocation);
+    return {
+      path: saveLocation,
+      removeArchive: destination.removeCallback,
+    };
+  } catch {
+    debug(
+      `${targetImage} does not exist locally, proceeding to pull image from remote registry.`,
+    );
+  }
+
+  if (await Docker.binaryExists()) {
+    try {
+      if (username || password) {
+        debug(
+          "using local docker binary credentials. the credentials you provided will be ignored",
+        );
+      }
+
+      await docker.pullCli(targetImage);
+      await docker.save(targetImage, saveLocation);
+      return {
+        path: saveLocation,
+        removeArchive: destination.removeCallback,
+      };
+    } catch (err) {
+      debug(`couldn't pull ${targetImage} using docker binary: ${err}`);
+    }
+  }
+
+  destination.removeCallback();
+  const { hostname, imageName, tag } = await extractImageDetails(targetImage);
+
+  debug(
+    `Attempting to pull: registry: ${hostname}, image: ${imageName}, tag: ${tag}`,
+  );
+  const pullResult = await docker.pull(
+    hostname,
+    imageName,
+    tag,
+    username,
+    password,
+  );
+  return {
+    path: path.join(pullResult.stagingDir!.name, "image.tar"),
+    removeArchive: pullResult.stagingDir!.removeCallback,
+  };
+}
+
+async function extractImageDetails(targetImage: string): Promise<ImageDetails> {
+  let remainder: string;
+  let hostname: string;
+  let imageName: string;
+  let tag: string;
+
+  // We need to detect if the `targetImage` is part of a URL. Based on the Docker specification,
+  // the hostname should contain a `.` or `:` before the first instance of a `/` otherwise the
+  // default hostname will be used (registry-1.docker.io). ref: https://stackoverflow.com/a/37867949
+  const i = targetImage.indexOf("/");
+  if (
+    i === -1 ||
+    (!targetImage.substring(0, i).includes(".") &&
+      !targetImage.substring(0, i).includes(":") &&
+      targetImage.substring(0, i) !== "localhost")
+  ) {
+    hostname = "registry-1.docker.io";
+    remainder = targetImage;
+    [imageName, tag] = remainder.split(":");
+    imageName =
+      imageName.indexOf("/") === -1 ? "library/" + imageName : imageName;
+  } else {
+    hostname = targetImage.substring(0, i);
+    remainder = targetImage.substring(i + 1);
+    [imageName, tag] = remainder.split(":");
+  }
+
+  // Assume the latest tag if no tag was found.
+  tag = tag || "latest";
+
+  return {
+    hostname,
+    imageName,
+    tag,
+  };
+}
+
 async function pullIfNotLocal(targetImage: string, options?: DockerOptions) {
   const docker = new Docker(targetImage);
   try {
@@ -20,5 +143,5 @@ async function pullIfNotLocal(targetImage: string, options?: DockerOptions) {
   } catch (err) {
     // image doesn't exist locally
   }
-  await docker.pull(targetImage);
+  await docker.pullCli(targetImage);
 }
--- a/lib/docker.ts
+++ b/lib/docker.ts
@@ -1,3 +1,8 @@
+import {
+  DockerPull,
+  DockerPullOptions,
+  DockerPullResult,
+} from "@snyk/snyk-docker-pull";
 import * as Debug from "debug";
 import * as Modem from "docker-modem";
 import { eventLoopSpinner } from "event-loop-spinner";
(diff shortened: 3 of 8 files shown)


lib/analyzer/image-inspector.ts
```diff
@@ -12,6 +18,83 @@ async function detect(
...
+  // a `.` or `:` before the first instance of a `/` otherwise the
+  // default hostname will be used (registry-1.docker.io).
+  // ref: https://stackoverflow.com/a/37867949
+  const i = targetImage.indexOf("/");
+  if (
+    i === -1 ||
+    (!targetImage.substring(0, i).includes(".") &&
+      !targetImage.substring(0, i).includes(":") &&
+      targetImage.substring(0, i) !== "localhost")
+  ) {
```
ivanstanev, line 73: Do you think it's worth pulling this to some well-named function e.g. `includesContainerRegistry` or something like that?
And the repeating part `targetImage.substring(0, i)` could get a variable

lib/analyzer/image-inspector.ts
```diff
@@ -12,6 +18,83 @@ async function detect(
...
+    hostname = "registry-1.docker.io";
+    remainder = targetImage;
+    const authRequest = await needle(
+      "get",
+      `https://auth.docker.io/token?service=${hostname}&scope=repository:${
+        remainder.split(":")[0]
+      }:pull`,
+      { json: true },
+    );
+    auth = authRequest.body.token;
```
ivanstanev: Would this be better suited for the `pull-lib`? It feels a bit odd to be outside of the library that deals with auth and registries

lib/docker.ts
```diff
@@ -101,7 +106,27 @@ class Docker {
     ]);
   }
 
-  public async pull(targetImage: string) {
+  public async pull(
```
ivanstanev, line 109: Would it make sense to put this split of `pull` and `pullCli` in separate commits? Feels like it can be its own self-contained change

lib/docker.ts
```diff
@@ -113,6 +138,7 @@ class Docker {
       statusCodes: {
         200: true,
         400: "bad request",
+        404: "not found",
```
ivanstanev, line 141: What is the piece of code that returns 404 now?

lib/analyzer/image-inspector.ts
```diff
@@ -71,21 +73,37 @@ async function getArchiveLocation(
   ) {
     hostname = "registry-1.docker.io";
```
ivanstanev, line 74: Wondering if there's some way (or value in doing it?) to have a function capture all of this for us, so we can simply do:
```js
const { registryHostname, remainder /* not sure what this is a remainder of? */, imageName, imageTag } = extractDetailsFromTargetImage(targetImage);
```

test/lib/analyzer/image-inspector.test.ts
```diff
@@ -84,4 +84,28 @@ test("get image as an archive", async (t) => {
     t.true(dockerPullSpy.calledOnce, "image pulled from remote registry");
     t.true(fs.existsSync(archiveLocation), "image exists on disks");
   });
+
+  t.test("from remote registry with authentication", async (t) => {
+    const dockerPullSpy: sinon.SinonSpy = sinon.spy(Docker.prototype, "pull");
+    const targetImage = "snykgoof/dockerhub-goof:alpine";
+
+    const username = process.env.DOCKER_HUB_USERNAME;
+    const password = process.env.DOCKER_HUB_PASSWORD;
```
ivanstanev, line 93: Should we improve the README to say this is now a requirement in order to run the test suite?

lib/analyzer/image-inspector.ts
```diff
@@ -12,6 +18,102 @@ async function detect(
...
+  targetImage: string,
+  username?: string,
+  password?: string,
+): Promise<string> {
+  const docker = new Docker(targetImage);
+
+  try {
+    // TODO: Eventually need to provide support for specifying a user-defined
+    // temporary directory. Defaulting to `path.join(os.tmpdir(), "image.tar")`
+    // for now as this matches the output from the `snyk-docker-pull` lib.
```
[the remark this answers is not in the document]
jmickey: Oh yes, you're correct. I made a mental note to update this to use the `tmp` package that `snyk-docker-pull` uses, but I forgot! 😅

lib/analyzer/image-inspector.ts
```diff
@@ -12,6 +18,83 @@ async function detect(
...
+  let remainder: string;
+  let hostname: string;
+  let auth: string | undefined;
+
+  // We need to detect if the `targetImage` is part of a URL.
+  // Based on the Docker specification, the hostname should contain
+  // a `.` or `:` before the first instance of a `/` otherwise the
+  // default hostname will be used (registry-1.docker.io).
+  // ref: https://stackoverflow.com/a/37867949
+  const i = targetImage.indexOf("/");
```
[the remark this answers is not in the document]
jmickey: ???

lib/analyzer/image-inspector.ts
jmickey: Yes, ideally this would be handled in the `docker-registry-v2-client`. I have a PR open here: https://github.com/snyk/docker-registry-v2-client/pull/50 which should hopefully make this code redundant.

lib/analyzer/image-inspector.ts
jmickey: I was of two minds regarding this, mainly because over-abstraction to functions can make a codebase really difficult to follow.

lib/docker.ts
jmickey: https://github.com/snyk/snyk-docker-plugin/pull/222/files#diff-f17f8b9c3d9c0c016f5247574ebcee34R48

This call to `docker.save` could return a 404 if the container isn't already saved in the local docker daemon.

test/lib/analyzer/image-inspector.test.ts
jmickey: Definitely, this is on my todo list. I expected this PR to change still as it's a WIP.

lib/analyzer/image-inspector.ts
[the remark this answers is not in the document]
ivanstanev: Haha I was thinking of a more clever name than `i` but probably that's an overkill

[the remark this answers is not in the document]
jmickey: Oh right! This is a habit I've picked up from Go, where it's much more "idiomatic" to use shorter variable names to keep code cleaner. Happy to change it though, I don't have strong feelings either way :D

lib/analyzer/image-inspector.ts
```diff
@@ -12,6 +18,83 @@ async function detect(
...
+ * @param {string} [password]
+ */
+async function getArchiveLocation(
+  targetImage: string,
+  username?: string,
+  password?: string,
+): Promise<string> {
+  const docker = new Docker(targetImage);
+
+  try {
```
shaimendel, line 43: what is the purpose of this try-catch? to use the docker binary if it exists?

lib/analyzer/image-inspector.ts
```diff
@@ -12,6 +18,83 @@ async function detect(
...
+  // default hostname will be used (registry-1.docker.io).
+  // ref: https://stackoverflow.com/a/37867949
+  const i = targetImage.indexOf("/");
+  if (
+    i === -1 ||
+    (!targetImage.substring(0, i).includes(".") &&
+      !targetImage.substring(0, i).includes(":") &&
+      targetImage.substring(0, i) !== "localhost")
+  ) {
+    hostname = "registry-1.docker.io";
```
shaimendel, line 74: the default registry (also for the auth URL) should in the end be also passed to the plugin from the outside (like we did in Kubernetes-monitor). just noting it :)

jmickey: @shaimendel Not really understanding what you're saying here?

lib/analyzer/image-inspector.ts
jmickey: No, it's checking if the image is available from the local docker daemon before we reach out to the remote registry to download it.

lib/analyzer/image-inspector.ts
ivanstanev: I think Shai's referring to when we add support to `registries.conf`!
Registries.conf allows us to override the default registry from which to pull images, by default it's Docker, but Red Hat have broken this and on OpenShift the default registry is registry.access.redhat.io 😕 Hence we would need a way to override this in the plugin. (Maybe not for this PR but when we _do_ add support for registries.conf!)

lib/analyzer/image-inspector.ts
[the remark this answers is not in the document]
shaimendel: @jmickey this is fixed right? just to make sure I understand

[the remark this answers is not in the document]
jmickey: @shaimendel Correct. The code is now using the tmp package to assign a temp dir. The comment needs to be updated, removing:

```
// Defaulting to `path.join(os.tmpdir(), "image.tar")`
// for now as this matches the output from the `snyk-docker-pull` lib.
```

lib/docker.ts
```diff
@@ -23,6 +28,17 @@ const SystemDirectories = ["dev", "proc", "sys"];
...
+    return new Promise<boolean>(async (resolve, reject) => {
+      try {
+        await subProcess.execute("docker", ["version"]);
+        resolve(true);
+      } catch (e) {
+        resolve(false);
+      }
+    });
+  }
+
```
ivanstanev, line 41:
```suggestion
  public static async binaryExists(): Promise<boolean> {
    try {
      await subProcess.execute("docker", ["version"]);
      return true;
    } catch (e) {
      return false
    }
  }

```

lib/docker.ts
```diff
@@ -101,7 +117,23 @@ class Docker {
     ]);
   }
 
-  public async pull(targetImage: string) {
+  public async pull(
+    registry: string,
+    repo: string,
+    tag: string | number,
```
ivanstanev, line 123: Why `string | number`, surely only `string` works?

lib/analyzer/image-inspector.ts
```diff
@@ -12,6 +17,137 @@ async function detect(
   return JSON.parse(info.stdout)[0];
 }
 
+export interface ArchiveResult {
+  path: string;
+  removeCallback(): void;
```
ivanstanev, line 22: Could we name this `removeArchive()`, I think callback may be misleading since we don't provide the remove function ourselves but the interface returns it to us!

test/lib/analyzer/image-inspector.test.ts
ivanstanev: @shaimendel in case this was missed!

test/lib/analyzer/image-inspector.test.ts
```diff
@@ -36,3 +39,140 @@ test("image id", async (t) => {
...
+        tag: "1.18",
+      },
+    },
+    "calico/cni:release-v3.14": {
+      expected: {
+        hostname: "registry-1.docker.io",
+        imageName: "calico/cni",
+        tag: "release-v3.14",
+      },
+    },
```
ivanstanev, line 72: Could we also add a test for the localhost case and for the registry-with-port case?

README.md
```diff
@@ -10,3 +10,13 @@ Snyk helps you find, fix and monitor for known vulnerabilities in your dependenc
...
+
+## Running Tests
+
+To run tests the following environment variables need to be set:
+
+`DOCKER_HUB_PRIVATE_IMAGE`
+`DOCKER_HUB_USERNAME`
+`DOCKER_HUB_PASSWORD`
+
+`DOCKER_HUB_PRIVATE_IMAGE` should refer to an image that is hosted on Docker Hub but not available publicly. During CI test this is set to `snykgoof/dockergoof:alpine`.
```
ivanstanev, line 22: Why do we need an env var here and not use `snykgoof/dockergoof:alpine` directly?

lib/analyzer/image-inspector.ts
```diff
@@ -12,6 +17,130 @@ async function detect(
...
+  const pullResult = await docker.pull(
+    hostname,
+    imageName,
+    tag,
+    username,
+    password,
+  );
+  return {
+    path: path.join(pullResult.stagingDir!.name, "image.tar"),
+    removeArchive: pullResult.stagingDir!.removeCallback,
```
ivanstanev, line 97: Can these be undefined? Is the pull library type lying to us and can we fix it in the library so we don't have to force a cast?

package.json
```diff
@@ -22,14 +22,17 @@
   },
   "dependencies": {
     "@snyk/rpm-parser": "^2.0.0",
+    "@snyk/snyk-docker-pull": "^3.1.3",
+    "@types/tmp": "^0.2.0",
```
ivanstanev, line 26: I think this type could be a devDependency

README.md
jmickey: 2 reasons mainly: 

1. This lib is open source, so on the off chance someone wants to contribute this enables them to test locally with their own private image.
2. If we add more tests that use this down the line we'll only have to change the value in 1 place.

lib/analyzer/image-inspector.ts
jmickey: The return type from the lib is `stagingDir: tmp.DirResult | null;`. I'm not sure if this can be changed, but myself or @shaimendel will check tomorrow.

shaimendel: I find this issue less critical, wdyt?

test/lib/analyzer/image-inspector.test.ts
shaimendel: added some, lmk what you think

lib/analyzer/image-inspector.ts
jmickey: Agree

lib/analyzer/image-inspector.ts
```diff
@@ -12,6 +17,130 @@ async function detect(
...
+  };
+}
+
+interface ImageDetails {
```
agranado2k, line 101: What do you think to move this interface to `types.ts`?

PR #224, opened 2020-06-08
Add OCI image support to static analysis

agatakrajewska:
- [x] Ready for review

#### What does this PR do?
Adding OCI image scanning to static analysis flow.

Refactored archive extractor into a docker image extractor and an oci image extractor.
Added an option to scan a oci image in the static scanning flow.

Commit by commit please :-) 

[RUN-952](https://snyksec.atlassian.net/browse/RUN-952)


lib/extractor/index.ts
```diff
@@ -8,21 +10,43 @@ import { DockerArchiveManifest, ExtractAction, ExtractedLayers } from "./types";
...
+        fileSystemPath,
+        extractActions,
+      );
+
+      return {
+        imageId: dockerExtractor.getImageIdFromManifest(dockerArchive.manifest),
+        manifestLayers: dockerExtractor.getManifestLayers(
+          dockerArchive.manifest,
+        ),
+        extractedLayers: processLayers(dockerArchive.layers),
```
ivanstanev, line 41: Could we name `processLayers` something more specific? Just the name doesn't convey what the processing is! Do we compute something with them? Do we reorder them? If you look at other functions we can tell really easily what they do e.g. `getManifestLayers` or `getImageIdFromManifest`

test/lib/extractor/image.test.ts
```diff
@@ -155,7 +146,7 @@ test("oci image extractor: extracted image content returned as expected", async
     },
   ];
 
-  const result = await extractOciArchive(
+  const result = await ociExtractor.extractArchive(
```
ivanstanev, line 149: Wondering if we should use `getArchiveLayersAndManifest()` here with `ImageType.OciArchive`?
I think that it may be best to test the most top-level function rather than `extractArchive` which is a bit "internal" 🤷

lib/extractor/index.ts
agatakrajewska: Yeah, you're right :-) With every PR I try to sneak (snyk) one "process" something 😆 
As I can see we basically flatten the array to an object, so I'm going to rename this to `flattenExtractedLayersArray` wdyt?

test/lib/extractor/image.test.ts
agatakrajewska: Yup, I agree, as this file are actually a higher level tests 👌

lib/extractor/index.ts
ivanstanev: I think flattening is part of it, what it does is it gets the latest modification of a file from all the layers. Maybe `extractedLayersWithLatestFileModifications`? Uhhh maybe it's a bit too long to include in the name (naming is hard 😄)... Or just `layersWithLatestFileModifications`?

lib/analyzer/static-analyzer.ts
```diff
@@ -82,38 +81,41 @@ export async function analyze(
     );
   }
 
-  const dockerArchive = await getDockerArchiveLayersAndManifest(
+  const {
+    imageId,
+    manifestLayers,
+    extractedLayers,
+  } = await archiveExtractor.getArchiveLayersAndManifest(
```
ivanstanev, line 88: I realise this returns more than the layers and manifest but I'm happy with the name for now, if we find that we're returning even more data in the future then it will get a slight rename I think!

lib/extractor/types.ts
```diff
@@ -29,7 +35,7 @@ export interface OciArchiveLayer {
 
 export interface OciArchiveManifest {
   schemaVersion: string;
-  config: string;
+  config: { digest: string };
```
ivanstanev, line 38: Oops, do we have a coverage for this field? Actually do we need it?
Just noticing the type changed but no tests did!

agatakrajewska: So, how we use it, is to extract `imageId`, like this: 
```
export function getImageIdFromManifest(manifest: OciArchiveManifest): string {
  try {
    return manifest.config.digest;
  } catch (err) {
    throw new Error("Failed to extract image ID from archive manifest");
  }
}
```
So if the `config` field is missing for example, it will throw. 
As for test coverage, I've added tests for image id, which covers that prop, like here: https://github.com/snyk/snyk-docker-plugin/pull/224/files#diff-745b4ac01a23bcadb4066e4c8fc69350R113
Do you think that's enough?

lib/analyzer/static-analyzer.ts
agatakrajewska: Yes is does, but I just didn't know what would be a better name 😆 Because it just essentially returns extraction result which is layers, some info from the manifest and imageId. I think I'm fine with it for now too as it's still related to manifest data & layers, sort of

lib/extractor/index.ts
agatakrajewska: Naming is very hard!! But I realised I did not quite understand what that function was doing, thanks!! 👍 I updated to `layersWithLatestFileModifications` which makes the most sense I think 😄

lib/extractor/types.ts
ivanstanev: Ah yes amazing, I missed it! Thanks 🙏

lib/extractor/index.ts
```diff
@@ -1,28 +1,56 @@
...
-): Promise<{
-  layers: ExtractedLayers;
-  manifest: DockerArchiveManifest;
-}> {
-  const dockerArchive = await extractDockerArchive(
-    fileSystemPath,
-    extractActions,
-  );
+): Promise<ExtractionResult> {
+  switch (imageType) {
```
agranado2k, line 18: [Minor point]
I think we could avoid some duplicated code here if we created `extractor` constant. What do you think?

```ts
...
  const extractor;
  switch (imageType) {
     case ImageType.OciArchive:
        extractor = ociExtractor;
     default:
        extractor = dockerExtractor;
  }

  const archive = await extractor.extractArchive(
     fileSystemPath,
     extractActions,
   );

   return {
       imageId: extractor.getImageIdFromManifest(archive.manifest),
       manifestLayers: extractor.getManifestLayers(
           archive.manifest,
       ),
       extractedLayers: flattenExtractedLayersArray(archive.layers),
   };
```

lib/extractor/oci-archive/index.ts
```diff
@@ -0,0 +1,15 @@
+import { OciArchiveManifest } from "../types";
+
+export { extractArchive } from "./layer";
```
agranado2k, line 3: [Minor point]
I think I'm missing something. Are we using this extractArchive here?

lib/extractor/oci-archive/layer.ts
```diff
@@ -0,0 +1,145 @@
...
+            streamToJson(jsonStream).catch(() => undefined),
+            extractImageLayer(layerStream, extractActions).catch(
+              () => undefined,
+            ),
+          ];
+          const [manifest, layer] = await Promise.all(promises);
+
+          // header format is /blobs/hash_name/hash_value
+          // we're extracting hash_name:hash_value format to match manifest digest
+          const headerParts = header.name.split("/");
```
agranado2k, line 55: [Minor point]
To improve the code readability, maybe we could push complexity about how to get `hashName` and `hashValue` in a function. What do you think?

lib/extractor/oci-archive/layer.ts
```diff
@@ -0,0 +1,145 @@
...
+  if (!imageIndex) {
+    return manifestCollection[Object.keys(manifestCollection)[0]];
+  }
+
+  const manifestInfo:
+    | OciManifestInfo
+    | undefined = imageIndex.manifests.find((item) =>
+    item.platform
+      ? item.platform.architecture === "amd64" && item.platform.os === "linux"
+      : item,
```
jonnyowenpowell, line 127: The platform check

PR #225, opened 2020-06-08
Add OCI support in experimental flow

agatakrajewska:
- [x] Ready for review

#### What does this PR do?
Adding OCI image scanning to the experimental flow.

Commit by commit please :-) 

[RUN-952](https://snyksec.atlassian.net/browse/RUN-952)

### DIFF
--- a/lib/experimental.ts
+++ b/lib/experimental.ts
@@ -5,7 +5,7 @@ import * as path from "path";
 import { pullIfNotLocal } from "./analyzer/image-inspector";
 import { Docker } from "./docker";
 import { DockerFileAnalysis } from "./docker-file";
-import { getDockerArchivePath, getImageType } from "./image-type";
+import { getArchivePath, getImageType } from "./image-type";
 import * as staticModule from "./static";
 import { ImageType, PluginResponse } from "./types";
 
@@ -18,8 +18,8 @@ export async function experimentalAnalysis(
   const imageType = getImageType(targetImage);
   switch (imageType) {
     case ImageType.DockerArchive:
-      return dockerArchive(targetImage, dockerfileAnalysis);
-
+    case ImageType.OciArchive:
+      return localArchive(targetImage, imageType, dockerfileAnalysis);
     case ImageType.Identifier:
       return distroless(targetImage, dockerfileAnalysis, options);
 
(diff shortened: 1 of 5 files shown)


lib/experimental.ts
```diff
@@ -18,8 +18,9 @@ export async function experimentalAnalysis(
   const imageType = getImageType(targetImage);
   switch (imageType) {
     case ImageType.DockerArchive:
-      return dockerArchive(targetImage, dockerfileAnalysis);
-
+      return localArchive(targetImage, imageType, dockerfileAnalysis);
+    case ImageType.OciArchive:
+      return localArchive(targetImage, imageType, undefined);
```
ivanstanev, line 23: Should we pass in here the dockerfileAnalysis too?

test/lib/image-type.test.ts
```diff
@@ -15,19 +15,20 @@ test("getImageType() returns the expected transports", (t) => {
...
-test("getDockerArchivePath() returns the expected results", (t) => {
+test("getArchivePath() returns the expected results", (t) => {
   t.plan(2);
 
   t.same(
-    getDockerArchivePath("docker-archive:/tmp/nginx.tar"),
+    getArchivePath("docker-archive:/tmp/nginx.tar"),
     "/tmp/nginx.tar",
     "returns extracted path",
   );
```
ivanstanev, line 32: Could we also add a test with `getArchivePath("oci-archive:/tmp/nginx.tar"),`

test/lib/image-type.test.ts
```diff
@@ -15,19 +15,20 @@ test("getImageType() returns the expected transports", (t) => {
...
-  t.same(
-    getDockerArchivePath("bad-path"),
-    "",
-    "does not handle errors and returns empty path",
-  );
+  t.throws(() => getArchivePath("bad-path"), "throws when provided bad path");
```
ivanstanev, line 33: Wondering if we should also validate the thrown error, e.g. `Error('The provided archive path is missing image specific prefix, eg."docker-archive:" or "oci-archive:"')`

lib/experimental.ts
```diff
@@ -92,11 +95,12 @@ async function getStaticAnalysisResult(
   targetImage: string,
   archivePath: string,
   dockerfileAnalysis: DockerFileAnalysis | undefined,
+  imageType?: ImageType | undefined,
```
ivanstanev, line 98: Why undefined here, the `localArchive()` passes in `ImageType` as defined

lib/experimental.ts
agatakrajewska: Omg, I think I got this wrong? Can we have a dockerfile analysis here?

test/lib/image-type.test.ts
agatakrajewska: We can totally.

lib/experimental.ts
agatakrajewska: But in `distroless` flow we pass `undefined`: https://github.com/snyk/snyk-docker-plugin/pull/225/files/549edc58ae6eae664023543fdcd9d6cfb0feeb94#diff-a29e3cbf53069cd5ba49e11beb3700efL81

agatakrajewska: Actually, I looked into it and I introduced a bug! So before in distroless flow we were assuming `docker-archive`.  I have fixed that now and set `ImageType.DockerArchive` in `distroless()` function, so we are still assuming if `docker-archive:` or `oci-archive:` the default is docker archive.

lib/experimental.ts
ivanstanev: The analysis was already done earlier (first step of calling `inspect()`), just not sure what it's used for but I think it should be included in the function call!

lib/experimental.ts
```diff
@@ -82,6 +85,7 @@ export async function distroless(
       targetImage,
       archiveFullPath,
       dockerfileAnalysis,
+      ImageType.DockerArchive,
```
ivanstanev, line 88: Nice!

PR #227, opened 2020-06-11
Feat/add unsupported archive error handling

agatakrajewska:
- [x] Ready for review

#### What does this PR do?
Adding an extra error handling, when no layers found in image provided.


lib/extractor/docker-archive/layer.ts
```diff
@@ -83,6 +83,10 @@ function getLayersContentAndArchiveManifest(
     .map((layerName) => layers[layerName])
     .reverse();
 
+  if (filteredLayers.length === 0) {
+    throw new Error("We found no layers in the provided image");
+  }
```
ivanstanev, line 88: Do we have a test case for this, any way we can add coverage?

agatakrajewska: Yeees, was just thinking that! 👍

agatakrajewska: Done!

agatakrajewska:
> Great job! 🎉
> I was wondering if the commit type should be `fix` instead of `feat`, what do you think?

Yes, so the first commit is, but the second one, I think it should be `feat` as it changes the functionality of the code, I think! 🤔

PR #228, opened 2020-06-12
fix: draft commit

agranado2k:
- [ ] Ready for review


lib/extractor/layer.ts
```diff
@@ -20,7 +20,7 @@ export async function extractImageLayer(
     const tarExtractor: Extract = extract();
 
     tarExtractor.on("entry", async (headers, stream, next) => {
-      if (headers.type === "file") {
+      if (headers.type === "file" || headers.type === "symlink") {
```
agranado2k, line 23: This change didn't make a difference to get the same output as the `snyk test --docker <image>` command.

PR #231, opened 2020-06-16
fix: include support to CentOS for static OS detector

agranado2k:
The static OS detector didn't detect CentOS images and returned an unknown OS name. And because we can't find vulns for unknown OS, it was making the static analyse doesn't get any vulns for the scanned image based on CentOS OS.

- [x] Ready for review

#### What does this PR do?
 It includes the CentOS support for the OS detection function.

Starts looking at the test. We've included a new test to detect OS statically.

You test manually scanning a CentOS image with the `--experimental` option and get that the OS was detected correctly.

It was a bug reported that the statically analysis was getting 0 vulns vs the dynamically analysis was getting 17 on the CentOS 6 image. It happened because there was a bug in the statically analysis that could not identify the OS using the statically detection function and it was returned `unknown`. And we can't identify vulns for `unknown` OS.

[RUN-1005](https://snyksec.atlassian.net/browse/RUN-1005)


test/lib/analyzer/os-release-detector.test.ts
```diff
@@ -218,6 +221,37 @@ test("os release detection", async (t) => {
...
+test("os release detection statically", async (t) => {
+  const examples = {
+    "centos:7": {
+      dir: "centos",
+      imageType: ImageType.DockerArchive,
+      imagePath: path.join(
+        __dirname,
+        "../../fixtures/docker-archives/skopeo-copy/centos-6.tar",
+      ),
+      expected: { name: "centos", version: "6", prettyName: "" },
```
ivanstanev, line 233: The key is named centos:7 but the expected is centos version 6? 😕

agranado2k: My bad! It was copy and paste! It didn't interfere in the test itself. It just a key. I'm changing it!

PR #235, opened 2020-06-22
fix: Adding a flag to activate application scans

mladkau:
- [x] Ready for review

#### What does this PR do?
Adding a flag to the docker plugin to activate parser based application scanning.

Add a flag next to the experimental flag to the CLI

See also this slack update: https://snyk.slack.com/archives/CLB5WDQRF/p1592497289002300
I open this at first as a draft PR since product needs to still decide about the actual name for the flag.

https://snyksec.atlassian.net/browse/DC-820

### DIFF
--- a/lib/analyzer/static-analyzer.ts
+++ b/lib/analyzer/static-analyzer.ts
@@ -144,10 +144,12 @@ export async function analyze(
   const binaries = getBinariesHashes(extractedLayers);
 
   const applicationDependenciesScanResults: ScannedProjectCustom[] = [];
-  const nodeDependenciesScanResults = await nodeFilesToScannedProjects(
-    getNodeAppFileContent(extractedLayers),
-  );
-  applicationDependenciesScanResults.push(...nodeDependenciesScanResults);
+  if (options.appScan) {
+    const nodeDependenciesScanResults = await nodeFilesToScannedProjects(
+      getNodeAppFileContent(extractedLayers),
+    );
+    applicationDependenciesScanResults.push(...nodeDependenciesScanResults);
+  }
   return {
     imageId,
     osRelease,
--- a/lib/experimental.ts
+++ b/lib/experimental.ts
@@ -19,7 +19,7 @@ export async function experimentalAnalysis(
   switch (imageType) {
     case ImageType.DockerArchive:
     case ImageType.OciArchive:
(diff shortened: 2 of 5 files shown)


lib/analyzer/static-analyzer.ts
```diff
@@ -144,10 +144,12 @@ export async function analyze(
   const binaries = getBinariesHashes(extractedLayers);
 
   const applicationDependenciesScanResults: ScannedProjectCustom[] = [];
-  const nodeDependenciesScanResults = await nodeFilesToScannedProjects(
-    getNodeAppFileContent(extractedLayers),
-  );
-  applicationDependenciesScanResults.push(...nodeDependenciesScanResults);
+  if (options.appScan) {
```
mladkau, line 147: We use the name `appScan` also in the DRA and in Registry.

mladkau: The if statement follows the same pattern as the distroless logic.

mladkau:
Flag on the CLI has been cheanged to: `app-vulns` see: https://snyk.slack.com/archives/CLB5WDQRF/p1593610184000200?thread_ts=1592825658.001300&cid=CLB5WDQRF

I intend to block the JSON output in the CLI code - IMO the plugin should just provide the data without bothering how it is used.

lib/experimental.ts
```diff
@@ -49,6 +50,7 @@ async function localArchive(
     archivePath,
     dockerfileAnalysis,
     imageType,
+    options["app-vulns"],
```
agranado2k, line 53: A minor comment. Could we use `appScan` for the option? Or is there a reason to be different?

mladkau: Yes there is a reason. This is the command line options which is given to the CLI i.e. `snyk --docker --app-vulns test`

The flag itself was agreed (after a loooong discussion). The CLI passes these flags through to the plugin. The function `getStaticAnalysisResult` seems to be the best place to convert the CLI flag to the internal flag appScan... Let me know if we should do it someplace else ...

mladkau:
On the commit prefix. This PR doesn't really add new functionality but merely introduces a flag to switch on/off existing functionality. I don't have a strong opinion here - happy to change the prefix if you think it should be a `feat`...

PR #236, opened 2020-07-07
feat: allow to choose the image save path for downloaded image

agranado2k:
Currently the plugin saves the downloaded image in a tmp folder. Now we allow the user to a custom location on disk with the best storage characteristics to download an image from a registry.

- [x] Ready for review

#### What does this PR do?
 Includes the option `imageSavePath` that if is set indicate the path where the downloaded image should be save.

You should follow the flow when the flag `--experimental` is set.

You can create a manual test and pass in the options `imageSavePath` with where you'd like to save the image. Be aware that it's a temporary file and the image will be removed once it finishes.

 It's good to notice when the image doesn't exist in local and there is no docker binary, it will pull the via `snyk-docker-pull`. It was changed to do the same.

- [RUN-582](https://snyksec.atlassian.net/browse/RUN-582)
- [RUN-654](https://snyksec.atlassian.net/browse/RUN-654)

### DIFF
--- a/lib/analyzer/image-inspector.ts
+++ b/lib/analyzer/image-inspector.ts
@@ -1,8 +1,15 @@
 import * as Debug from "debug";
+import * as fs from "fs";
+import * as mkdirp from "mkdirp";
 import * as path from "path";
-import * as tmp from "tmp";
+
 import { Docker, DockerOptions } from "../docker";
-import { ArchiveResult, DockerInspectOutput, ImageDetails } from "./types";
+import {
+  ArchiveResult,
+  DestinationDir,
+  DockerInspectOutput,
+  ImageDetails,
+} from "./types";
 
 export { detect, getImageArchive, extractImageDetails, pullIfNotLocal };
 
@@ -15,6 +22,16 @@ async function detect(
   const docker = new Docker(targetImage, options);
   const info = await docker.inspectImage(targetImage);
   return JSON.parse(info.stdout)[0];
+}
+
+function cleanupCallback(imagePath: string, imageName: string) {
+  return () => {
+    const fullImagePath = path.join(imagePath, imageName);
+    if (fs.existsSync(fullImagePath)) {
+      fs.unlinkSync(fullImagePath);
+    }
+    fs.rmdirSync(imagePath);
+  };
 }
 
 /**
@@ -39,17 +56,21 @@ async function detect(
  */
 async function getImageArchive(
   targetImage: string,
+  imageSavePath: string,
   username?: string,
   password?: string,
 ): Promise<ArchiveResult> {
   const docker = new Docker(targetImage);
-  const destination = tmp.dirSync({ unsafeCleanup: true });
+  mkdirp.sync(imageSavePath);
+  const destination: DestinationDir = {
+    name: imageSavePath,
+    removeCallback: cleanupCallback(imageSavePath, "image.tar"),
+  };
   const saveLocation: string = path.join(destination.name, "image.tar");
 
   try {
-    // TODO: Eventually need to provide support for specifying a user-defined
-    // temporary directory.
     await docker.save(targetImage, saveLocation);
+
     return {
       path: saveLocation,
       removeArchive: destination.removeCallback,
@@ -79,22 +100,21 @@ async function getImageArchive(
     }
   }
 
-  destination.removeCallback();
   const { hostname, imageName, tag } = await extractImageDetails(targetImage);
-
   debug(
     `Attempting to pull: registry: ${hostname}, image: ${imageName}, tag: ${tag}`,
   );
-  const pullResult = await docker.pull(
+  await docker.pull(
     hostname,
     imageName,
     tag,
+    imageSavePath,
     username,
     password,
   );
   return {
-    path: path.join(pullResult.stagingDir!.name, "image.tar"),
-    removeArchive: pullResult.stagingDir!.removeCallback,
+    path: saveLocation,
+    removeArchive: destination.removeCallback,
   };
 }
 
--- a/lib/docker.ts
+++ b/lib/docker.ts
@@ -119,6 +119,7 @@ class Docker {
     registry: string,
     repo: string,
     tag: string,
+    imageSavePath: string,
     username?: string,
     password?: string,
   ): Promise<DockerPullResult> {
@@ -127,6 +128,7 @@ class Docker {
       username,
       password,
       loadImage: false,
+      imageSavePath,
     };
     return await dockerPull.pull(registry, repo, tag, opt);
   }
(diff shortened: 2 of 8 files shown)


lib/analyzer/image-inspector.ts
```diff
@@ -1,8 +1,16 @@
 import * as Debug from "debug";
+import * as fs from "fs";
+import * as fx from "mkdir-recursive";
```
shaimendel, line 3: why is it called fx?

lib/analyzer/image-inspector.ts
```diff
@@ -17,6 +25,19 @@ async function detect(
...
+function cleanupCallback(imagePath: string, imageName: string) {
+  return () => {
+    const baseDir = path.isAbsolute(imagePath) ? __dirname : ".";
+    const imageDir = path.resolve(baseDir, imagePath);
+    const fullImagePath = path.join(imageDir, imageName);
+    if (fs.existsSync(fullImagePath)) {
+      fs.unlinkSync(fullImagePath);
+    }
+
+    rmdirSyncRecursive(imageDir.split(path.sep));
```
shaimendel, line 37: do we want to do it? what if this is a folder that already existed before us?

shaimendel: why not leaving the folder as is and just delete the file?

lib/analyzer/utils.ts
```diff
@@ -0,0 +1,24 @@
+import * as fs from "fs";
+import * as path from "path";
+
+/**
+ *  Remove directory recursively for all path passed as arg. It will stop removing
+ *  when it has removed everything or when the directory is not empty.
+ *
+ * @param {string[]} dirPath
+ */
+export function rmdirSyncRecursive(dirPath: string[]): void {
```
shaimendel, line 10: lets not do that 😆  sounds bad.. what if the folder is `/var/tmp/blabla` - will we try to delete `var`?

lib/docker.ts
```diff
@@ -121,12 +121,14 @@ class Docker {
     tag: string,
     username?: string,
     password?: string,
+    imageSavePath?: string,
```
shaimendel, line 124: is it really optional?

package.json
```diff
@@ -22,12 +22,13 @@
   },
   "dependencies": {
     "@snyk/rpm-parser": "^2.0.0",
-    "@snyk/snyk-docker-pull": "^3.1.3",
+    "@snyk/snyk-docker-pull": "3.2.0",
```
shaimendel, line 25: don't forget to have the ^ back

test/system/experimental.test.ts
```diff
@@ -82,3 +82,54 @@ test("docker-archive image type throws on bad files", async (t) => {
     "throws when the provided path is a directory",
   );
 });
+
+test("static scan for Identifier type image (nginx:1.19.0)", async (t) => {
```
shaimendel, line 86: why is this test part of this PR?

lib/experimental.ts
```diff
@@ -65,8 +66,10 @@ export async function distroless(
     );
   }
 
+  const imageSavePath = options?.imageSavePath || tmp.dirSync().name;
```
shaimendel, line 69: is `tmp.dirSync().name;` how we used to create the tmp location before? what about the `({ unsafeCleanup: true });` part?

lib/analyzer/image-inspector.ts
agranado2k: I got `fx` from the example in lib [repo](https://github.com/hex7c0/mkdir-recursive). I can change it to `mkdir-recursive`.
However you mentioned that you're afraid because of the license error. I can change the lib for another, do you recommend some? This one I got from Magma recommendation when I was discussing the changes in the `snyk-docker-pull`.  I'm just using it because I can't use `fs.mkdirSync({recursive: true})`, this `recursive options is not available on Node 8.

lib/analyzer/utils.ts
agranado2k: I cannot disagree with you! 😆 
The idea is clean up everything. But It doesn't make sense to remove the folder. I'm changing it.

test/system/experimental.test.ts
agranado2k: There is no test for the flow without `docker-archive` or `oci-archive`. I created one to help me to follow the flow and include the option. It's giving me confidence that I'm not broken nothing.
Maybe there is this test in another place and I missed it. If it already exists I can delete this test.

lib/experimental.ts
agranado2k: Before, we were using `tmp.dirSync({ unsafeCleanup: true })`, to create it. And it returns an object with `name` (directory path) and `removeCallback` to clean it up. It was creating and cleaning everything for us. Now there is the option `imageSavePath`, and when it exists we need to create the folder, if it doesn't exist, and after clean it up. (we don't need to remove the folder as you point out above, only the image that was downloaded). 
However when there is no this option (`imageSavePath`) , we need to generate the tmp folder, the idea here is to keep the same behaviour and get only the name for this tmp folder (`tmp.dirSync().name`). it's because we're already going to create and clean it ourselves. So we don't need to use the `removeCallback`, and consequently we don't need the option `unsafeCleanup: true`.
Not sure if I could explain well.

lib/analyzer/image-inspector.ts
```diff
@@ -39,17 +55,21 @@ async function detect(
...
+  imageSavePath: string,
   username?: string,
   password?: string,
 ): Promise<ArchiveResult> {
   const docker = new Docker(targetImage);
-  const destination = tmp.dirSync({ unsafeCleanup: true });
+  mkdirp.sync(imageSavePath);
+  const destination: DestinationDir = {
+    name: imageSavePath,
+    removeCallback: cleanupCallback(imageSavePath, "image.tar"),
```
shaimendel, line 66: that's not the name of the image now, it doesn't delete the right file.
this also points to a missing test :)

agranado2k: Thanks to show me it! :)

agranado2k: I was reviewing the solution and it's better that the image name keeps that same, because on `snyk-docker-pull` it uses `image.tar`. And there is no way to customaise it, unless we make another change there.
However I got your idea to create a tmp folder in the path and save the image there. I think it has the same effect. Unless I'm missing something.

lib/analyzer/image-inspector.ts
agranado2k: I changed the lib because of the license complaints. I won't be accept by the CLI.

lib/analyzer/image-inspector.ts
shaimendel: it might could have been better to make `snyk-docker-pull` to get the full path of the image as a parameter and not only the folder :)
anyway - now we don't delete the folders we create... are we ok with that?

agranado2k: I agree with you. That would be better. However we're not ok to not delete the folder. Let's delete it too.

PR #240, opened 2020-07-23
chore: enforce release after tests, lint and build

agranado2k:
- [x] Ready for review

#### What does this PR do?
It requires the pipeline steps Lint, Build and Tests for Release Github step.

 It includes only the required options in the pipeline workflow.

- [RUN-1061](https://snyksec.atlassian.net/browse/RUN-1061)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,10 @@ commands:
     steps:
       - run:
           name: Notify Slack on failure
-          command: ./.circleci/slack-notify-failure.sh "${CIRCLE_JOB}"
+          command: |
+            if [[ "$CIRCLE_BRANCH" == "master" ]]; then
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -144,3 +144,7 @@ workflows:
             branches:
               only:
                 - master
+          requires:
+            - Lint
+            - Build
+            - Test
```
ivanstanev, line 150: I think we won't be notified on Slack if either of these steps fails on the `master` branch, `- notify_slack_on_failure` currently runs only for the "release to GitHub" job

agranado2k: You're completely right! I included the `notification_slack_on_failure` for all steps, but only when merged to `master` to avoid noise in the Slack channel.

PR #244, opened 2020-08-07
Feat/windows tests in cicd pipeline

ivanstanev:
- [x] Ready for review

#### What does this PR do?

Adds support for Windows in the plugin and adjusts the tests and CI pipeline accordingly.

Commit by commit 😉 

[Jira ticket RUN-1100](https://snyksec.atlassian.net/browse/RUN-1100)


test/lib/docker-file.test.ts
```diff
@@ -15,9 +15,9 @@ test("Dockerfile not supplied", async (t) => {
 });
 
 test("Dockerfile not found", async (t) => {
-  t.rejects(
+  await t.rejects(
     () => dockerFile.readDockerfileAndAnalyse("missing/Dockerfile"),
-    new Error("ENOENT: no such file or directory, open 'missing/Dockerfile'"),
+    undefined,
```
ivanstanev, line 20: I couldn't figure out what the error message is for windows so I just swapped it to undefined since we only care that it throws, if you think this is wrong maybe worth checking what error is returned on windows 😄  but otherwise I'm happy just to leave it as undefined

test/system/experimental.test.ts
```diff
@@ -121,15 +123,15 @@ test("static scan for Identifier type image (nginx:1.19.0)", async (t) => {
...
+    pluginResult.plugin.imageLayers.length,
+    5,
+    "Returns expected number of layers",
+  );
+  t.ok(
+    pluginResult.plugin.imageLayers.every((layer) =>
+      layer.endsWith("layer.tar"),
+    ),
+    "Every found layer has the correct name",
   );
```
ivanstanev, line 136: Since this test will run only on Windows and not on Linux, we probably don't need to do this - we can compare layers like before

test/windows/rpm.test.ts
```diff
@@ -0,0 +1,51 @@
+import * as path from "path";
```
ivanstanev, line 1: Maybe we don't need the rpm tests specifically, I think we already test those above for the OS + 2 apps test case (I see it already uses amazonlinux:2)

test/windows/static.test.ts
```diff
@@ -0,0 +1,269 @@
...
+import { DepTree, ImageType, PluginResponseStatic } from "../../lib/types";
+
+const getFixture = (fixturePath) =>
+  path.join(__dirname, "../fixtures/docker-archives", fixturePath);
+
+test("static analysis builds the expected response", async (t) => {
+  const thisIsJustAnImageIdentifierInStaticAnalysis = "node:doesnotexist";
+  const dockerfile = undefined;
+  const pluginOptionsWithSkopeoCopy = {
+    staticAnalysisOptions: {
```
ivanstanev, line 14: I think all the Windows test cases with `staticAnalysisOptions` are not needed: this is a path used only by `kubernetes-monitor` which won't ever run on Windows, so I think we don't need them 🙏

test/lib/docker-file.test.ts
agatakrajewska: Having a look! 🙂

test/system/experimental.test.ts
agatakrajewska: yeah, 100%!!

test/lib/docker-file.test.ts
agatakrajewska: Actually, we don't run this test on windows any more as decided on separate ones

package.json
```diff
@@ -17,6 +17,7 @@
     "lint": "prettier --check '{lib,test}/**/*.ts' && tslint --format stylish '{lib,test}/**/*.ts'",
     "format": "prettier --loglevel warn --write '{lib,test}/**/*.ts' && tslint --fix --format stylish '{lib,test}/**/*.ts'",
     "test": "npm run lint && npm run unit-test",
+    "test-windows": "prettier --check \"{lib,test}/**/*.ts\" && tap test/windows -R=spec --timeout=300",
```
shaimendel, line 20: should the `prettier` part be in here?

lib/image-type.ts
```diff
@@ -1,3 +1,4 @@
+import { normalize as normalizePath } from "path";
```
shaimendel, line 1: we do we sometimes use `normalize` as is and sometimes rename is to `normalizePath`?

test/windows/application-scans.test.ts
```diff
@@ -0,0 +1,81 @@
+import * as path from "path";
```
shaimendel, line 1: what's the difference in the windows specific tests? why do we need them?

package.json
agatakrajewska: well, actually I can just `npm run lint` instead 👍

test/windows/application-scans.test.ts
agatakrajewska: There isn't much difference, just normalising the paths, etc. We've decided to go this way, as it wasn't feasible to run all the tests on windows, lots of them are irrelevant and/or we didn't quite work out why they are failing (like the docker socket one). So the best solution for now was to have few dedicated tests, covering high level paths. So there's duplication but I think it's nicer than having a check if platform is windows in every test, wdyt?

lib/image-type.ts
agatakrajewska: I will unify that as use just `normalizePath` as it's more explicit! 👍

PR #245, opened 2020-08-13
feat: #project-envelope: add scan() API returning artifacts

ivanstanev:
TODO

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -1,4 +1,6 @@
 import { PkgTree } from "snyk-nodejs-lockfile-parser";
+import { DockerFileAnalysis } from "./docker-file";
+import { DockerFilePackages } from "./instruction-parser";
 
 export interface StaticAnalysisOptions {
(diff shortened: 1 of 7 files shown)


lib/types.ts
```diff
@@ -64,21 +75,61 @@ export interface ScannedProjectCustom {
...
-  ManifestFiles = "ManifestFiles",
+export interface ScanResult {
+  /** TODO: This would be OPTIONAL for ManifestFiles. */
+  artifacts: ScannedArtifact[];
+  /** Not to pollute with actual data; reserved for actual metadata. */
+  meta?: { [key: string]: any };
+}
+
+export interface ScannedArtifact {
+  type: "depTree" | "pkgTree" | "hashes" | "manifestFile";
```
ivanstanev, line 86: `depGraph` should be here too. Perhaps we want to stop using `depTree` or `pkgTree` and only have `depGraph`!

PR #249, opened 2020-08-24
Feat/add base image detection

agatakrajewska:
- [x] Ready for review

#### What does this PR do?

Adding base image detection support.

First commit is a small refactor to accommodate extracting another type of manifest.
 
Second commit we're adding rootfs array extraction from image config in docker archive. Extracted array is added to the plugin response as one of the PluginMetadata.

commit by commit :-) 

https://snyksec.atlassian.net/browse/RUN-1096

### DIFF
--- a/lib/response-builder.ts
+++ b/lib/response-builder.ts
@@ -83,6 +83,7 @@ function pluginMetadataRes(
     packageManager: depsAnalysis.packageManager,
     dockerImageId: depsAnalysis.imageId,
     imageLayers: depsAnalysis.imageLayers,
+    rootFs: depsAnalysis.rootFsLayers,
   };
 }
 
--- a/lib/analyzer/static-analyzer.ts
+++ b/lib/analyzer/static-analyzer.ts
@@ -90,6 +90,7 @@ export async function analyze(
     imageId,
     manifestLayers,
     extractedLayers,
+    rootFsLayers,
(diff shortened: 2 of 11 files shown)


lib/response-builder.ts
```diff
@@ -83,6 +83,7 @@ function pluginMetadataRes(
     packageManager: depsAnalysis.packageManager,
     dockerImageId: depsAnalysis.imageId,
     imageLayers: depsAnalysis.imageLayers,
+    rootFs: depsAnalysis.rootLayers,
```
shaimendel, line 86: why do we sometimes call it rootFs and sometimes rootLayers, while the "formal" name is rootFsLayers? 😆

shaimendel (commented):
great work, this is just an initial review, the real one will be made by Josh :)

lib/response-builder.ts
agatakrajewska: yeah, I was wondering if that wasn't confusing! I wanted to give it a more user friendly name passing it around, rather than "rootFs", so I named it "rootLayers" everywhere and the we just have "rootFs" object on the plugin response, I can rename, don't mind too much :-)

agatakrajewska: I've renamed it to "rootFsLayers"

jmickey (approved):

PR #254, opened 2020-09-04
Feat/add platform to docker pull

agatakrajewska:
- [x] Ready for review

#### What does this PR do?
We're adding ability to pull architecture specific docker image like: 'docker pull --platform=linux/arm64' and scan it.
As a first step the user will be able to scan an `arm` image if `docker` available on the host machine.

The `platform` format is the same as `docker pull` or `docker buildx` so it's familiar to the user.

https://snyksec.atlassian.net/browse/RUN-1158

### DIFF
--- a/lib/analyzer/image-inspector.ts
+++ b/lib/analyzer/image-inspector.ts
@@ -15,13 +15,20 @@ export { detect, getImageArchive, extractImageDetails, pullIfNotLocal };
 
 const debug = Debug("snyk");
 
+async function getInspectResult(
+  docker: Docker,
+  targetImage: string,
+): Promise<DockerInspectOutput> {
+  const info = await docker.inspectImage(targetImage);
+  return JSON.parse(info.stdout)[0];
+}
+
 async function detect(
   targetImage: string,
   options?: DockerOptions,
 ): Promise<DockerInspectOutput> {
   const docker = new Docker(targetImage, options);
-  const info = await docker.inspectImage(targetImage);
-  return JSON.parse(info.stdout)[0];
+  return getInspectResult(docker, targetImage);
 }
 
 function cleanupCallback(imagePath: string, imageName: string) {
@@ -53,12 +60,14 @@ function cleanupCallback(imagePath: string, imageName: string) {
  *    `latest`.
  * @param {string} [username] - Optional username for private repo auth.
  * @param {string} [password] - Optional password for private repo auth.
+ * @param {string} [platform] - Optional platform parameter to pull specific image arch.
  */
 async function getImageArchive(
   targetImage: string,
   imageSavePath: string,
   username?: string,
   password?: string,
+  platform?: string,
 ): Promise<ArchiveResult> {
   const docker = new Docker(targetImage);
   mkdirp.sync(imageSavePath);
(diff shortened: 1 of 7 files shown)


lib/analyzer/image-inspector.ts
```diff
@@ -69,6 +77,14 @@ async function getImageArchive(
   const saveLocation: string = path.join(destination.name, "image.tar");
 
   try {
+    const inspectResult = await getInspectResult(docker, targetImage);
+    if (
+      platform !== undefined &&
+      !isLocalImageSameArchitecture(platform, inspectResult.Architecture)
+    ) {
+      throw new Error("Pulling an image with updated architecture required");
```
shaimendel, line 85: throwing here is not ideal, but might be good enough for now

lib/analyzer/image-inspector.ts
```diff
@@ -155,6 +171,22 @@ async function extractImageDetails(targetImage: string): Promise<ImageDetails> {
...
 }
 
+function isLocalImageSameArchitecture(
+  platformOption: string,
+  inspectResultArchitecture: string,
+): boolean {
+  let platformArchitecture: string;
+  try {
+    // Note: this is using the same flag/input pattern as the new Docker buildx: eg. linux/arm64/v8
+    platformArchitecture = platformOption.split("/")[1];
```
shaimendel, line 181: are we sure that's the correct logic?

lib/analyzer/image-inspector.ts
```diff
@@ -155,6 +171,22 @@ async function extractImageDetails(targetImage: string): Promise<ImageDetails> {
...
+  } catch (error) {
+    debug(`Error parsing platform flag: '${error}'`);
+    return false;
+  }
+
+  return platformArchitecture === inspectResultArchitecture;
```
shaimendel, line 187: this check is case sensitive, are we ok with? do we want to support users that write Arm insteadd of arm? (product question)

agatakrajewska: Ah, great question, I'll ask Gareth.

lib/analyzer/image-inspector.ts
agatakrajewska: I can't find an official format anywhere, but in all examples for `docker pull` or `docker buildx` the format is <os>/<arch>/<variant,version> and based on @garethr 's slack: 

> Assuming we’re happy to support multiple architectures while we’re supporting ARM64, the valid values should be:
> linux/amd64
> linux/arm64
> linux/riscv64
> linux/ppc64le
> linux/s390x
> linux/386
> linux/arm/v7
> linux/arm/v6

lib/analyzer/image-inspector.ts
agatakrajewska: Yes, 100%! The `getImageArchive()` function is built based on throwing errors 😬 so for me not to throw an error, it would need a bit of refactoring, which I think would be great but maybe not a part of this task?

lib/analyzer/image-inspector.ts
agatakrajewska: Basically, we should just support valid values and throw an error if wrong value provided, like docker does.The error handling will be added.

lib/analyzer/image-inspector.ts
agatakrajewska: Maybe when adding error handling it will be a good time :-)

shaimendel (approved):

PR #255, opened 2020-09-07
Adding error handling for `docker --platform`

agatakrajewska:
Fixed getImageArchive() so the logic doesn't depend on the function throwing errors.

- [x] Ready for review

#### What does this PR do?

First commit refactors `getImageArchive()` so the logic doesn't depend on the function throwing errors, the code is hopefully more readable and it can be extended easier. 

`Update tests for ARM` commit: we're removing a test depending on an experimental docker feature and adding a test covering scanning an ARM based image.

And last commit we're adding the error handling for `docker --platform`. We're throwing only when we get a specific error from `docker pull`: `unknown operating system or architecture`, otherwise if `pull` or `save` fails for any other reason we want to fall back to pulling using our pull lib.

Commit by commit please!

https://snyksec.atlassian.net/browse/RUN-1162

### DIFF
--- a/lib/analyzer/image-inspector.ts
+++ b/lib/analyzer/image-inspector.ts
@@ -41,6 +41,83 @@ function cleanupCallback(imagePath: string, imageName: string) {
   };
 }
 
+async function pullWithDockerBinary(
+  docker: Docker,
+  targetImage: string,
+  saveLocation: string,
+  username: string | undefined,
+  password: string | undefined,
+  platform: string | undefined,
+): Promise<boolean> {
+  let pullAndSaveSuccessful = false;
+  try {
+    if (username || password) {
+      debug(
+        "using local docker binary credentials. the credentials you provided will be ignored",
+      );
+    }
+    await docker.pullCli(targetImage, { platform });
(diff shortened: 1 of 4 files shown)


lib/analyzer/image-inspector.ts
```diff
@@ -59,6 +59,10 @@ async function pullWithDockerBinary(
     await docker.save(targetImage, saveLocation);
   } catch (err) {
     debug(`couldn't pull ${targetImage} using docker binary: ${err}`);
+
+    if (err.stderr.includes("unknown operating system or architecture")) {
```
ivanstanev, line 63: stderr is always present in the `err` object?

ivanstanev (dismissed):
Looks good! I have only a tiny question 😄

lib/analyzer/image-inspector.ts
agatakrajewska: Ah good catch, thank you! This is the output coming back from `subProcess.execute` (so `docker.pullCli` in this case) but from `docker.save` we'll be getting a standard error. I've fixed that :-)

ivanstanev (approved):

PR #257, opened 2020-09-10
Feat/add platform to plugin result

agatakrajewska:
- [x] Ready for review

#### What does this PR do?
Add platform to plugin result.
We're adding platform information to static scanning result. It will return platform for either image pulled from container registry or local docker archive.

Commit by commit

https://snyksec.atlassian.net/browse/RUN-1155

### DIFF
--- a/test/system/application-scans.test.ts
+++ b/test/system/application-scans.test.ts
@@ -69,9 +69,14 @@ test("scanning a container image with 2 applications", async (t) => {
 
   t.ok(pluginResult.scannedProjects[0].meta, "os scan meta is not falsy");
   t.same(
-    pluginResult.scannedProjects[0].meta,
-    pluginResult.scannedProjects[1].meta,
-    "os scan meta and app meta are identical",
+    pluginResult.scannedProjects[0].meta.platform,
+    "linux/amd64",
+    "os scan meta includes platform information",
+  );
+  t.same(
+    pluginResult.scannedProjects[0].meta.imageName,
+    pluginResult.scannedProjects[1].meta.imageName,
+    "os scan meta matches app scan meta imageName",
   );
   t.same(
     pluginResult.scannedProjects[1].meta,
--- a/test/windows/application-scans.test.ts
+++ b/test/windows/application-scans.test.ts
@@ -69,9 +69,14 @@ test("scanning a container image with 2 applications", async (t) => {
 
   t.ok(pluginResult.scannedProjects[0].meta, "os scan meta is not falsy");
   t.same(
-    pluginResult.scannedProjects[0].meta,
-    pluginResult.scannedProjects[1].meta,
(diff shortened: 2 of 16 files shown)


lib/types.ts
```diff
@@ -41,6 +41,7 @@ export interface PluginMetadata {
   dockerImageId: string;
   imageLayers: string[];
   rootFs?: string[];
+  platform?: string;
```
ivanstanev, line 44: Just an FYI that as part of #project-envelope unfortunately the `PluginMetadata` no longer exists in the response from the plugin, so all the fields in this interface will find a new home 😄 https://gist.github.com/ivanstanev/7f894d9bc3e4611452eadce1fe2c6729#file-scan-ts-L13-L16

Is it possible to put it for example in `scannedProject[0].meta` as this will most likely be the place we will use in envelope?

agatakrajewska: Great, thanks! 👍

agatakrajewska: on it 👍

test/system/application-scans.test.ts
```diff
@@ -69,9 +69,9 @@ test("scanning a container image with 2 applications", async (t) => {
 
   t.ok(pluginResult.scannedProjects[0].meta, "os scan meta is not falsy");
   t.same(
-    pluginResult.scannedProjects[0].meta,
-    pluginResult.scannedProjects[1].meta,
-    "os scan meta and app meta are identical",
+    pluginResult.scannedProjects[0].meta.platform,
+    "linux/amd64",
+    "os scan meta includes platform information",
   );
```
ivanstanev, line 75: I think this was comparing that the `imageName` is contained in the meta, is it possible to also do this check?

```typescript
t.same(
  pluginResult.scannedProjects[0].meta.imageName,
  pluginResult.scannedProjects[1].meta.imageName,
  "os scan meta matches app scan meta imageName", // or something like this
);
```

test/windows/application-scans.test.ts
```diff
@@ -69,9 +69,9 @@ test("scanning a container image with 2 applications", async (t) => {
 
   t.ok(pluginResult.scannedProjects[0].meta, "os scan meta is not falsy");
   t.same(
-    pluginResult.scannedProjects[0].meta,
-    pluginResult.scannedProjects[1].meta,
-    "os scan meta and app meta are identical",
+    pluginResult.scannedProjects[0].meta.platform,
+    "linux/amd64",
+    "os scan meta includes platform information",
   );
```
ivanstanev, line 75: Same here 🙏

test/system/application-scans.test.ts
agatakrajewska: Sure think, makes sense! Done

ivanstanev (approved):

PR #259, opened 2020-09-15
feat: upgrade node-lockfile-parser (uuid upgrade)

FauxFaux:
This snyk-nodejs-lockfile-parser upgrade uses the newer uuid library,
which we're already using. (uuid 8)

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "gunzip-maybe": "^1.4.2",
     "mkdirp": "^1.0.4",
     "semver": "^6.1.0",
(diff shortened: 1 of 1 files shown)


ivanstanev:
@FauxFaux just to double check, what's different in this newer version? 😄 Is this preparing some follow up PR?

FauxFaux:
It literally just changes the version of `uuid` used, and the import style you have to use to get the same behaviour.

https://github.com/snyk/nodejs-lockfile-parser/releases

I'm trying to get the deprecated uuid out of the tree; this is it.

ivanstanev (approved):

PR #260, opened 2020-09-17
feat: display the scan result formatted

agranado2k:
As part of the envelope project, now the plugin is responsible to format the scanned information to be displayed. It gives more flexibility for the plugins owners to format the information.

- [x] Ready for review

#### What does this PR do?
Create the first iteration for the `display()` function that formats the scanned results.

The `display()` function handles only the human readable output.

For the specific `display()` test, you can run `npx tap test/lib/display.test.ts -R=spec --timeout=300`

As part of the project envelope, the plugin got the responsibility to format the scanned results to be displayed. 

[RUN-1175](https://snyksec.atlassian.net/browse/RUN-1175)

#### Screenshots

Screenshot 2020-09-28 at 17 59 44

Screenshot 2020-09-28 at 18 00 03

Screenshot 2020-09-29 at 12 23 15

### DIFF
--- a/lib/display.ts
+++ b/lib/display.ts
@@ -1,0 +1,290 @@
+import chalk from "chalk";
+import * as os from "os";
+
+import { createFromJSON, DepGraph } from "@snyk/dep-graph";
+import {
+  BaseImageRemediationAdvice,
+  ContainerTarget,
+  Issue,
+  Options,
+  ScanResult,
+  TestResult,
+} from "./types";
+
+const BREAK_LINE = os.EOL;
+const SECTION_PADDING_TO_FORMAT_METADATA = 19;
+
+export async function display(
+  scanResults: ScanResult[],
+  testResults: TestResult[],
+  errors: string[],
+  options?: Options,
+): Promise<string> {
+  const result: string[] = [];
+  let index = 0;
+  for (const testResult of testResults) {
+    const formattedIssue: string[] = [];
+    for (const issue of testResult.issues) {
+      formattedIssue.push(formatIssue(testResult, issue));
+    }
+    result.push(formattedIssue.join(BREAK_LINE));
+    result.push(includeSectionSeparator());
+
+    const scanResult: ScanResult = scanResults[index];
+    const metadata = formatMetadataSection(scanResult, testResult);
+    result.push(metadata);
+    result.push(includeSectionSeparator());
+
+    const summary = formatSummary(testResult);
+    result.push(summary);
+    result.push(includeSectionSeparator());
+
+    const remediations = formatRemediations(testResult);
+    if (remediations) {
+      result.push(remediations);
+      result.push(includeSectionSeparator());
+    }
+
+    const suggestions = formatSuggestions(options);
+    if (suggestions) {
+      result.push(suggestions);
+      result.push(includeSectionSeparator());
+    }
+
+    const userCTA = formatUserCTA(options);
+    if (userCTA) {
+      result.push(userCTA);
+    }
+
+    index += 1;
+  }
+
+  return result.join(BREAK_LINE);
+}
+
+function includeSectionSeparator(): string {
+  return BREAK_LINE;
+}
+
+function formatIssue(testResult: TestResult, issue: Issue): string {
+  const result: string[] = [];
+  const issueData = testResult.issuesData[issue.issueId];
+  const severity = capitalize(issueData.severity);
+  const pkg = issue.pkgName;
+  const color = getColor(issueData.severity);
+
+  const header = color(`✗ ${severity} severity vulnerability found in ${pkg}`);
+  const description = `  Description: ${issueData.title}`;
+  const info = `  Info: https://snyk.io/vuln/${issue.issueId}`;
+  const introduced = `  Introduced through: ${formatIntroduced(
+    issueData.from,
+  )}`;
+  const from = formatFrom(issueData.from);
+  const fixedIn = formatFixedIn(issue);
+
+  result.push(header);
+  result.push(description);
+  result.push(info);
+  result.push(introduced);
+  result.push(from);
+  if (fixedIn) {
+    result.push(fixedIn);
+  }
+  result.push("");
+
+  return result.join(BREAK_LINE);
+}
+
+function capitalize(word: string): string {
+  return word[0].toUpperCase() + word.slice(1);
+}
+
+function formatIntroduced(fromList: string[][]) {
+  const result: string[] = [];
+
+  for (const from of fromList) {
+    result.push(from[0]);
+  }
+
+  return result.join(", ");
+}
+function formatFrom(fromList: string[][]): string {
+  const result: string[] = [];
+  let counter = 0;
+  const max = 3;
+  for (const localFrom of fromList) {
+    if (counter >= max) {
+      break;
+    }
+    counter += 1;
+
+    result.push(`  From: ${localFrom.join(" > ")}`);
+  }
+  if (fromList.length > max) {
+    result.push(`  and ${(fromList.length = max)} more...`);
+  }
+
+  return result.join(BREAK_LINE);
+}
+
+function formatFixedIn(issue: Issue): string | undefined {
+  if (!issue.fixInfo || !issue.fixInfo.nearestFixedInVersion) {
+    return undefined;
+  }
+
+  return chalk.bold.green(`  Fixed in: ${issue.fixInfo.nearestFixedInVersion}`);
+}
+
+function formatMetadataSection(
+  scanResult: ScanResult,
+  testResult: TestResult,
+): string {
+  const result: string[] = [];
+  result.push(formatMetadataLine("Organization:", testResult.org));
+
+  const packageManager = scanResult.identity.type;
+  result.push(formatMetadataLine("Package manager:", packageManager));
+
+  const target: ContainerTarget = scanResult.target as ContainerTarget;
+  const projectName = target.image;
+  const image = target.image.replace("docker-image|", "");
+  result.push(formatMetadataLine("Project name:", projectName));
+  result.push(formatMetadataLine("Docker image:", image));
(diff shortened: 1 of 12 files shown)


lib/display.ts
```diff
@@ -0,0 +1,175 @@
...
+function formatIssue(testResult: TestResult, issue: Issue): string {
+  const result: string[] = [];
+  const issueData = testResult.issuesData[issue.issueId];
+  const severity = capitalize(issueData.severity);
+  const pkg = issue.pkgName;
+  const color = getColor(issueData.severity);
+
+  const header = color(`✗ ${severity} severity vulnerability found in ${pkg}`);
+  const description = `  Description: ${issueData.title}`;
+  const info = `  Info: https://snyk.io/vuln/${issue.issueId}`;
```
ivanstanev, line 47: I wonder if the URL should be read from the one stored in `snyk config`, maybe not something to cover for now.

lib/display.ts
```diff
@@ -0,0 +1,175 @@
...
+  errors: string[],
+  options?: Options,
+): Promise<string> {
+  const testingPath = options?.path;
+
+  const result: string[] = [];
+  if (options && options.path) {
+    const prefix = chalk.bold.white("\nTesting " + testingPath + "...\n\n\n");
+    result.push(prefix);
+  }
```
ivanstanev, line 18: I think this line won't even have the chance to be seen by the user, the whole `display` function seems like one big synchronous block so I think maybe we don't need this line?

ivanstanev (commented):
Looks good! 

I wonder if there should be a new line between the different vulns and also between the last line "Tested xyz dependencies, found abc vulnerabilities"?

lib/display.ts
ivanstanev: I still think `const prefix = chalk.bold.white("\nTesting " + testingPath + "...\n\n\n");` won't get displayed to the user like for example when the spinner is waiting for the results, do we need this line?

lib/display.ts
```diff
@@ -0,0 +1,289 @@
...
+import {
+  BaseImageRemediationAdvice,
+  ContainerTarget,
+  Issue,
+  Options,
+  ScanResult,
+  TestResult,
+} from "./types";
+
+const BREAK_LINE = "\n";
```
ivanstanev, line 13: Can we make this `os.EOL` so that it also works on Windows?

lib/display.ts
```diff
@@ -0,0 +1,289 @@
...
+  const projectName = target.image;
+  const image = target.image.replace("docker-image|", "");
+  result.push(formatMetadataLine("Project name:", projectName));
+  result.push(formatMetadataLine("Docker image:", image));
+
+  return result.join(BREAK_LINE);
+}
+
+function formatMetadataLine(header: string, info: string = ""): string {
+  const padLength = 19;
```
ivanstanev, line 163: I understand this comes from the CLI but what is the reason behind it being 19?

lib/display.ts
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  return color;
+}
+
+export function formatRemediations(res: TestResult) {
+  if (!res.docker || !res.docker.baseImageRemediation) {
+    return "";
+  }
+  const { advice, message } = res.docker.baseImageRemediation;
```
ivanstanev, line 214: `/test-dependencies` does not return this yet, could we sync with @darscan and @gitphill if this is an OK format to keep? Just noticing that `docker` is a very ecosystem-specific field for a generic /test endpoint.

lib/display.ts
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  const dockerSuggestion: string[] = [];
+  if (options.config && options.config.disableSuggestions !== "true") {
+    const optOutSuggestions =
+      "To remove this message in the future, please run `snyk config set disableSuggestions=true`";
+    if (!options.file) {
+      dockerSuggestion.push(
+        chalk.bold.white(
+          "Pro tip: use `--file` option to get base image remediation advice.",
```
ivanstanev, line 252: Wondering if we will be removing this in the future due to the `rootFs` work automatically detecting the base image? CC @garethr

lib/display.ts
```diff
@@ -0,0 +1,289 @@
...
+        ),
+      );
+      dockerSuggestion.push(
+        chalk.bold.white(
+          `Example: $ snyk test --docker ${options.path} --file=path/to/Dockerfile`,
```
ivanstanev, line 257: This should now be `snyk container test`

lib/display.ts
```diff
@@ -0,0 +1,289 @@
...
+        ),
+      );
+      dockerSuggestion.push(BREAK_LINE);
+      dockerSuggestion.push(optOutSuggestions);
+    } else if (!options["exclude-base-image-vulns"]) {
+      dockerSuggestion.push(
+        chalk.bold.white(
+          "Pro tip: use `--exclude-base-image-vulns` to exclude from display Docker base image vulnerabilities.",
```
ivanstanev, line 265: I wonder if this should be `container` instead of `Docker` (also there was another message above similarly saying `Docker image:`) CC @garethr

test/fixtures/display/no-issues.txt
```diff
@@ -0,0 +1,18 @@
...
+[32mPackage manager:   [39m rpm
+[32mProject name:      [39m docker-image|snyk/kubernetes-monitor
+[32mDocker image:      [39m snyk/kubernetes-monitor
+
+
+[32m✓ Tested 90 dependencies for known issues, no vulnerable paths found.[39m
+
+
+[1m[37mPro tip: use `--file` option to get base image remediation advice.[39m[22m
+[1m[37mExample: $ snyk test --docker undefined --file=path/to/Dockerfile[39m[22m
```
ivanstanev, line 14: Undefined?

test/lib/display.test.ts
```diff
@@ -0,0 +1,254 @@
...
+      },
+    ];
+    const errors: string[] = [];
+    const options: Options = {
+      config: {},
+    } as Options;
+
+    const result = await display(scanResults, testResults, errors, options);
+
+    t.same(result, expectedDisplay, "the display is not as expected");
```
ivanstanev, line 48: I think the message should be positive, e.g. "the display is as expected".

test/lib/display.test.ts
```diff
@@ -0,0 +1,254 @@
...
+    async (t) => {
+      const expectedDisplay = await readFixture("display", "a-few-issues.txt");
+      const debDepGraphData: DepGraphData = JSON.parse(
+        await readFixture("display", "deb-dep-graph.json"),
+      );
+      const rpmImageScanResult: ScanResult = JSON.parse(
+        await readFixture("scan-results", "rpm.json"),
+      );
+      const scanResults: ScanResult[] = [rpmImageScanResult];
+      const testResults: TestResult[] = [
```
ivanstanev, line 97: Should we make these a fixture as well? Makes the test a bit easier to read!

test/lib/display.test.ts
```diff
@@ -0,0 +1,254 @@
...
+      "only-base-image-remediations.txt",
+    );
+    const debDepGraphData: DepGraphData = JSON.parse(
+      await readFixture("display", "deb-dep-graph.json"),
+    );
+    const rpmImageScanResult: ScanResult = JSON.parse(
+      await readFixture("scan-results", "rpm.json"),
+    );
+    const scanResults: ScanResult[] = [rpmImageScanResult];
+    const testResults: TestResult[] = [
```
ivanstanev, line 211: This could also be a fixture

ivanstanev (commented):
Well done! 👏

shaimendel:
Regarding the output - shouldn't it be exactly the same as we have today?
The missing parts:
1. another newline to separate different issues
2. the low severity color seems different
3. the colors of the last paragraph (Organization, Package manager etc) are different
4. did we remove the `Licenses` on purpose?
5. the `--file` tip is missing

image

lib/display.ts
agranado2k: I agree! But I just copied it from the way it's being done in the CLI. We can remove it. Wdyt?

lib/display.ts
agranado2k: I don't know. I'll try to give a better name, something more meaningful. But why it is 19 I don't know... :/

test/lib/display.test.ts
```diff
@@ -0,0 +1,254 @@
...
+    readFile(file, "utf-8", (err, data) => {
+      if (err) {
+        reject(err);
+      }
+      resolve(data);
+    });
+  });
+}
+
+test("display", async (c) => {
```
agatakrajewska, line 22: Ah, this is a fantastic candidate for jest snapshot testing that @ivanstanev added! I think if Ivan's PR has to go in first for any reason, it would be worth considering adding :-)

lib/display.ts
agranado2k: Yep. We can remove if we think it doesn't worth anymore has it.

test/fixtures/display/no-issues.txt
agranado2k: Ops... my bad!

test/lib/display.test.ts
agranado2k: Definitely!

lib/display.ts
agatakrajewska: Also, let's not forget tests for windows 😬

agranado2k: To test for Windows do I need to copy the test file to `test/windows` folder?

lib/display.ts
shaimendel: not sure we need to - what was the cli doing so far? we should support exactly what it used to

lib/display.ts
shaimendel: we need it anyway to show which line in the dockerfile caused which vuln

shaimendel (dismissed):

shaimendel (approved):

PR #261, opened 2020-09-21
feat: add scan() to replace inspect() and return ScanResult[]

ivanstanev:
- [x] Ready for review

#### What does this PR do?

BREAKING CHANGE: We are changing the format of Snyk CLI plugins to a new interface.

As part of this change we are also rolling out the following:
- enabling Distroless scanning by default
- enabling application dependencies scanning by default (currently Node.js)
- enabling static scanning by default

Additionally adds Jest system tests as part of the release pipeline.

`npm run test` and `npm run test-jest`

[Jira ticket RUN-1192](https://snyksec.atlassian.net/browse/RUN-1192)


lib/types.ts
```diff
@@ -34,52 +24,94 @@ export interface ManifestFile {
...
-  hashes: string[];
-}
+export type FactType =
+  | "depGraph"
+  | "keyBinariesHashes"
+  | "imageLayers"
+  | "dockerfileAnalysis"
+  | "rootFs"
+  | "imageId"
+  | "manifestFile";
```
ivanstanev, line 34: Added by mistake, will remove

lib/analyzer/index.ts
```diff
@@ -1,4 +1,3 @@
-import { analyze as analyzeDynamically } from "./docker-analyzer";
 import { analyze as analyzeStatically } from "./static-analyzer";
 
-export { analyzeDynamically, analyzeStatically };
+export { analyzeStatically };
```
agatakrajewska, line 3: Since we're getting rid of dynamic scanning it would be really great to do some renaming here, but it's just a note and I appreciate it could be done as a separate PR when we modularise/restructure `sdp` for example, like you mentioned before

lib/analyzer/applications/types.ts
```diff
@@ -1,3 +1,8 @@
+import { ScanResult } from "../../types";
+
+export interface AppDepsScanResultWithoutTarget
+  extends Omit<ScanResult, "target"> {}
```
agatakrajewska, line 4: Omg, that's cool, I've never seen this utility type before!

.circleci/config.yml
```diff
@@ -129,6 +129,24 @@ jobs:
           command: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" >> .npmrc
       - run: npm install
       - run: npm run test-windows
+  test_jest:
```
agatakrajewska, line 132: that's so great 👍

lib/analyzer/index.ts
ivanstanev: Agreed! But some tests call this directly and I didn't want to touch too many things while in the middle of refactoring the tests, I'll definitely note it down for a future clean up though!

test/system/app-os/globs.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+    const image = "debian:10";
+    const pluginResult = await scan({
+      path: image,
+      globsToFind: {
+        include: ["**/os-release"],
+        exclude: [],
+      },
+    });
+
+    expect(pluginResult.manifestFiles).toMatchSnapshot();
```
agatakrajewska, line 19: Fantastic :-)

lib/analyzer/index.ts
agatakrajewska: Makes total sense, I thought as much :-) 👍

lib/analyzer/static-analyzer.ts
```diff
@@ -47,20 +43,13 @@ import { ImageAnalysis, OSRelease, StaticAnalysis } from "./types";
...
-
 export async function analyze(
   targetImage: string,
   dockerfileAnalysis: DockerFileAnalysis | undefined,
-  options: StaticAnalysisOptions,
+  imageType: ImageType,
+  imagePath: string,
+  globsToFind: { include: string[]; exclude: string[] },
 ): Promise<StaticAnalysis> {
-  if (!supportedArchives.includes(options.imageType)) {
```
agatakrajewska, line 60: Have we agreed that this is not required any more? As far as I can remember that error was requested to throw if image type is unknown, just an FYI

jest.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  preset: "ts-jest",
+  testEnvironment: "node",
+  testMatch: ["<rootDir>/test/**/*.spec.ts", "!<rootDir>/test/windows/**"],
```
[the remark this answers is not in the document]
ivanstanev: Yes, it means "not files in the windows folder"

lib/analyzer/static-analyzer.ts
ivanstanev: This was only exercised in the static scanning if whoever called it with `staticScanningOptions` gave a bad image type. In reality this never happened in the CLI for example. 

We already have code up the stack that throws if ImageType is not the one we expect 🙏

lib/analyzer/applications/node.ts
```diff
@@ -32,12 +32,22 @@ export async function nodeFilesToScannedProjects(
     );
 
     const strippedLabelsParserResult = stripUndefinedLabels(parserResult);
+    const depGraph = await legacy.depTreeToGraph(
```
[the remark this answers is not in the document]
ivanstanev: I checked last release and I don't see an exported function for dep graphs

lib/experimental.ts
```diff
@@ -43,76 +58,55 @@ async function localArchive(
...
+
   // The target image becomes the base of the path, e.g. "archive.tar" for "/var/tmp/archive.tar"
   const imageIdentifier = path.basename(archivePath);
-  return await getStaticAnalysisResult(
+  return await staticModule.analyzeStatically(
     imageIdentifier,
-    archivePath,
     dockerfileAnalysis,
     imageType,
-    options["app-vulns"],
```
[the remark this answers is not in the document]
ivanstanev: Need to check with 🌋 actually!

lib/experimental.ts
```diff
@@ -43,76 +58,55 @@ async function localArchive(
...
-    );
-  }
+  const excludeBaseImageVulns = options["exclude-base-image-vulns"] || false;
+  const globToFind = {
+    include: options.globsToFind?.include || [],
+    exclude: options.globsToFind?.exclude || [],
+  };
 
-  const imageSavePath = fullImageSavePath(options?.imageSavePath);
+  const imageSavePath = fullImageSavePath(options["image-save-path"]);
```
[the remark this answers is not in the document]
ivanstanev: So that it can be controlled for example from the CLI, can be a useful flag for CI/CD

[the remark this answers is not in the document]
ivanstanev: OK discussed, it would be a breaking change, so reverting to `imageSavePath`

lib/index.ts
```diff
@@ -1,172 +1,38 @@
...
+    throw new Error("No plugin options provided");
   }
 
-  if (staticUtil.isRequestingStaticAnalysis(options)) {
-    return await staticUtil.analyzeStatically(
-      targetImage,
-      dockerfileAnalysis,
-      options,
-    );
+  const targetImage = options.path;
```
shaimendel, line 27: what is this `path` parameter? it's the first time I see it

lib/index.ts
```diff
@@ -1,172 +1,38 @@
...
+    dockerfilePath,
   );
-}
-
-// TODO: return type should be "DynamicAnalysisOptions" or something that extends DockerOptions
-function getDynamicAnalysisOptions(options?: any): any {
-  return options
-    ? {
-        host: options.host,
-        tlsverify: options.tlsverify,
```
shaimendel, line 91: what was the meaning of the tls flags?

ivanstanev: According to the [docs](https://docs.docker.com/engine/security/https/) it's about validating to a remote Docker daemon. I don't think it's used by 🌋 any more and it doesn't look like it's used by the CLI either

lib/index.ts
ivanstanev: This used to be `root` - the first parameter passed to the plugin's `inspect`. It was always referred to as `path` in the CLI, I think it is because of the assumption the CLI is scanning some filesystem path.

shaimendel (approved):

PR #264, opened 2020-09-25
feat: [RUN-1197] Fail if Docker not experimental

agatakrajewska:
- [x] Ready for review

#### What does this PR do?
We support the '--platform' flag only when Docker experimental features enabled, throwing a user friendly error otherwise.

https://snyksec.atlassian.net/browse/RUN-1197

### DIFF
--- a/lib/docker.ts
+++ b/lib/docker.ts
@@ -137,12 +137,24 @@ class Docker {
     return await dockerPull.pull(registry, repo, tag, opt);
   }
 
+  public async isDockerExperimentalEnabled(): Promise<boolean> {
+    const output = await subProcess.execute("docker", ["system", "info"]);
+    return output.stdout.includes("Experimental: true");
+  }
+
   public async pullCli(targetImage: string, options?: DockerOptions) {
-    return subProcess.execute("docker", [
-      "pull",
-      options?.platform ? `--platform=${options.platform}` : "",
-      targetImage,
-    ]);
+    let platformOption: string = "";
+    if (options?.platform) {
+      if (await this.isDockerExperimentalEnabled()) {
+        platformOption = `--platform=${options.platform}`;
+      } else {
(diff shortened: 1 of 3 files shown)


lib/docker.ts
```diff
@@ -137,12 +137,24 @@ class Docker {
     return await dockerPull.pull(registry, repo, tag, opt);
   }
 
+  public async isDockerExperimentalEnabled(): Promise<boolean> {
+    const output = await subProcess.execute("docker system info");
```
ivanstanev, line 141: Interesting! Isn't every argument its own element of a list, for example
```typescript
await subProcess.execute("docker", ["system", "info"]);
```

test/lib/docker.test.ts
```diff
@@ -533,11 +533,25 @@ test("pullCli", async (t) => {
...
       "args passed to subProcess.execute as expected",
     );
   });
+
+  t.test(
+    "with platform arg, when no experimental features enabled",
+    async (t) => {
+      stub.resolves({ stdout: "Experimental: false" });
+
+      t.rejects(
```
ivanstanev, line 549: I think we should `await t.rejects(` since it returns a Promise

ivanstanev (commented):
Looks good!

I think the release notes could be changed, we have a reference to a Jira ticket that users cannot see or access, it's also something internal to our team so it won't make sense to them. WDYT of `feat: fail if --platform cannot be used` or something similar?

lib/docker.ts
agatakrajewska: Ah you're right, I think that is the way the arguments should be passed!

test/lib/docker.test.ts
agatakrajewska: 100%, thanks!

ivanstanev (approved):

PR #274, opened 2020-10-15
Chore/docs

ivanstanev:
- [x] Ready for review

#### What does this PR do?

List supported functionality provided by the plugin. Makes it easy to reference what is supported and what not!

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,16 +11,6 @@ Snyk helps you find, fix and monitor for known vulnerabilities in your dependenc
 
 This plugin provides dependency metadata for Docker images.
 
-## Running Tests
+## Tests
 
-To run tests the following environment variables need to be set:
-
-`DOCKER_HUB_PRIVATE_IMAGE`
-`DOCKER_HUB_USERNAME`
-`DOCKER_HUB_PASSWORD`
-
-`DOCKER_HUB_PRIVATE_IMAGE` should refer to an image that is hosted on Docker Hub but not available publicly. During CI test this is set to `snykgoof/dockergoof:alpine`.
-
-## Writing Tests
-
-Refer to [test/README.md](test/README.md).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -11,16 +11,39 @@ Snyk helps you find, fix and monitor for known vulnerabilities in your dependenc
...
-`DOCKER_HUB_PRIVATE_IMAGE`
-`DOCKER_HUB_USERNAME`
-`DOCKER_HUB_PASSWORD`
+- rpm, apk, deb
 
-`DOCKER_HUB_PRIVATE_IMAGE` should refer to an image that is hosted on Docker Hub but not available publicly. During CI test this is set to `snykgoof/dockergoof:alpine`.
+Operating systems:
 
-## Writing Tests
+- Debian, Red Hat, Alpine, Oracle, CentOS, vanilla Linux, SLES
```
shaimendel, line 22: A few things I'm worried about here
1. we will never remember to update it 😆 once we add new OSs
2. there are maybe others that we support? like Amazon Linux for example
3. there's a different between what the SDP supports and what Snyk supports. so for example the plugin also supports OpenSuse but Snyk does not

agranado2k (commented):
Have you bump the `rpm-parser`? It's complaining about the `RpmParserResponse`.
Screenshot 2020-10-15 at 11 51 51

README.md
ivanstanev: It's on us to keep the documentation up to date 🙂  similarly to how we keep our code and tests, they should be an inseparable part of our work. And also up until now we didn't have any reference as to what we support - it's either in your head or you had to look up the code to figure it out.

For the others we support I will add them to the list, I think this is related to what the _plugin_ supports and not what Snyk does.

shaimendel (dismissed):

shaimendel (approved):

PR #278, opened 2020-10-16
feat: add flag imageNameAndTag to preserve this data in archive scans

ivanstanev:
- [x] Ready for review

#### What does this PR do?

Override the "name" and "version" fields of the OS dependencies result when scanning image archives.

This is NOT used by the Snyk CLI!
It is used by K8s-Monitor and DRA to preserve the image identifier when scanning archives.
The archives do not contain any data about the image's name and tag (since they are only known by the container registry) and in some contexts we know this data and want to keep it.

[Jira ticket RUN-1220](https://snyksec.atlassian.net/browse/RUN-1220)

### DIFF
--- a/test/system/flags/image-name-and-tag.spec.ts
+++ b/test/system/flags/image-name-and-tag.spec.ts
@@ -1,0 +1,60 @@
+import { DepGraph } from "@snyk/dep-graph";
+import { join as pathJoin } from "path";
+import { scan } from "../../../lib/index";
+
+function getFixture(fixturePath: string): string {
+  return pathJoin(__dirname, "../../fixtures", fixturePath);
+}
+
+describe("imageNameAndTag tests", () => {
+  it("it overrides name and version when reading a docker archive", async () => {
+    const fixturePath = getFixture("docker-archives/docker-save/nginx.tar");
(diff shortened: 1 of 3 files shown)


test/system/flags/image-name-and-tag.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    });
+
+    const depGraph: DepGraph = pluginResult.scanResults[0].facts.find(
+      (fact) => fact.type === "depGraph",
+    )!.data;
+    expect(depGraph.rootPkg.name).toEqual("docker-image|nginx");
+    expect(depGraph.rootPkg.version).toEqual("1.23.4");
+  });
+
+  it("it overrides name and version when reading an oci archive", async () => {
```
agranado2k, line 27: Do we think that we need one more case where we have the archive `path` but not the `imageNameAndTag` option?

ivanstanev: We have this already in regular archives scanning tests (both Docker and OCI) 🙏

agranado2k (approved):
Good job! LGTM! 🎉

PR #282, opened 2020-10-19
feat: manifest files contents are now returned as a ScanResult Fact

ivanstanev:
- [x] Ready for review

#### What does this PR do?

Having the manifest files returned at the top-level plugin response does not conform with other Snyk CLI plugins.
The manifest files have only two use-cases: kubernetes-monitor and DRA.

We prefer to "hide" the manifest files as a Fact in the OS dependencies ScanResult (the 1st ScanResult).
These manifest files need further processing so we can extract actual package manager information and to create projects in Snyk.

Additionally, as part of this change, change the interface of the manifest files to return their contents as a Base64 string.
This is needed for two reasons:
- the existing Buffer interface cannot cleanly be transferred over the network (considering we use JSON in our APIs)
- we cannot make assumptions about the file encoding otherwise we can accidentally corrupt the data we receive

[Discussion on Slack](https://snyk.slack.com/archives/CBXS33GJY/p1589787181070700?thread_ts=1589385083.016900&cid=CBXS33GJY)

[Jira ticket RUN-1219](https://snyksec.atlassian.net/browse/RUN-1219)

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -23,7 +23,12 @@ export enum OsReleaseFilePath {
 export interface ManifestFile {
   name: string;
   path: string;
-  contents: Buffer;
+  /**
+   * Base64-encoded file contents.
+   * We use Base64 to avoid any assumptions about the original file encoding,
+   * which is difficult to infer and may be corrupted when the data is transferred over the network.
+   */
+  contents: string;
 }
 
 export type FactType =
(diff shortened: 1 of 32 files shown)


lib/types.ts
```diff
@@ -33,20 +38,12 @@ export type FactType =
   | "dockerfileAnalysis"
   | "rootFs"
   | "imageId"
-  | "imageOsReleasePrettyName";
+  | "imageOsReleasePrettyName"
+  | "manifestFiles";
```
shaimendel, line 42: do we want to change to a more container specific name? just in case someone else in the future will want to add manifest file as a fact

shaimendel (dismissed):

shaimendel (dismissed):

shaimendel (approved):

PR #287, opened 2020-10-23
fix: avoid race condition to remove pulled image and tmp folder

agranado2k:
- [x] Ready for review

#### What does this PR do?
On some platforms like Windows we're getting race conditions when we synchrony remove folders. To avoid that we're removing it in async mode.

The problem seems to happen only in the CI when running tests for Windows platform [here](https://app.circleci.com/pipelines/github/snyk/snyk/2820/workflows/8dc49982-9d50-4c66-ab4a-245a94cd6783/jobs/15580).

[Slack thread](https://snyk.slack.com/archives/CDSMEJ29E/p1603218013373900)

### DIFF
--- a/test/lib/analyzer/image-inspector.test.ts
+++ b/test/lib/analyzer/image-inspector.test.ts
@@ -141,10 +141,6 @@ test("get image as an archive", async (t) => {
       fs.existsSync(path.join(imageSavePath, "image.tar")),
       "image should not exists on disk",
     );
-    t.notOk(
-      fs.existsSync(imageSavePath),
-      "tmp folder should not exist on disk",
-    );
     t.ok(fs.existsSync(customPath), "custom path should exist on disk");
   });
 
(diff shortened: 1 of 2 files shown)


test/lib/analyzer/image-inspector.test.ts
```diff
@@ -141,10 +141,6 @@ test("get image as an archive", async (t) => {
       fs.existsSync(path.join(imageSavePath, "image.tar")),
       "image should not exists on disk",
     );
-    t.notOk(
```
agranado2k, line 144: I need to remove because now the folder is remove asynchrony. So we can't guarantee that it has been already removed when we check it.

ivanstanev: I think there is a way to test this with Jest/sinon by doing the following:
1. In the code where you call `fs.rmdir(imageFolderPath, (err) => {` you pull out the anonymous  function `(err) => {}` and put it in a separate file and export it -- this makes it possible that we can mock it with Jest/sinon and detect when it is being called.
2. Modify the tests here to mock the exported function and detect when it is being called. Once it's called, we can check the folder really does not exist on disk.

ivanstanev (approved):

test/lib/analyzer/image-inspector.test.ts
agranado2k: Ok! I'm going to do it later! :)

PR #289, opened 2020-11-02
[RUN-1233] Feat/detect jar hashes

agatakrajewska:
- [x] Ready for review

#### What does this PR do?
We're adding jar scanning support. Each jar binary is hashed with sha1 algorithm, aggregated on path and returned as scanResult containing 'jarFingerprints' fact.

Commit by commit :-) 

https://snyksec.atlassian.net/browse/RUN-1233


lib/analyzer/applications/java.ts
```diff
@@ -0,0 +1,56 @@
...
+import { AppDepsScanResultWithoutTarget, FilePathToContent } from "./types";
+
+function mapContentToJarFingerprints(input: {
+  [fileName: string]: string;
+}): {
+  [path: string]: JarFingerprint[];
+} {
+  const jarFingerprints: JarFingerprint[] = Object.entries(input).map(
+    ([key, value]) => {
+      const path = key.substring(0, key.lastIndexOf("/"));
```
ivanstanev, line 11: I think it's better to handle this with the `path` module as it can deal with `/` vs `\\` in paths

agatakrajewska: ah great point, thanks!!

lib/stream-utils.ts
```diff
@@ -54,6 +57,14 @@ export async function streamToHash(stream: Readable): Promise<string> {
   });
 }
 
+export async function streamToHash256(stream: Readable): Promise<string> {
```
shaimendel, line 60: both functions should be called with `ShaXXX` and not `HashXXX`

lib/analyzer/static-analyzer.ts
```diff
@@ -130,11 +135,22 @@ export async function analyze(
   const binaries = getBinariesHashes(extractedLayers);
 
   const applicationDependenciesScanResults: AppDepsScanResultWithoutTarget[] = [];
+
   if (appScan) {
     const nodeDependenciesScanResults = await nodeFilesToScannedProjects(
-      getNodeAppFileContent(extractedLayers),
+      getAppFileContent(
```
shaimendel, line 141: I guess we should move the `app` part of the function name?

lib/analyzer/applications/java.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export async function jarFilesToScannedProjects(
+  filePathToContent: FilePathToContent,
+): Promise<AppDepsScanResultWithoutTarget[]> {
+  const mappedResult = mapContentToJarFingerprints(filePathToContent);
+  const scanResults: AppDepsScanResultWithoutTarget[] = [];
+
+  for (const key in mappedResult) {
+    if (Object.prototype.hasOwnProperty.call(mappedResult, key)) {
```
shaimendel, line 38: 1. why do you need this `if`? is if because of TS?
2. maybe do the opposite?
```JS
if (!Object.prototype.hasOwnProperty.call(mappedResult, key)) {
    continue;
}
rest of code
```

lib/analyzer/applications/java.ts
```diff
@@ -0,0 +1,56 @@
...
+      const element = mappedResult[key];
+      scanResults.push({
+        facts: [
+          {
+            type: "jarFingerprints",
+            data: element,
+          },
+        ],
+        identity: {
+          type: "maven",
```
shaimendel, line 48: where is this used?

shaimendel (commented):
amazing PR!!

lib/analyzer/static-analyzer.ts
shaimendel: or maybe not? I really don't know what's better, maybe leave it 😆

lib/analyzer/applications/java.ts
agatakrajewska: Yeah, ts! I can do that, less arrow! 👍

lib/analyzer/applications/java.ts
agatakrajewska: I think this is then used in the registry to identify the package manager, but I'm not sure, will confirm with @ivanstanev tomorrow

jest.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   preset: "ts-jest",
   testEnvironment: "node",
-  testMatch: ["<rootDir>/test/system/**/*.spec.ts"],
+  testMatch: ["<rootDir>/test/**/*.spec.ts"],
```
ivanstanev, line 4: This makes also the `windows` tests run unfortunately! There might be a way to exclude the windows path here as a new entry, something like `"!<rootDir>/test/windows/**/*"`

agatakrajewska: Yeah, exactly, I'm in progress of fixing it 😄 🙏

lib/analyzer/applications/java.ts
```diff
@@ -0,0 +1,58 @@
...
+    ([filePath, digest]) => {
+      return {
+        location: filePath,
+        digest,
+      };
+    },
+  );
+
+  const resultAggregatedByPath = jarFingerprints.reduce(
+    (accumulator, current) => {
```
ivanstanev, line 20: nit: these could have more descriptive names like `(jarsAggregatedByPath, jarFingerprint)`?

lib/analyzer/applications/java.ts
```diff
@@ -0,0 +1,58 @@
...
+      const location = path.dirname(current.location);
+      accumulator[location] = accumulator[location] || [];
```
ivanstanev, line 22: nit: could write to `accumulator[location]` only the first time, when the array is not initialized and needs to be created

lib/analyzer/applications/java.ts
```diff
@@ -0,0 +1,58 @@
...
+  const jarFingerprints: JarFingerprint[] = Object.entries(input).map(
+    ([filePath, digest]) => {
+      return {
+        location: filePath,
+        digest,
+      };
+    },
+  );
+
+  const resultAggregatedByPath = jarFingerprints.reduce(
```
ivanstanev, line 19: nit: for readability this could be a for-loop too! Since you're not necessarily reducing the array to a single value but mainly restructuring the data

lib/analyzer/applications/java.ts
```diff
@@ -0,0 +1,58 @@
+import * as path from "path";
+import { JarFingerprint } from "../types";
+import { AppDepsScanResultWithoutTarget, FilePathToContent } from "./types";
+
+function mapContentToJarFingerprints(input: {
```
ivanstanev, line 5: nit: "mapping the content" doesn't convey enough information about what the function is doing so it means the abstraction is gone - you have to go into the code to understand what it is doing. WDYT of `groupJarFingerprintsByPath()` or something like that?

lib/analyzer/applications/java.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export async function jarFilesToScannedProjects(
+  filePathToContent: FilePathToContent,
+): Promise<AppDepsScanResultWithoutTarget[]> {
+  const mappedResult = mapContentToJarFingerprints(filePathToContent);
+  const scanResults: AppDepsScanResultWithoutTarget[] = [];
+
+  for (const key in mappedResult) {
+    if (!Object.prototype.hasOwnProperty.call(mappedResult, key)) {
```
ivanstanev, line 39: It seems odd that we have to call this at all. Maybe the `for loop` is giving us the wrong data? Maybe instead of `for ... in ...` we need `for ... of ...`? Or we could use something like `Object.entries()`?

lib/analyzer/applications/java.ts
```diff
@@ -0,0 +1,58 @@
...
+      accumulator[location].push(current);
+      return accumulator;
+    },
+    {},
+  );
+
+  return resultAggregatedByPath;
+}
+
+export async function jarFilesToScannedProjects(
```
ivanstanev, line 32: nit: Hehe, just noticed that `scanned projects` used to be the old name for `ScanResult`. Just something that might want to rename in the future for consistency

lib/analyzer/static-analyzer.ts
```diff
@@ -73,6 +72,12 @@ export async function analyze(
     );
   }
 
+  if (appScan) {
+    staticAnalysisActions.push(
+      ...[getNodeAppFileContentAction, getJarFileContentAction],
+    );
```
ivanstanev, line 78: nit: (if it's possible 😄)
```diff
    staticAnalysisActions.push(
-      ...[getNodeAppFileContentAction, getJarFileContentAction],
+      getNodeAppFileContentAction,
+      getJarFileContentAction,
    );
```

lib/index.ts
```diff
@@ -21,3 +21,19 @@ export {
...
+        "docker-archive:/Users/agatakrajewska/Source/snyk-docker-plugin/test/fixtures/docker-archives/docker-save/java.tar",
+      "app-vulns": true,
+    });
+    // tslint:disable-next-line: no-console
+    console.log(result);
+  } catch (error) {
+    // tslint:disable-next-line: no-console
+    console.log(error);
+  }
+});
```
ivanstanev, line 39: Accidentally checked in? 🙏

lib/stream-utils.ts
```diff
@@ -54,6 +57,14 @@ export async function streamToHash(stream: Readable): Promise<string> {
...
 }
 
+export async function streamToSha256(stream: Readable): Promise<string> {
+  return streamToHash(stream, HashAlgorithm.Sha256);
+}
+
+export async function streamToSha1(stream: Readable): Promise<string> {
+  return streamToHash(stream, HashAlgorithm.Sha1);
+}
+
```
ivanstanev, line 67: Amazing 👏

lib/types.ts
```diff
@@ -43,7 +48,9 @@ export type FactType =
   // Package manager manifests (e.g. requirements.txt, Gemfile.lock) collected as part of an application scan.
   | "imageManifestFiles"
   // Used for application dependencies scanning; shows which files were used in the analysis of the dependencies.
-  | "testedFiles";
+  | "testedFiles"
+  // Hashes of extracted *.jar binaries, hashed with sha1 algorithm
```
ivanstanev, line 52: Love the comment 🙏

package.json
```diff
@@ -20,7 +20,7 @@
     "test-jest": "jest --bail --logHeapUsage",
     "test-windows": "npm run lint && tap test/windows/**/*.test.ts -R=spec --timeout=300",
     "test-jest-windows": "jest --bail --config test/windows/jest.config.js --logHeapUsage",
-    "unit-test": "tap test/system/**/*.test.ts test/lib -R=spec --timeout=300",
+    "unit-test": "tap test/system/**/*.test.ts test/lib/**/*.test.ts -R=spec --timeout=300",
```
ivanstanev, line 23: This seems odd actually, unit tests are also running the "system" tests? Perhaps we should only be testing only things from `test/lib` like you added? Not sure why we had until now unit tests doing the system tests 😄

test/lib/analyzer/java.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+describe("jarFilesToScannedProjects function", () => {
+  it("should return expected scannedProject[] result", async () => {
+    const filePathToContent = {
+      "/libs/another_dir/test.jar": "485de3a253e23f645037828c07f1d7f1af40763a",
+    };
+
+    const result = await jarFilesToScannedProjects(filePathToContent);
+    expect(result[0].facts[0].type).toEqual("jarFingerprints");
+    expect(result[0].identity.type).toEqual("maven");
+    expect(result[0].identity.targetFile).toEqual("/libs/another_dir");
```
ivanstanev, line 12: Should we add an expect() about the facts?

test/system/app-os/java.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { join as pathJoin } from "path";
```
ivanstanev, line 1: I know it's a bit confusing but this test should probably be under `application-scans` not `app-os` 🙏  `app-os` is about the manifestGlobs.
Actually it's probably a better idea to merge both `app-os` and `application-scans` into only a single directory `application-scans` so it's just less confusing!

ivanstanev (changes requested):
Amazing work, looks super tidy! 🎉

lib/analyzer/applications/java.ts
agatakrajewska: yes, changed 👍

package.json
agatakrajewska: You're totally right! Let me change that whiles we're at it :-)

test/system/app-os/java.spec.ts
agatakrajewska: So I thought that was the purpose of this dir: "app-os" being "app&os". I can change the name if it's confusing.

test/lib/analyzer/java.spec.ts
agatakrajewska: Why not!

test/system/app-os/java.spec.ts
agatakrajewska: Ah ok, I can see it now!! I haven't noticed that there was already a dir for app scans!!! I'll merge them 👍

shaimendel (approved):

PR #292, opened 2020-11-12
test: show a bug exists with injected base image in Dockerfiles

ivanstanev:
- [x] Ready for review

#### What does this PR do?

If the base image name and tag are injected using ARG in the Dockerfile, the analysis will incorrectly mark the baseImage as null:null.

[ZenDesk ticket 7256](https://snyk.zendesk.com/agent/tickets/7256)

### DIFF
--- a/test/system/bugs/Dockerfile.injected
+++ b/test/system/bugs/Dockerfile.injected
@@ -1,0 +1,6 @@
+ARG FROM
+ARG TAG
+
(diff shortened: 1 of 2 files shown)


shaimendel (dismissed):
amazing! might be easier to use the tar fixtures instead of pulling images in this case, for you consideration

shaimendel (approved):

PR #295, opened 2020-11-13
fix: allow base image to be missing if it cannot be read correctly

ivanstanev:
- [x] Ready for review

#### What does this PR do?

In some cases where the base image in the Dockerfile's FROM instruction contains arguments that are substituted at build-time, we would incorrectly detect the base image as "null:null" which is a bad user experience. We fix this by ensuring that we report the base image only if we can read every variable in the FROM instruction.

[ZenDesk #7256](https://snyk.zendesk.com/agent/tickets/7256)


test/lib/dockerfile/index.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+      "FROM image:tag",
+      "FROM image:tag AS foo",
+      "FROM image@sha256:abcd",
+      "FROM image@sha256:abcd AS foo",
+    ];
+    for (const dockerfile of dockerfiles) {
+      expect(
+        getDockerfileBaseImageName(DockerfileParser.parse(dockerfile)),
+      ).toBeDefined();
+    }
```
michael-c-snyk, line 34: Did we also think about using [it.each()](https://jestjs.io/docs/en/api#testeachtablename-fn-timeout) here, so the test report shows a separate test name for each individual dockerfile? I believe this helps with parallelisation and the "running failures first" stuff Jest does to speed up the feedback cycle too. Not sure how important those considerations are here though, esp since the for..of loop is arguably easier to read than the it.each() syntax 🤔

agranado2k (dismissed):
LGTM! 🎉

agatakrajewska (dismissed):

agatakrajewska (approved):

PR #296, opened 2020-11-13
refactor: Move a tap test to jest

michael-c-snyk:
Refactoring ahead of changes needed for RUN-1041

- [x ] Ready for review
- [x ] Follows CONTRIBUTING rules

#### What does this PR do?
Replaces image-inspector.ts's `Tap` test with equivalent functionality in `Jest`

This might benefit from looking at in a diff tool that _isn't_ GH's. Intellij's for example shows a clearer side-by-side.

This is in the spirit of "you touch it, you jestify it" for tap tests, as RUN-1041 is likely to introduce changes to this file.

RUN-1041


test/lib/analyzer/image-inspector.spec.ts
```diff
@@ -0,0 +1,265 @@
...
+  if (customPath.length < 2) {
+    return;
+  }
+
+  fs.rmdirSync(path.join(...customPath));
+  const next = customPath.slice(0, customPath.length - 1);
+  rmdirRecursive(next);
+}
+
+// prettier-ignore
```
ivanstanev, line 21: Why so? 😄

ivanstanev (approved):
Nice one! Don't forget to assign the PR to yourself (it's just something we do, it means you will make sure it gets merged, tests pass, etc.)

test/lib/analyzer/image-inspector.spec.ts
michael-c-snyk: Prettier would break the table formatting on the "it.each" block, making it hard to read :) I learnt that this comment only makes Prettier ignores the block immediately after it, so Prettier starts up again after this `describe`.

PR #299, opened 2020-11-17
initial commit

dkontorovskyy:
- [ ] Ready for review

### DIFF
--- a/lib/inputs/go/index.ts
+++ b/lib/inputs/go/index.ts
@@ -1,0 +1,104 @@
+import * as elf from "elfy";
+import * as path from "path";
+import { Readable } from "stream";
+import { ExtractAction } from "../../extractor/types";
+
+const ignoredPaths = [
+  path.normalize("/boot"),
+  path.normalize("/dev"),
+  path.normalize("/etc"),
+  path.normalize("/home"),
+  path.normalize("/media"),
+  path.normalize("/mnt"),
+  path.normalize("/proc"),
+  path.normalize("/root"),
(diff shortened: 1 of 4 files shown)


lib/inputs/go/index.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        return resolve();
+      }
+    });
+
+    stream.on('error', (error) => reject(error));
+
+    stream.once('data', (chunk) => {
+      const elfHeaderMagic = '\x7FELF';
+      const first16Bytes = chunk.toString('binary', 0, 4);
```
agatakrajewska, line 68: we can say first 4 now :-D

lib/inputs/go/index.ts
```diff
@@ -0,0 +1,84 @@
...
+
+      if (first16Bytes !== elfHeaderMagic) {
+        stream.destroy();
+      }
+    });
+
+    stream.on('data', (chunk) => {
+      // It actually exists from Node.js v8.0.0
+      if ((stream as any).destroyed) {
```
agatakrajewska, line 77: Awesome! Do we need to support earlier versions of Node? I suppose 8 is quite ancient history

dkontorovskyy: Because it's gonna be used in CLI, we need support from Node v8 and up. When we drop Node 8 support in CLI, we might update plugins.

agatakrajewska:
We've completed the work on another branch, which was named correctly hehe

PR #300, opened 2020-11-17
refactor: Correct the scope of non-unit unit test

michael-c-snyk:
- [x] Ready for review

This PR means we are longer testing either the snyk-docker-pull module OR Dockerhub itself in a unit test

This brings the following advantages:

- Lower chance of false negatives / flaky result stemming from external dependencies. The scope of the unit test is more appropriately limited to the module under test.
- Because there is no network call, the unit test suite runs faster, allowing for a tighter feedback loop
- This simplifies execution of the _unit_ test suite by removing the need for the env vars described in the readme.
- This in turn reduces chance of confusion when the SNYK_REGISTRY_USERNAME / SNYK_REGISTRY_PASSWORD env vars are added in RUN-1041


test/lib/analyzer/image-inspector.spec.ts
```diff
@@ -238,28 +231,17 @@ describe("getImageArchive", () => {
         password,
       );
 
-      expect(dockerPullSpy).toHaveBeenCalled();
-      expect(archiveLocation.path).toEqual(
-        path.join(imageSavePath, "image.tar"),
-      );
-
-      const imageExistsOnDisk: boolean = fs.existsSync(
```
ivanstanev, line 246: But does it mean we no longer create archives on disk? Do we not need to test that? Also do we not need to test that the archive was properly removed after calling `archiveLocation.removeArchive();`?

ivanstanev (commented):
Great description! 👏

test/lib/analyzer/image-inspector.spec.ts
michael-c-snyk: Good shout on removeArchive, I missed that this was set as a callback if the function under test works properly. I think I need to update the test so that the stub implementation creates a dummy file and then add removeArchive() back in, then assert the file gets removed as before.

michael-c-snyk: @ivanstanev  updated :)

test/lib/analyzer/image-inspector.spec.ts
```diff
@@ -254,12 +265,6 @@ describe("getImageArchive", () => {
         path.join(imageSavePath, "image.tar"),
       );
       expect(imageExistsOnDiskAfterDelete).toBe(false);
-
-      const tmpFolderExistsOnDisk: boolean = fs.existsSync(imageSavePath);
-      expect(tmpFolderExistsOnDisk).toBe(false);
-
-      const customPathExistsOnDisk: boolean = fs.existsSync(customPath);
-      expect(customPathExistsOnDisk).toBe(true);
```
ivanstanev, line 262: Do we not need these assertions any more? Are we not interested in knowing if we leave any directories or similar behind us after `getImageArchive()` finishes?

ivanstanev (approved):

PR #303, opened 2020-11-18
chore: stop lint running multiple times

michael-c-snyk:
As https://snyk.slack.com/archives/CDSMEJ29E/p1605708656050800, piping lint failures into tests outcomes can misleadingly mark test jobs in CI as failed when in fact only lint has failed, and tests are fine. This prevents that behaviour

- [x] Ready for review


shaimendel (commented):
Do we want this change? I thought we can just change circleci config to run lint as a prerequisite to those tests instead

PR #304, opened 2020-11-18
chore: Dont run tests in CI if lint fails

michael-c-snyk:
- [x] Ready for review

Fail fast by requiring lint to have succeeded before running tests.

As https://snyk.slack.com/archives/CDSMEJ29E/p1605708656050800, coupling lint failures to tests outcomes can misleadingly mark test jobs in CI as failed when in fact only lint has failed, and tests are fine. This prevents that behaviour

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -184,25 +184,29 @@ workflows:
       - build:
           name: Build
           requires:
-            - Install
+            - Lint
       - test:
           name: Test
           context: nodejs-install
           requires:
-            - Install
+            - Lint
       - test_jest:
           name: Test Jest
           context: nodejs-install
           requires:
-            - Install
+            - Lint
       - test_windows:
           name: Test Windows
           context: nodejs-install
           node_version: "8.17.0"
+          requires:
+            - Lint
       - test_jest_windows:
           name: Test Jest Windows
           context: nodejs-install
           node_version: "10.19.0"
+          requires:
+            - Lint
       - release:
           name: Release to GitHub
           context: nodejs-lib-release


.circleci/config.yml
```diff
@@ -184,25 +184,29 @@ workflows:
...
           name: Build
           requires:
-            - Install
+            - Lint
       - test:
           name: Test
           context: nodejs-install
           requires:
-            - Install
+            - Lint
```
ivanstanev, line 192: This should probably be
```yaml
- Lint
- Build
```
I think we can't run tests before Build happened! (I think we mount the built files from earlier)

ivanstanev: Not sure here though! Ignore me if it doesn't make sense haha

michael-c-snyk: Build and test have the same relationship after this PR as the old way (pictured) though don't they? 
Screenshot 2020-11-18 at 17 13 25

michael-c-snyk: Or are you saying that the old way was wrong too? Sorry if Im being slow - long day with lots of CI 😅

michael-c-snyk: @ivanstanev I think this conversation got lost, can I bump it when you have a mo please?

ivanstanev: Test Jest right now (because it requires Node 10) does its own Install and doesn't need to wait for any other steps (similarly the Windows tests don't wait for anything because they cannot use the modules built in the Install step)! So I would guess the pipeline should look like this:

```
Test Windows
Test Jest Windows
Test Jest
Install -> Lint -> Test
                   Build
```

michael-c-snyk: Ah I see! Since test-jest does npm run lint anyway, I think it's probably not worth the effort changing this then, at least until we complete move to jest

PR #305, opened 2020-11-18
Use ENV for credentials for downloading images on CLI

michael-c-snyk:
- [x] Ready for review

[RUN-1041] Setting SNYK_REGISTRY_USERNAME and/or SNYK_REGISTRY_PASSWORD environment variables allows these credentials to be used to log in to remote registries. The existing --username and --password flags, if present, take precedence

Documentation updates will follow in a separate PR

[RUN-1041]: https://snyksec.atlassian.net/browse/RUN-1041

### DIFF
--- a/lib/scan.ts
+++ b/lib/scan.ts
@@ -9,12 +9,22 @@ import { getArchivePath, getImageType } from "./image-type";
 import * as staticModule from "./static";
 import { ImageType, PluginOptions, PluginResponse } from "./types";
 
+// Registry credentials may also be provided by env vars. When both are set, flags take precedence.
+export function mergeEnvVarsIntoCredentials(
+  options: Partial<PluginOptions>,
+): void {
+  options.username = options.username || process.env.SNYK_REGISTRY_USERNAME;
+  options.password = options.password || process.env.SNYK_REGISTRY_PASSWORD;
+}
+
 export async function scan(
   options?: Partial<PluginOptions>,
 ): Promise<PluginResponse> {
   if (!options) {
     throw new Error("No plugin options provided");
   }
+
+  mergeEnvVarsIntoCredentials(options);
 
   const targetImage = options.path;
   if (!targetImage) {
--- a/test/lib/scan.spec.ts
+++ b/test/lib/scan.spec.ts
@@ -1,0 +1,74 @@
+import { mergeEnvVarsIntoCredentials } from "../../lib/scan";
+
+describe("mergeEnvVarsIntoCredentials", () => {
+  const oldEnvVars = { ...process.env };
+  const FLAG_USER = "flagUser";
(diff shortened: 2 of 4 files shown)


lib/scan.ts
```diff
@@ -9,13 +9,23 @@ import { getArchivePath, getImageType } from "./image-type";
...
+}
+
 export async function scan(
   options?: Partial<PluginOptions>,
 ): Promise<PluginResponse> {
   if (!options) {
     throw new Error("No plugin options provided");
   }
 
+  mergeEnvVarsIntoCredentials(options);
```
ivanstanev, line 27: Mutating the data is not great as it makes it harder to reason about how values change over time. If you assume they are set only once it makes it much easier to debug!
A suggestion I can give is to make the decision which variable to use closer to the place where it's needed. Right now credentials are needed only in `imageIdentifierAnalysis()`, so maybe instead of passing options.username/options.password here: https://github.com/snyk/snyk-docker-plugin/blob/095a3a020ce8db747739759f7b538f6b862bd7e1/lib/scan.ts#L106-L107 we could do something like this:
```typescript
const username = options.username || process.env.SNYK_REGISTRY_USERNAME;
const password = options.password || process.env.SNYK_REGISTRY_PASSWORD;
const archiveResult = await getImageArchive(
    targetImage,
    imageSavePath,
    username,
    password,
    options.platform,
  );
```

ivanstanev (approved):
Great work! 👏

lib/scan.ts
michael-c-snyk: Very spooky because these were both points I thought about and discussed in chat with @jmickey yesterday! Totally agree on mutation. My instinct was to use a new var for the resolved fields, but we agreed in the end that the benefits were outweighed by the risk of accidentally referring to the "username" at a later point in the code and forgetting that it may not be the actual username resolved from considering env vars too. We also talked about resolving in the env vars much later in the flow, at the latest possible moment, just like you suggested. Great minds eh! Eventually decided that doing it as near as poss to start of the plugin's scope would be preferable for similar reasons to the above, ie we could refer to "options.username" at any point and know that it would have been resolved and abstracted away from the config sources (flags, env vars) - things that the plugin logic shouldnt need to know about. Would be happy to chat over this again, let me know what you think :)

michael-c-snyk: I've realised that next time I should document these sorts of decisions in the PR at the very least, to save us all a bunch of typing!

jmickey: My reasoning for advocating for mutation in this case is that my understanding was that we pass the `options` object further down the stack. There is an argument that having the `PluginOptions` type and not utilising it further down the stack might also cause confusion.

ivanstanev: Sounds great! Thanks for explaining the reasoning

PR #306, opened 2020-11-18
[RUN-1271] Feat/identify go binaries

agatakrajewska:
- [x] Ready for review

#### What does this PR do?
refactor: Extract only not empty layers
    
    Before, applyCallbacks result was always defined, because we were matching on file path. For Go modules parsing, we also match on the filepath, but now the callback function does not necessarily return a result and we are filtering the undefined results out.

 feat: Identify go binaries
    
    We're identifying go binaries which are ELF executables, which are then returned as extractedLayers. No results from go modules are being processed into a scanResult yet.

Commit by commit, the first one is a refactor and the second one we're adding go binaries extraction.

https://snyksec.atlassian.net/browse/RUN-1271

### DIFF
--- a/lib/extractor/callbacks.ts
+++ b/lib/extractor/callbacks.ts
@@ -7,7 +7,6 @@ export async function applyCallbacks(
   fileContentStream: Readable,
 ): Promise<FileNameAndContent> {
   const result: FileNameAndContent = {};
-
   const actionsToAwait = matchedActions.map((action) => {
     // Using a pass through allows us to read the stream multiple times.
     const streamCopy = new PassThrough();
@@ -22,11 +21,17 @@ export async function applyCallbacks(
 
     return promise.then((content) => {
       // Assign the result once the Promise is complete.
-      result[action.actionName] = content;
+      if (content) {
+        result[action.actionName] = content;
+      }
     });
   });
 
   await Promise.all(actionsToAwait);
 
   return result;
+}
+
+export function isResultEmpty(result: FileNameAndContent): boolean {
+  return Object.keys(result).length === 0;
 }
--- a/lib/extractor/callbacks.ts
+++ b/lib/extractor/callbacks.ts
@@ -5,6 +5,7 @@ import { ExtractAction, FileNameAndContent } from "./types";
 export async function applyCallbacks(
   matchedActions: ExtractAction[],
   fileContentStream: Readable,
+  streamSize?: number,
 ): Promise<FileNameAndContent> {
   const result: FileNameAndContent = {};
   const actionsToAwait = matchedActions.map((action) => {
@@ -15,7 +16,7 @@ export async function applyCallbacks(
     // Queue the promise but don't await on it yet: we want consumers to start around the same time.
     const promise =
       action.callback !== undefined
-        ? action.callback(streamCopy)
+        ? action.callback(streamCopy, streamSize)
         : // If no callback was provided for this action then return as string by default.
           streamToString(streamCopy);
 
--- a/lib/go-parser/index.ts
+++ b/lib/go-parser/index.ts
@@ -1,0 +1,115 @@
+import * as elf from "elfy";
+import * as path from "path";
+import { Readable } from "stream";
+import { ExtractAction } from "../extractor/types";
+
+const ignoredPaths = [
+  path.normalize("/boot"),
+  path.normalize("/dev"),
+  path.normalize("/etc"),
+  path.normalize("/home"),
+  path.normalize("/media"),
+  path.normalize("/mnt"),
+  path.normalize("/proc"),
+  path.normalize("/root"),
+  path.normalize("/run"),
+  path.normalize("/sbin"),
+  path.normalize("/sys"),
+  path.normalize("/tmp"),
+  path.normalize("/var"),
+];
+
+function filePathMatches(filePath: string): boolean {
+  const dirName = path.dirname(filePath);
+  return (
+    !path.parse(filePath).ext &&
+    !ignoredPaths.some((ignorePath) => dirName.startsWith(ignorePath))
+  );
+}
+
+export const getGoModulesContentAction: ExtractAction = {
+  actionName: "gomodules",
+  filePathMatches,
+  callback: findGoBinaries,
+};
+
+async function findGoBinaries(
+  stream: Readable,
+  streamSize?: number,
+): Promise<any> {
+  return new Promise((resolve, reject) => {
+    const encoding = "binary";
+    const buildIdMagic = "Go";
+    const elfHeaderMagic = "\x7FELF";
+    const buildInfoMagic = "\xff Go buildinf:";
+    const elfBuildInfoSize = 64 * 1024; // ELF section headers and so ".go.buildinfo" & ".note.go.buildid" blobs are available in the first 64kb
+
+    const buffer: Buffer = Buffer.alloc(streamSize ?? elfBuildInfoSize);
+    let bytesWritten = 0;
+
+    stream.on("end", () => {
+      // Discard
+      if (bytesWritten === 0) {
+        return resolve();
+      }
+
+      const binaryFile = elf.parse(buffer);
+
+      const goBuildInfo = binaryFile.body.sections.find(
+        (section) => section.name === ".go.buildinfo",
+      );
+      // Could be found in file headers
+      const goBuildId = binaryFile.body.sections.find(
+        (section) => section.name === ".note.go.buildid",
+      );
+
+      if (!goBuildInfo && !goBuildId) {
+        return resolve();
+      } else if (goBuildInfo) {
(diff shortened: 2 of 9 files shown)


lib/extractor/callbacks.ts
```diff
@@ -20,13 +18,17 @@ export async function applyCallbacks(
         : // If no callback was provided for this action then return as string by default.
           streamToString(streamCopy);
 
-    return promise.then((content) => {
-      // Assign the result once the Promise is complete.
-      result[action.actionName] = content;
-    });
-  });
+    const content = await promise;
```
ivanstanev, line 21: I think this was intentionally made to fire all promises at the same time and we should bring it back to creating all promises and then awaiting all. If you await them one by one the stream duplication and processing cannot work because by the time the 2nd action is run, the stream is already drained/read.

lib/go-parser/index.ts
```diff
@@ -0,0 +1,104 @@
...
+};
+
+async function findGoBinaries(stream: Readable): Promise<any> {
+  return new Promise((resolve, reject) => {
+    const encoding = "binary";
+    const buildIdMagic = "Go";
+    const elfHeaderMagic = "\x7FELF";
+    const buildInfoMagic = "\xff Go buildinf:";
+
+    const result: Buffer[] = [];
```
ivanstanev, line 43: Wondering if we could just allocate a single buffer and push incoming bytes to it instead of having an array of buffers? This way we allocate the memory only once, otherwise we do it 2 times: once to build the array and once to concat all the buffers into a new buffer

lib/go-parser/index.ts
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        return resolve();
+      } else if (goBuildId) {
+        const strings = goBuildId.data
+          .toString()
+          .split(/\0+/g)
+          .filter(Boolean);
+        const go = strings[strings.length - 2];
+        const buildIdParts = strings[strings.length - 1].split("/");
```
ivanstanev, line 79: Why -2 and -1? We could add a comment/variable explaining what it means 🙏

lib/go-parser/index.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    stream.on("error", (error) => reject(error));
+
+    stream.once("data", (chunk) => {
+      const first4Bytes = chunk.toString(encoding, 0, 4);
+
+      if (first4Bytes === elfHeaderMagic) {
+        result.push(Buffer.from(chunk));
+        // Listen to next chunks only if it's an ELF executable
+        stream.addListener("data", (chunk) => result.push(Buffer.from(chunk)));
```
ivanstanev, line 100: Loving this optimization! 👏

ivanstanev (commented):
Amazing work! 👏

lib/extractor/callbacks.ts
agatakrajewska: Thanks a lot for catching, updating 👌

lib/go-parser/index.ts
agatakrajewska: We've refactored it slightly, hopefully it's clearer now, let us know otherwise :-)

lib/go-parser/index.ts
agatakrajewska: I've done a bit of research and because you can't really create a buffer of dynamic size (you have to provide size) and you can't change the size later either it might not be so straight forward. 
So we would either need to create a buffer of a size that should always be enough fo the binary data we need (for this PR it could be small, because we're only processing on the headers but the next PR we're moving to parse other sections which are further in the file) or we would need to create a buffer of a certain size and if we can't fit more into that buffer, create another buffer, not sure! 🤔 Wdyt?

ivanstanev: Is there anything in the `tar-stream` extraction where it gives us the "headers" that can tell us the expected file size? https://github.com/snyk/snyk-docker-plugin/blob/master/lib/extractor/layer.ts#L22 Something like `headers.fileSize`? Maybe we could pre-allocate a buffer of that size?

I think the risk here is loading some massive go binary in memory twice - one for each chunk of buffer of the file and another for concat()-ing all chunks.

> or we would need to create a buffer of a certain size and if we can't fit more into that buffer, create another buffer, not sure

I think something along the lines could work! Or just keep only 1 buffer and every time you receive a chunk you can concat it to the existing buffer. The issue here would be lots of allocations so may not be a great option...

I would say do what you think works OK as a first iteration, measure it, and improve later!

ivanstanev (dismissed):
Looks great! But how come some commits from a previous PR managed to appear here, is this to unblock releases?

lib/go-parser/index.ts
ivanstanev: Perfect, thanks!

lib/go-parser/index.ts
agatakrajewska:
```
    stream.once("data", (chunk) => {
      const first4Bytes = chunk.toString(encoding, 0, 4);

      if (first4Bytes === elfHeaderMagic) {
        // result.push(Buffer.from(chunk));
        buffer = Buffer.concat([buffer, Buffer.from(chunk)]);
        // Listen to next chunks only if it's an ELF executable
        stream.addListener(
          "data",
          (chunk) => (buffer = Buffer.concat([buffer, Buffer.from(chunk)])),
        );
      }
    });
```
Could do something like above, starting with: 
```
    let buffer: Buffer = new Buffer(0);
```
But, like you said, there's a lot of assignments, I'm not sure how to understand which is better option, I think this will use less memory but there's more expensive operations, what do you think?

agatakrajewska: It was a nice an easy change to do this: 
```
Buffer.from(chunk).copy(buffer, bytesWritten, 0);
        bytesWritten += chunk.length;
```
BUT the `streamSize` I am passing can be undefined (see latest commit), so I was trying to investigate another way to find the size but we can't from the stream and I don't think we know how much file we want to parse. But maybe I can check with @dkontorovskyy 🤔 Wdyt, if you have any ideas, shoot!

lib/inputs/file-pattern/static.ts
```diff
@@ -34,7 +34,8 @@ export function generateExtractAction(
   return {
     actionName: "find-files-by-pattern",
     filePathMatches: generatePathMatcher(globsInclude, globsExclude),
-    callback: (dataStream) => streamToString(dataStream, "base64"),
+    callback: (dataStream, streamSize) =>
+      streamToString(dataStream, streamSize, "base64"),
```
ivanstanev, line 38: We don't need `streamSize` in `streamToString()` I think! It's not used currently

lib/go-parser/index.ts
```diff
@@ -0,0 +1,114 @@
...
+  stream: Readable,
+  streamSize?: number,
+): Promise<any> {
+  return new Promise((resolve, reject) => {
+    const encoding = "binary";
+    const buildIdMagic = "Go";
+    const elfHeaderMagic = "\x7FELF";
+    const buildInfoMagic = "\xff Go buildinf:";
+
+    const buffer: Buffer = Buffer.alloc(streamSize ?? 64 * 1024);
```
ivanstanev, line 46: I think 64KB might be tiny, I would be happy to pre-allocate even something as big as 4MB or 8MB 😄

ivanstanev (commented):
Well done!

lib/inputs/file-pattern/static.ts
agatakrajewska: We do, because it's of type `ExtractCallback` and that now has the size. It's confusing because in here we don't use `streamToString` directly but in other cases we do.

ivanstanev: Maybe I'm not understanding but if I check the implementation of `streamToString` the streamSize parameter is not used

agatakrajewska: yes but because it used in here for example: 
```
const getFallbackOsReleaseAction: ExtractAction = {
  actionName: "os-release-fallback",
  filePathMatches: (filePath) =>
    filePath === normalizePath(OsReleaseFilePath.LinuxFallback),
  callback: streamToString,
};
```
it had to implement `ExtractCallback ` interface, I mean, I could pass undefined, if that is what you mean? Because it's not used, that is true

ivanstanev: Aa I see 👍

ivanstanev: I think it's my bad here, we shouldn't have written `callback: streamToString,` but instead be more explicit like: `callback: async (stream) => streamToString(stream)` (and in any other place that does this).

ivanstanev (approved):

lib/inputs/file-pattern/static.ts
agatakrajewska: Yes :-) Small refactor opportunity here for sure 👍

PR #307, opened 2020-11-20
Fix/typescript promise issues

agatakrajewska:
- [x] Ready for review

#### What does this PR do?

There has been typescript changes, where the value is required in the resolve callback for a new Promise. When resolve is called without arguments, we need to specify type void on the new Promise. More information: https://github.com/microsoft/TypeScript/pull/39817

Also we're pinning typescript and typescript types to specific versions to avoid similar issues in the future.

### DIFF
--- a/lib/stream-utils.ts
+++ b/lib/stream-utils.ts
@@ -42,13 +42,13 @@ async function streamToHash(
   stream: Readable,
   hashAlgorithm: string,
 ): Promise<string> {
-  return new Promise((resolve, reject) => {
+  return new Promise<string>((resolve, reject) => {
     const hash = crypto.createHash(hashAlgorithm);
     hash.setEncoding(HASH_ENCODING);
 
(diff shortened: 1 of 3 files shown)


lib/stream-utils.ts
```diff
@@ -42,13 +42,13 @@ async function streamToHash(
...
 ): Promise<string> {
-  return new Promise((resolve, reject) => {
+  return new Promise<string>((resolve, reject) => {
     const hash = crypto.createHash(hashAlgorithm);
     hash.setEncoding(HASH_ENCODING);
 
     stream.on("end", () => {
       hash.end();
-      resolve(hash.read());
+      resolve(hash.read().toString());
```
ivanstanev, line 51: What about the encoding, do we need to specify `hex`?

ivanstanev (approved):

PR #308, opened 2020-11-23
feat: parse go binary and get dependency graph for this binary

dkontorovskyy:
- [ ] Ready for review

#### What does this PR do?
Add support to scan `non-stripped` Go binaries

Algorythm: 
1. Find binaries in the image
2. Check if the binary is ELF executable.
3. Get relevant ELF sections and programs.
4. Parse them.
5. Build DepGraph

### DIFF
--- a/lib/analyzer/static-analyzer.ts
+++ b/lib/analyzer/static-analyzer.ts
@@ -1,8 +1,11 @@
 import * as Debug from "debug";
-import { DockerFileAnalysis } from "../dockerfile/types";
+import { DockerFileAnalysis } from "../dockerfile";
 import * as archiveExtractor from "../extractor";
-import { getGoModulesContentAction } from "../go-parser";
-import { getFileContent } from "../inputs";
+import {
+  getGoModulesContentAction,
+  goModulesToScannedProjects,
+} from "../go-parser";
+import { getElfFileContent, getFileContent } from "../inputs";
 import {
   getApkDbFileContent,
   getApkDbFileContentAction,
@@ -149,10 +152,14 @@ export async function analyze(
       getFileContent(extractedLayers, getJarFileContentAction.actionName),
       targetImage,
     );
+    const goModulesScanResult = await goModulesToScannedProjects(
+      getElfFileContent(extractedLayers, getGoModulesContentAction.actionName),
+    );
 
     applicationDependenciesScanResults.push(
       ...nodeDependenciesScanResults,
       ...jarFingerprintScanResults,
+      ...goModulesScanResult,
     );
   }
 
--- a/lib/go-parser/index.ts
+++ b/lib/go-parser/index.ts
@@ -1,7 +1,16 @@
 import * as elf from "elfy";
+import { eventLoopSpinner } from "event-loop-spinner";
 import * as path from "path";
 import { Readable } from "stream";
+
+import {
+  AppDepsScanResultWithoutTarget,
+  FilePathToElfContent,
+} from "../analyzer/applications/types";
 import { ExtractAction } from "../extractor/types";
+import { DepGraphFact } from "../facts";
+import { parseGoBinary } from "./parser";
+import { Elf } from "./types";
 
 const ignoredPaths = [
   path.normalize("/boot"),
@@ -18,6 +27,8 @@ const ignoredPaths = [
   path.normalize("/tmp"),
   path.normalize("/var"),
 ];
+
+export const DEP_GRAPH_TYPE = "gomodules";
 
 function filePathMatches(filePath: string): boolean {
   const dirName = path.dirname(filePath);
@@ -42,7 +53,8 @@ async function findGoBinaries(
     const buildIdMagic = "Go";
     const elfHeaderMagic = "\x7FELF";
     const buildInfoMagic = "\xff Go buildinf:";
-    const elfBuildInfoSize = 64 * 1024; // ELF section headers and so ".go.buildinfo" & ".note.go.buildid" blobs are available in the first 64kb
+    // ELF section headers and so ".go.buildinfo" & ".note.go.buildid" blobs are available in the first 64kb
+    const elfBuildInfoSize = 64 * 1024;
 
     const buffer: Buffer = Buffer.alloc(streamSize ?? elfBuildInfoSize);
     let bytesWritten = 0;
@@ -63,7 +75,16 @@ async function findGoBinaries(
         (section) => section.name === ".note.go.buildid",
       );
 
+      const interp = binaryFile.body.sections.find(
+        (section) => section.name === ".interp",
+      );
+
       if (!goBuildInfo && !goBuildId) {
+        return resolve();
+      } else if (interp) {
+        // Compiled using cgo
+        // we wouldn't be able to extract modules
+        // TODO: cgo-compiled binaries are not supported in this iteration
         return resolve();
       } else if (goBuildInfo) {
         const info = goBuildInfo.data
@@ -112,4 +133,40 @@ async function findGoBinaries(
       }
     });
   });
+}
+
+/**
+ * Build depGraphs for each Go executable
+ * @param filePathToContent
+ */
+export async function goModulesToScannedProjects(
+  filePathToContent: FilePathToElfContent,
+): Promise<AppDepsScanResultWithoutTarget[]> {
+  const scanResults: AppDepsScanResultWithoutTarget[] = [];
+
+  for (const [filePath, goBinary] of Object.entries(filePathToContent)) {
+    if (eventLoopSpinner.isStarving()) {
+      await eventLoopSpinner.spin();
+    }
+
+    const depGraph = await parseGoBinary(goBinary as Elf);
+
+    if (!depGraph) {
+      continue;
+    }
+
+    const depGraphFact: DepGraphFact = {
+      type: "depGraph",
+      data: depGraph,
+    };
+    scanResults.push({
+      facts: [depGraphFact],
+      identity: {
+        type: DEP_GRAPH_TYPE,
+        targetFile: filePath,
+      },
+    });
+  }
+
+  return scanResults;
 }
--- a/lib/go-parser/parser.ts
+++ b/lib/go-parser/parser.ts
@@ -1,0 +1,97 @@
+import * as depGraph from "@snyk/dep-graph";
+import { eventLoopSpinner } from "event-loop-spinner";
+
+import { DEP_GRAPH_TYPE } from "./";
+import { parserStrTab } from "./str-tab-parser";
+import { Elf, GoPackage, PackageVersionTable } from "./types";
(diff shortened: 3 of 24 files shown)


lib/go-parser/version-parser.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  return { name, modules };
+}
+
+/**
+ * Function finds and returns the Go version and
+ * module version information in the executable binary
+ * @param binary
+ */
+function findVers(binary: Elf) {
```
ivanstanev, line 52: Should we add a return type to the function?

lib/inputs/index.ts
```diff
@@ -20,3 +25,27 @@ export function getFileContent(
...
+
+  for (const filePath of Object.keys(extractedLayers)) {
+    for (const actionName of Object.keys(extractedLayers[filePath])) {
+      if (actionName !== searchedAction) {
+        continue;
+      }
+
+      const elfFile = extractedLayers[filePath][actionName] as Elf;
+
+      if (!(elfFile.body && elfFile.body.programs && elfFile.body.sections)) {
```
ivanstanev, line 43: This could be a type guard function

lib/go-parser/parser.ts
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Build depGraphs for each Go executable
+ * @param filePathToContent
+ */
+export async function goModulesToScannedProjects(
+  filePathToContent: FilePathToElfContent,
+): Promise<AppDepsScanResultWithoutTarget[]> {
+  const scanResults: AppDepsScanResultWithoutTarget[] = [];
+
+  for (const [filePath, goBinary] of Object.entries(filePathToContent)) {
```
ivanstanev, line 114: Is it possible any of these steps in the loop could throw? Or we want to surface any errors/bugs in our logic instead of doing a best-effort attempt?

lib/inputs/index.ts
```diff
@@ -20,3 +25,27 @@ export function getFileContent(
...
+        throw new Error("elf file expected to contain programs and sections");
+      }
+      foundAppFiles[filePath] = extractedLayers[filePath][actionName];
```
michael-c-snyk, line 46: This overwrites `foundAppFiles[filePath]` with the last valid `extractedLayers[filePath][actionName]` right? Is it possible there are multiple valid actionNames for each filePath, or will there only ever be one?

package.json
```diff
@@ -17,9 +17,9 @@
     "lint": "prettier --check \"{lib,test}/**/*.ts\" && tslint --format stylish \"{lib,test}/**/*.ts\"",
     "format": "prettier --loglevel warn --write '{lib,test}/**/*.ts' && tslint --fix --format stylish '{lib,test}/**/*.ts'",
     "test": "npm run lint && npm run unit-test",
-    "test-jest": "jest --bail --logHeapUsage",
+    "test-jest": "jest --ci --maxWorkers=3 --logHeapUsage",
```
michael-c-snyk, line 20: Do we know why `3` works best here? :)

lib/go-parser/version-parser.ts
```diff
@@ -0,0 +1,196 @@
...
+    } else {
+      readPtr = (buffer) => Number(buffer.readBigInt64LE());
+    }
+  }
+
+  const version: string = readString(
+    binary,
+    ptrSize,
+    readPtr,
+    readPtr(data.slice(16, 16 + ptrSize)),
```
ivanstanev, line 93: Is it worth adding a comment or assigning to a variable to understand what `data.slice(16, 16 + ptrSize)` is pointing at?

lib/go-parser/version-parser.ts
```diff
@@ -0,0 +1,196 @@
...
+    return result;
+  }
+
+  result.version = version;
+
+  const mod: string = readString(
+    binary,
+    ptrSize,
+    readPtr,
+    readPtr(data.slice(16 + ptrSize, 16 + 2 * ptrSize)),
```
ivanstanev, line 106: Same here - is it worth assigning to a var with a descriptive name or adding a comment?

lib/inputs/index.ts
dkontorovskyy: The combination of `file path` and `action` is always unique. And this function extracts specifically based on the action name of all `filePaths: actionResult` combination. So it's not possible to be overwritten. 
This is how we call the function https://github.com/snyk/snyk-docker-plugin/pull/308/files#diff-ef7473bc4617398b357238645eb8280929cc7617ebf63ad15e8fb1ea8d82cf34R154

lib/go-parser/version-parser.ts
```diff
@@ -0,0 +1,196 @@
...
+  );
+
+  if (mod.length >= 33 && mod[mod.length - 17] === "\n") {
+    result.mod = mod.slice(16, mod.length - 16);
+  } else {
```
ivanstanev, line 111: Should we describe what this is trying to do?

package.json
dkontorovskyy: Took it from other repos we have. Found smth online :)
https://www.peterbe.com/plog/ideal-number-of-workers-in-jest-maxWorkers

michael-c-snyk: Ah thats good to know about! TIL. Thanks!

ivanstanev (commented):
Looks amazing 👏  I have a suggestion to use https://github.com/snyk/event-loop-spinner for any loops, we do it in `rpm-parser` too even though sometimes the analysis is lightning fast ⚡

lib/go-parser/index.ts
```diff
@@ -42,7 +42,8 @@ async function findGoBinaries(
     const buildIdMagic = "Go";
     const elfHeaderMagic = "\x7FELF";
     const buildInfoMagic = "\xff Go buildinf:";
-    const elfBuildInfoSize = 64 * 1024; // ELF section headers and so ".go.buildinfo" & ".note.go.buildid" blobs are available in the first 64kb
+    // ELF section headers and so ".go.buildinfo" & ".note.go.buildid" blobs are available in the first 64kb
```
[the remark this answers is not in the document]
dkontorovskyy: Yes, indeed :)

lib/go-parser/parser.ts
```diff
@@ -0,0 +1,135 @@
...
+import {
+  AppDepsScanResultWithoutTarget,
+  FilePathToElfContent,
+} from "../analyzer/applications/types";
+import { DepGraphFact } from "../facts";
+import { parserStrTab } from "./str-tab-parser";
+import { Elf, GoPackage, PackageVersionTable } from "./types";
+import { extractModulesFromBinary } from "./version-parser";
+
+const DEP_GRAPH_TYPE = "gomodules";
```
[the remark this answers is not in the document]
dkontorovskyy: https://github.com/snyk/registry/blob/f53372e302eec80fc8b0573ae26246d8da89fb25/src/lib/package-managers.ts#L44

lib/analyzer/static-analyzer.ts
```diff
@@ -1,8 +1,9 @@
 import * as Debug from "debug";
-import { DockerFileAnalysis } from "../dockerfile/types";
+import { DockerFileAnalysis } from "../dockerfile";
 import * as archiveExtractor from "../extractor";
 import { getGoModulesContentAction } from "../go-parser";
-import { getFileContent } from "../inputs";
+import { goModulesToScannedProjects } from "../go-parser/parser";
```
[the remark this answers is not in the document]
dkontorovskyy: Sounds like a legit suggestion

lib/go-parser/parser.ts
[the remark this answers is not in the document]
ivanstanev: Danke schön

lib/go-parser/version-parser.ts
dkontorovskyy: I want code to mirror go source code, added comments to explain why is that happening.

lib/go-parser/version-parser.ts
dkontorovskyy: I want code to mirror go source code, added comments to explain why is that happening.

lib/go-parser/parser.ts
dkontorovskyy: Can't find a place with a possible error. My only guess we might get thrown out if `address`, `offset`, etc would be out of range of `Number`, but don't think it would actually be there.

dkontorovskyy:
@ivanstanev does it make sense to use `event-loop-spinner`, cos all of the parsing is actually synchronous. Why would we want to kick in `async` function inside?

lib/go-parser/version-parser.ts
dkontorovskyy: Done :)

ivanstanev:
> @ivanstanev does it make sense to use `event-loop-spinner`, cos all of the parsing is actually synchronous. Why would we want to kick in `async` function inside?

Even if the parsing doesn't take long (maybe 100ms) we still want to spin the loop to avoid blocks, I think unless we can guarantee the parsing takes something super quick like <10ms the rule of thumb is to spin. This plugin will be used not only in the CLI (which may tolerate blocks) but also in our internal services (like DRA) which may not be able to tolerate blocks!

dkontorovskyy:
@ivanstanev Added 2 spins in the parts I believe can be blocking :)

test/lib/analyzer/gomodules.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+  it("should return expected scannedProject[] result", async () => {
+    // Arrange
+    const fixturePath = getFixture("docker-archives/docker-save/yq.tar");
+    const imageNameAndTag = `docker-archive:${fixturePath}`;
+
+    // Act
+    const pluginResult = await scan({
+      path: imageNameAndTag,
+      "app-vulns": true,
+    });
```
ivanstanev, line 16: This looks like a full system tests which we already do in `test/system/application-scans/gomodules.spec.ts`, do you think there is any need for this one too?

test/lib/analyzer/gomodules.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+import { getFixture } from "../../util";
+
+describe("goModulesToScannedProjects function", () => {
+  it("should return expected scannedProject[] result", async () => {
+    // Arrange
+    const fixturePath = getFixture("docker-archives/docker-save/yq.tar");
+    const imageNameAndTag = `docker-archive:${fixturePath}`;
+
+    // Act
+    const pluginResult = await scan({
```
agranado2k, line 13: I agree with @ivanstanev. I'm wondering if it should be a system test.

agranado2k (commented):
Amazing work man! 🎉 
Added one comment.

ivanstanev (dismissed):

agranado2k (dismissed):
Amazing work! LGTM! 🎉

agranado2k (approved):
LGTM! 🎉

PR #311, opened 2020-12-15
Append latest if an image without a tag was provided

RotemS:
- [x] Ready for review

#### What does this PR do?

Most of the docker commands implicitly add `latest` to images without tags. In case of `docker save`, the behavior is not the same and can result in performance issues as well as inconsistent results.

Using `scan` with/without latest tag, with multiple different images from the same repo, should result in the same `scanResults`


agranado2k (approved):
LGTM! 🎉

PR #312, opened 2020-12-23
user instruction info from image

RotemS:
- [x] Ready for review

#### What does this PR do?

Adds a new returned fact to the plugin response - `autoDetectedLayers`
This will extract the layer information from the image's config file, removing non user instruction layers or empty ones, and detect the packages installed (same behavior as the `dockerfileAnalysis`).

**Known gaps**
1. Does not resolve variables at the moment
2. The definition on how a "RUN" instruction looks like in the config is naive, assuming the instruction starts with `/bin/sh` or `RUN /bin/sh`
3. Since the instructions are not identical between a dockerfile and the config, the results will not be identical.

`scan` should now return another fact `autoDetectedLayers`
If there are user instruction layers (and not only base image layers), it will return the packages and layers.


lib/dockerfile/instruction-parser.ts
```diff
@@ -6,11 +6,12 @@ export {
   getLayersFromPackages,
   getPackagesFromDockerfile,
   instructionDigest,
+  getPackagesFromRunInstructions,
 };
 
 // Naive regex; see tests for cases
 // tslint:disable-next-line:max-line-length
-const installRegex = /\s*(rpm\s+-i|rpm\s+--install|apk\s+((--update|-u|--no-cache)\s+)*add(\s+(--update|-u|--no-cache))*|apt-get\s+((--assume-yes|--yes|-y)\s+)*install(\s+(--assume-yes|--yes|-y))*|apt\s+((--assume-yes|--yes|-y)\s+)*install|yum\s+install|aptitude\s+install)\s+/;
+const installRegex = /(rpm\s+-i|rpm\s+--install|apk\s+((--update|-u|--no-cache)\s+)*add(\s+(--update|-u|--no-cache))*|apt-get\s+((--assume-yes|--yes|-y)\s+)*install(\s+(--assume-yes|--yes|-y))*|apt\s+((--assume-yes|--yes|-y)\s+)*install|yum\s+install|aptitude\s+install)\s+/;
```
shaimendel, line 14: did you remove the `\s*` on purpose? why?

lib/dockerfile/instruction-parser.ts
```diff
@@ -54,7 +55,10 @@ function getPackagesFromRunInstructions(runInstructions: string[]) {
 
         packages.forEach((pkg) => {
           // Use package name without version as the key
-          const name = pkg.split("=")[0];
+          let name = pkg.split("=")[0];
+          if (name.slice(0, 1) === "$") {
```
shaimendel, line 59: isn't `name.slice(0,1)` equal to just `name[0]`? might be more readable, not critical

lib/dockerfile/instruction-parser.ts
```diff
@@ -64,6 +68,17 @@ function getPackagesFromRunInstructions(runInstructions: string[]) {
   }, {});
 }
 
+function removeRunDefFromInstruction(instruction: string) {
+  let cleanedInstruction = instruction;
+  const runDefs = ["RUN ", "/bin/sh ", "RUN /bin/sh"];
```
shaimendel, line 73: what about the `-c` ? of `/bin/sh -c`

lib/extractor/docker-archive/index.ts
```diff
@@ -34,3 +38,32 @@ export function getPlatformFromConfig(
...
+  const layers = getLayersFromPackages(packages);
+  return { packages, layers };
+}
+
+export function getUserInstructionLayersFromConfig(imageConfig) {
+  const diffInHours = (d1, d2) => Math.abs(d1 - d2) / 1000 / (60 * 60);
+  const maxDiffInHours = 5;
+
+  const history = imageConfig.history;
+  const lastInstructionTime = new Date(history.slice(-1)[0].created);
```
shaimendel, line 57: does that `-1` work in node? #til!

lib/types.ts
```diff
@@ -233,3 +234,20 @@ export interface Options {
   isDockerUser?: boolean;
   config?: any;
 }
+
+export interface AutoDetectedLayers {
+  packages: PackagesFromUserInstructions;
+  layers: LayersFromUserInstructions;
+}
+
+export interface PackagesFromUserInstructions {
```
shaimendel, line 243: how come this is the first time we need to define this interface? isn't it used by the regular dockerfile analysis?

shaimendel (commented):
Amazing PR!

lib/dockerfile/instruction-parser.ts
RotemS: Yep, since there are a lot of tests around this regex, I made an assumption that if it didn't break any existing tests it should be ok.

The reason I removed is that if the leading whitespace isn't remove, this should also support some other cases with instruction from config - for example the `/bin/bash -c` since the `-c` is recognized as a flag. Should work for other flags following that as well. Let me know if that makes sense :)

lib/dockerfile/instruction-parser.ts
RotemS: That's the answer to the other question - https://github.com/snyk/snyk-docker-plugin/pull/312#discussion_r548010339 :)

lib/types.ts
RotemS: Was about to comment here since I'm not 100% on this decision. We do already have `DockerFilePackages` and `DockerFilePackages` in https://github.com/snyk/snyk-docker-plugin/blob/bf9afd0fc80fec6b0d098b321bdbf570930fa5df/lib%2Fdockerfile%2Ftypes.ts#L13
I didn't want to re-use the same types since it felt confusing, but the right thing might be to change the dockerfile ones to use the more generic type?

lib/dockerfile/instruction-parser.ts
michael-c-snyk: or `name.startsWith("$")`?

lib/extractor/docker-archive/index.ts
```diff
@@ -34,3 +38,32 @@ export function getPlatformFromConfig(
...
+    .filter((instruction) => !instruction.empty_layer)
+    .map((instruction) => instruction.created_by);
+
+  const packages = getPackagesFromRunInstructions(runInstructions);
+  const layers = getLayersFromPackages(packages);
+  return { packages, layers };
+}
+
+export function getUserInstructionLayersFromConfig(imageConfig) {
+  const diffInHours = (d1, d2) => Math.abs(d1 - d2) / 1000 / (60 * 60);
```
michael-c-snyk, line 53: Have we looked at using Luxon for time/date calculations? There are a lot of hairy edge cases around time and date that it handles, plus it gives a nice readable API.

test/lib/instructions-parser.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+    [["apt install curl  vim"], ["curl", "vim"]],
+    [["apt install curl wget vim"], ["vim", "curl", "wget"]],
+    [["apt install curl && apt install wget"], ["curl", "wget"]],
+    [['apt install curl; apt install vim; echo "bitwise"'], ["curl", "vim"]],
+    [
+      ["apt install curl && apt       install vim", "apt install   -y  wget"],
+      ["curl", "vim", "wget"],
+    ],
+    [["apt install 389-admin"], ["389-admin"]],
+  ];
```
michael-c-snyk, line 25: nice chunk of test cases :)

lib/dockerfile/instruction-parser.ts
```diff
@@ -59,6 +68,17 @@ function getPackagesFromRunInstructions(
...
```
michael-c-snyk, line 73: Maybe again part of the old implementation but does this handle `RUN /bin/bash -l -c` ?

test/lib/instructions-parser.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+    [["    apt   install   curl   vim   "], ["curl", "vim"]],
+    [["apt install curl  vim"], ["curl", "vim"]],
+    [["apt install curl wget vim"], ["vim", "curl", "wget"]],
+    [["apt install curl && apt install wget"], ["curl", "wget"]],
+    [['apt install curl; apt install vim; echo "bitwise"'], ["curl", "vim"]],
+    [
+      ["apt install curl && apt       install vim", "apt install   -y  wget"],
+      ["curl", "vim", "wget"],
+    ],
+    [["apt install 389-admin"], ["389-admin"]],
```
michael-c-snyk, line 24: Just checking that things like nvm install, rvm install etc are out of scope for us?

michael-c-snyk: Also, what about curl commands that are piped into bash, eg `RUN \curl -sSL https://get.rvm.io | bash -s stable`?

michael-c-snyk (commented):
Awesome piece of work! Few questions to nitpick and check my own understanding :)

test/lib/instructions-parser.spec.ts
RotemS: Stolen from the original tests for instruction parser 🙈

lib/extractor/docker-archive/index.ts
shaimendel: one last question on that :) any chance that `imageConfig.history` can be empty? should we check that?

lib/types.ts
shaimendel: what should make it more generic? it looks like `PackagesFromUserInstructions` and `DockerFilePackages` are the same (didn't talk about `DockerFileLayers`)

test/lib/instructions-parser.spec.ts
shaimendel: we support only commands that use the package manager :) so only `apt`/`rpm`/`apk` commands for now. regarding `curl`ing - we partially use nodejs/java that is installed that way with the key binaries feature

lib/extractor/docker-archive/index.ts
RotemS: Personally I feel that a lib for this one line is a bit much, especially since it's not too important that it will be right on the second (there's a lot of buffer time)
WDYT?

lib/dockerfile/instruction-parser.ts
RotemS: It related to the tiny change to `installRegex`, see https://github.com/snyk/snyk-docker-plugin/pull/312#discussion_r548010339

lib/extractor/docker-archive/index.ts
RotemS: That's a good question. I'm not sure. I want to say that it doesn't make sense that it would be empty, but worth a check. Will add!

lib/types.ts
RotemS: Making sure I followed, you mean that `dockerfileAnalysis` should also return `PackagesFromUserInstructions`? 
Or keep it seperate?

shaimendel: if `DockerFilePackages` and `PackagesFromUserInstructions` so maybe we can just use one of them?

lib/extractor/docker-archive/index.ts
michael-c-snyk: Don't have a strong opinion for this use case

lib/extractor/docker-archive/index.ts
```diff
@@ -34,3 +38,35 @@ export function getPlatformFromConfig(
...
+  return { dockerfilePackages, dockerfileLayers };
+}
+
+export function getUserInstructionLayersFromConfig(imageConfig) {
+  const diffInHours = (d1, d2) => Math.abs(d1 - d2) / 1000 / (60 * 60);
+  const maxDiffInHours = 5;
+
+  const history = imageConfig.history;
+  if (!history) {
+    throw new Error("Failed to extract history from image config");
```
shaimendel, line 58: what will happen in this case? will the entire scan fail? should we log something and return [] instead?

michael-c-snyk (approved):

agatakrajewska (approved):

PR #313, opened 2021-01-04
fix: return transitive packages when getting auto detected packages

RotemS:
- [x] Ready for review

#### What does this PR do?
When detecting user instruction layers from the image config, also checking for transitive dependencies and returning them as part of the packages found.


agatakrajewska:
Could we possibly add a tiny bit more detailed info to the commit, like the line from the PR: "When detecting user instruction layers from the image config, also checking for transitive dependencies and returning them as part of the packages found"
This: "fix: return transitive packages as well" - is maybe not super clear when read without context, wdyt?

lib/response-builder.ts
```diff
@@ -165,14 +173,14 @@ function collectDockerfilePkgs(
     return;
   }
 
-  return getDockerfileDependencies(dockerAnalysis.dockerfilePackages, deps);
```
agatakrajewska, line 168: I'd rather be keen to make any renaming changes in separate "refactor" commits, it makes it a tad cleaner, not a deal breaker in this case though as it's a small change and easy to read, I'll leave it to you!

agatakrajewska (approved):

PR #317, opened 2021-01-21
[CAP-120] - Extracting content manifests from layers

RotemS:
- [x] Ready for review

#### What does this PR do?

*Context*
Red Hat images contain repositories information for each layer.
This information is found in `/root/buildinfo/content_manifests/`

This PR is for extracting this information while scanning the image.

By commits, first one is a refactor followed by the feature itself. 
Tests describe the content we're looking for and expected results.


.gitignore
```diff
@@ -9,6 +9,7 @@ yarn.lock
 test/fixtures/ls-output/xxx.txt
 .nyc_output/
 .vscode/
+tsconfig.local.json
```
agatakrajewska, line 12: There is a way to do it without having to include all those local `json` file in gitignore: 
In your project dir you'll have `.git/info/exclude` file and you can include it there like: 
```
# git ls-files --others --exclude-from=.git/info/exclude
# Lines that start with '#' are comments.
# For a project mostly in C, the following would be a good set of
# exclude patterns (uncomment them if you want to use them):
# *.[oa]
# *~

# vscode local settings directory
.vscode

#tests output
.nyc_output

tsconfig.local.json
```

agatakrajewska: Just a suggestion though, if we see a benefit of including in gitignore, why not, this way we do it for the whole team as well

lib/inputs/rhel/static.ts
```diff
@@ -0,0 +1,27 @@
+import { ExtractAction, ExtractedLayers } from "../../extractor/types";
+import { streamToJson } from "../../stream-utils";
+
+export const getRedHatReposContentAction: ExtractAction = {
+  actionName: "redhat-content-manifests",
+  filePathMatches: (filePath) => isRedHatContentManifest(filePath),
+  callback: streamToJson,
```
agatakrajewska, line 7: I believe you could already here add a type here that would only return us object like: 
```
{
  layerIndex: number,
  repositories: [ ]
}
```
and use `streamToJson<T>` as a callback:
https://github.com/snyk/snyk-docker-plugin/blob/0026c770b4dea73feff4a9547fcbee653006d2f6/lib/stream-utils.ts#L69

Then we already have it handled and don't have to return whole json :-)

agatakrajewska: Then we don't also have to worry about stuff we don't need in tests

lib/inputs/rhel/static.ts
```diff
@@ -0,0 +1,27 @@
+import { ExtractAction, ExtractedLayers } from "../../extractor/types";
+import { streamToJson } from "../../stream-utils";
+
+export const getRedHatReposContentAction: ExtractAction = {
+  actionName: "redhat-content-manifests",
+  filePathMatches: (filePath) => isRedHatContentManifest(filePath),
```
michael-c-snyk, line 6: Is there a reason for the isRedHatContentManifest to be wrapped in another function here, instead of using `filePathMatches: isRedHatContentManifest,`? :)

lib/inputs/rhel/static.ts
```diff
@@ -0,0 +1,27 @@
...
+  callback: streamToJson,
+};
+export function getRedHatReposFromExtractedLayers(
+  extractedLayers: ExtractedLayers,
+) {
```
michael-c-snyk, line 11: Is it worth adding a return type?

lib/inputs/rhel/static.ts
```diff
@@ -0,0 +1,27 @@
...
+  const reposByLayer = {};
+  for (const filePath in extractedLayers) {
+    if (isRedHatContentManifest(filePath)) {
+      const content = extractedLayers[filePath][
+        "redhat-content-manifests"
+      ] as any;
+      const layerIndex = content.metadata.image_layer_index;
+      reposByLayer[layerIndex] = content.content_sets;
```
agatakrajewska, line 19: I think I'd like this object to more explicitly mention repositories, so then it's going to be clearer to read the code when doing all the matching stuff, wdyt??

lib/inputs/rhel/static.ts
```diff
@@ -0,0 +1,27 @@
...
+};
+export function getRedHatReposFromExtractedLayers(
+  extractedLayers: ExtractedLayers,
+) {
+  const reposByLayer = {};
+  for (const filePath in extractedLayers) {
+    if (isRedHatContentManifest(filePath)) {
+      const content = extractedLayers[filePath][
+        "redhat-content-manifests"
+      ] as any;
```
michael-c-snyk, line 17: Could this be undefined?

michael-c-snyk (commented):
Nice and readable! Couple of questions added

lib/inputs/rhel/static.ts
RotemS: Leaving an FYI note - we discussed offline, not applicable currently. Adding the type only

.gitignore
RotemS: Thanks for the new knowledge! Personally I prefer widely used files to be in gitignore, but will definitely remember in the future 💯

lib/inputs/rhel/static.ts
RotemS: No reason 🙈 Missed it, thanks!

lib/inputs/rhel/static.ts
RotemS: Since it's using the same function to determine, I want to say no. But, I'm wrapping in a try catch as a better practice in case something goes horribly wrong there (malformed files etc). Thanks!

lib/inputs/rhel/static.ts
```diff
@@ -0,0 +1,37 @@
...
+  for (const filePath in extractedLayers) {
+    if (isRedHatContentManifest(filePath)) {
+      try {
+        const content = extractedLayers[filePath][
+          "redhat-content-manifests"
+        ] as any;
+        const layerIndex = content.metadata.image_layer_index;
+        reposByLayer[layerIndex] = content.content_sets as RedHatRepo;
+      } catch (err) {
+        throw new Error("Failed to extract repos information");
```
michael-c-snyk, line 28: So this will give up altogether if a single extracted layer is bad? Is this defintely what we want, rather than doing some kind of "this extracted layer is bad so I'll skip it, but I'll still handle the other ones that look legit" thing? (Genuine question, I don't know how we handle similar issues elsewhere :) )

RotemS: Great catch, I think it's way better to not break everything. Will change 💯

lib/extractor/types.ts
```diff
@@ -76,3 +76,9 @@ export interface DetectedImageLayers {
   packages;
   layers;
 }
+
+export interface RedHatRepos {
```
agatakrajewska, line 80: ❤️

agatakrajewska (approved):

lib/inputs/rhel/static.ts
```diff
@@ -0,0 +1,36 @@
...
+): RedHatRepos {
+  const reposByLayer = {};
+  for (const filePath in extractedLayers) {
+    if (isRedHatContentManifest(filePath)) {
+      const content = extractedLayers[filePath][
+        "redhat-content-manifests"
+      ] as any;
+      const layerIndex = content?.metadata?.image_layer_index;
+      const repos = content?.content_sets;
+      if (layerIndex && repos) {
```
agatakrajewska, line 26: 👍

PR #319, opened 2021-02-05
feat: Extract labels for docker-archive based images

jurglic:
Extract all image labels from docker-archive based images from their image config and include them in the scan result.

#### What does this PR do?

It adds extraction of labels, adds one test example, and updates the test snapshots that are affected by this change.

At the beginning, the change is quite small and most of it is in types.

You can run scan on an image and track the scan result given by plugin... I've done through DRA tho.

https://snyksec.atlassian.net/browse/DC-1093

#### Additional questions

The question is, if do we need or want to do similarly for oci-archive image annotations? 
- Not at the moment. [More info here](https://snyk.slack.com/archives/C01EPSMGZML/p1612885007094500?thread_ts=1611595377.023500&cid=C01EPSMGZML).


lib/analyzer/types.ts
```diff
@@ -77,6 +77,7 @@ export interface StaticAnalysis {
   autoDetectedUserInstructions?: DockerFileAnalysis;
   applicationDependenciesScanResults: AppDepsScanResultWithoutTarget[];
   manifestFiles: ManifestFile[];
+  imageLabels?: { [key: string]: string };
```
Shesekino, line 80: why is this optional?

jurglic: Hi Amir 👋 
It's because we only extract labels from images based on docker-archive, and not also oci-archive based images which are only supported in cli. (We have just decided that we are taking this trade-off, you can see conversation [here](https://snyk.slack.com/archives/C01EPSMGZML/p1612885007094500?thread_ts=1611595377.023500&cid=C01EPSMGZML))

Shesekino: that's okay, and thanks for the context 🙏
but why have the property optional, and not just leave it empty when it's "off"?

jurglic: Sure, that would also work.
If you prefer it I can change now, or we change it later when we add oci-archive support.

Shesekino: now is preferable please 🙏

agatakrajewska: I'm missing something in here probably, but why is empty object better than `undefined`? Isn't `undefined` more explicit and easier to check later when we're creating the fact, so if we change to empty object we have to modify this: 
```
 if (depsAnalysis.imageLabels)
 ```
 To be more complicated checking if the object is empty?

agatakrajewska (approved):

lib/analyzer/types.ts
Shesekino: it's probably more of a style choice rather than better/worse.

in my opinion, being very explicit and deterministic about what our types look like is preferable.
```undefined``` is the opposite of explicit because if it's ```undefined``` then it would just be missing from the object entirely instead of being present with an empty value.

I prefer "actively" knowing than no image labels were associated with a payload over having it implicitly hidden.
that's just my style though!

jurglic: In that specific case, I think it's actually better, as we don't really need that conditional.. and can just iterate over labels. Hence simpler and cleaner code.

It's a small change with different trade-offs... imho you can argue on both sides. :)

agatakrajewska: I see I see! Thanks Amir! You're right, I like the explanation! I guess coming from the c# world I'm very used to/like nullable props and we do use that approach more in the plugin (I would have to confirm that information hehe) but it's great to consider different approaches :-)

PR #320, opened 2021-02-09
Fix/try catch go parsing

ivanstanev:
- [x] Ready for review

#### What does this PR do?

Ensure to capture and rethrow errors from image layer parsing.

Right now any errors that occur during image layer parsing are not properly caught (because they occur either in an async or EventEmitter context) and result in uncaught exceptions.

One of these places is the Go binaries parser, which we now properly try-catch and reject the promise that wraps it.
Additionally we need to try-catch the image layer parsing process as a whole, which was missing before.

[Alerts in Slack](https://snyk.slack.com/archives/C01KWP7CKQC/p1612857668002100) and [discussion](https://snyk.slack.com/archives/C01KWP7CKQC/p1612857959002400)
[Node.js documentation on error propagation and interception](https://nodejs.org/docs/latest/api/errors.html#errors_error_propagation_and_interception)

### DIFF
--- a/lib/extractor/layer.ts
+++ b/lib/extractor/layer.ts
@@ -1,9 +1,12 @@
+import * as Debug from "debug";
 import * as gunzip from "gunzip-maybe";
 import * as path from "path";
 import { Readable } from "stream";
 import { extract, Extract } from "tar-stream";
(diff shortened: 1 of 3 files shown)


agranado2k (approved):
LGTM! 🎉

PR #323, opened 2021-02-17
fix: Added filtering history entries with no "created_by"

agatakrajewska:
- [x] Ready for review

#### What does this PR do?
For docker archives we extract auto detected user instructions from the image config. Some of the entries are missing "created_by" so we were throwing an error - this PR is fixing that.

### DIFF
--- a/lib/extractor/docker-archive/index.ts
+++ b/lib/extractor/docker-archive/index.ts
@@ -41,7 +41,7 @@ export function getPlatformFromConfig(
 
 export function getDetectedLayersInfoFromConfig(imageConfig) {
   const runInstructions = getUserInstructionLayersFromConfig(imageConfig)
-    .filter((instruction) => !instruction.empty_layer)
+    .filter((instruction) => !instruction.empty_layer && instruction.created_by)
     .map((instruction) => instruction.created_by.replace("# buildkit", ""));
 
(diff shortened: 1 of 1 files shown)


RotemS (dismissed):

lib/extractor/docker-archive/index.ts
```diff
@@ -41,7 +41,9 @@ export function getPlatformFromConfig(
 
 export function getDetectedLayersInfoFromConfig(imageConfig) {
   const runInstructions = getUserInstructionLayersFromConfig(imageConfig)
-    .filter((instruction) => !instruction.empty_layer)
+    .filter(
+      (instruction) => !instruction.empty_layer || !instruction.created_by,
```
agranado2k, line 45: Hey, just check if I got it. We're filtering for instructions that doesn't have `empty_layer` or `created_by` and we're mapping to convert the content from `created_by` from `"# buildkit"` to `""`. Not sure if I'm missing something, but I think it won't work.

agranado2k (commented):
A comment.

lib/extractor/docker-archive/index.ts
agatakrajewska: You're right! Fixed, we're just re running to check 👍 Thanks!

agranado2k (approved):
Brilliant! 🎉

PR #330, opened 2021-03-22
test: show bug with being unable to scan CentOS 6 images

ivanstanev:
- [x] Ready for review

#### What does this PR do?

This bug happens because the image in question has 2 OS release files - lsb_release and centos_release.
Our detection logic runs the LSB parser first. Unfortunately, in this specific image lsb_release is malformed and does not contain enough information about the OS release. The real OS release data is in centos_release, but our parser throws an error before examining it.

[Jira ticket #9475](https://snyk.zendesk.com/agent/tickets/9475)

I'll create the follow up fix for this :)

### DIFF
--- a/test/system/bugs/centos6.spec.ts
+++ b/test/system/bugs/centos6.spec.ts
@@ -1,0 +1,23 @@
+import { scan } from "../../../lib";
+import { execute } from "../../../lib/sub-process";
+
+describe("BUG: CentOS 6 image cannot be scanned", () => {
+  afterAll(async () => {
+    await execute("docker", [
+      "image",
+      "rm",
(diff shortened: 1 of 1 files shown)


ivanstanev:
```shell
 FAIL  test/system/bugs/image-layers.spec.ts (10.235 s, 68 MB heap size)
  ● demonstrates a potential bug with image layers › should return different image layers when pulling with docker and with pull library

    ENOENT: no such file or directory, open '/tmp/tmp-163-ybVdfDPXcWyC/a2f8b8a6-0073-4677-8b1f-82fabf0e62bf/image.tar'
```

Is this a flake?

agranado2k (approved):
LGTM!

PR #332, opened 2021-03-22
fix: use the correct debug package to log errors

ivanstanev:
- [x] Ready for review

#### What does this PR do?

This was previously logging a console message because the wrong `debug` function was used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,20 +46,21 @@
     "uuid": "^8.2.0"
   },
   "devDependencies": {
+    "@types/debug": "^4.1.5",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -46,20 +46,21 @@
     "uuid": "^8.2.0"
   },
   "devDependencies": {
+    "@types/debug": "^4.1.5",
```
ivanstanev, line 49: Installed the @types/debug here, the rest is `npm` rearranging imports

agranado2k (approved):

PR #336, opened 2021-04-18
feat: Add hash algorithm prefix ('sha256:') to docker imageId [RHINO-45]

snaftaly:
The prefix is added to the imageId to 

- [x] Ready for review

#### What does this PR do?
This PR adds the hash algorithm prefix (sha256:) to the docker imageId.

Start the review in lib/extractor/docker-archive/index.ts getImageIdFromManifest() funciton.

This PR aligns with the id shown in 'docker manifest' and 'docker images --no-trunc' commands, and in v2 API get manifest. It aligns with the OCI [spec](https://github.com/opencontainers/image-spec/blob/master/config.md#imageid) and the imageId returned from OCI images.

The imageId prefix is added as part of the work for querying images scan history by their imageId in Snyk's external API: [RHINO-45](https://snyksec.atlassian.net/browse/RHINO-45)

### DIFF
--- a/lib/extractor/docker-archive/index.ts
+++ b/lib/extractor/docker-archive/index.ts
@@ -3,6 +3,7 @@ import {
   getLayersFromPackages,
   getPackagesFromRunInstructions,
 } from "../../dockerfile/instruction-parser";
+import { HashAlgorithm } from "../../types";
 
 import { DockerArchiveImageConfig, DockerArchiveManifest } from "../types";
 export { extractArchive } from "./layer";
@@ -15,7 +16,12 @@ export function getImageIdFromManifest(
   manifest: DockerArchiveManifest,
 ): string {
   try {
-    return manifest.Config.split(".")[0];
+    const imageId = manifest.Config.split(".")[0];
+    if (imageId.includes(":")) {
+      // imageId includes the algorithm prefix
+      return imageId;
+    }
+    return `${HashAlgorithm.Sha256}:${imageId}`;
   } catch (err) {
     throw new Error("Failed to extract image ID from archive manifest");
   }
--- a/test/lib/extractor/docker-archive/index.spec.ts
+++ b/test/lib/extractor/docker-archive/index.spec.ts
@@ -1,0 +1,36 @@
+import { getImageIdFromManifest } from "../../../../lib/extractor/docker-archive";
+import { DockerArchiveManifest } from "../../../../lib/extractor/types";
+
+describe("getImageIdFromManifest", () => {
+  describe("when manifest config string contains algorithm prefix", () => {
+    it("returns the imageId with prefix", () => {
+      const manifest: DockerArchiveManifest = {
+        Config:
+          "sha256:2565821efb5e5b95b36541004fa0287732a11f97a4a0ff807cc065746f82538.json",
+        RepoTags: [],
+        Layers: [],
+      };
+
+      const imageId = getImageIdFromManifest(manifest);
+      expect(imageId).toEqual(
+        "sha256:2565821efb5e5b95b36541004fa0287732a11f97a4a0ff807cc065746f82538",
+      );
+    });
+  });
+
+  describe("when manifest config string does not contain algorithm prefix", () => {
+    it("returns the imageId with prefix", () => {
+      const manifest: DockerArchiveManifest = {
+        Config:
+          "2565821efb5e5b95b36541004fa0287732a11f97a4a0ff807cc065746f82538.json",
+        RepoTags: [],
+        Layers: [],
+      };
+
+      const imageId = getImageIdFromManifest(manifest);
+      expect(imageId).toEqual(
+        "sha256:2565821efb5e5b95b36541004fa0287732a11f97a4a0ff807cc065746f82538",
+      );
+    });
+  });
+});
--- a/test/system/app-os/__snapshots__/globs.spec.ts.snap
+++ b/test/system/app-os/__snapshots__/globs.spec.ts.snap
@@ -1862,7 +1862,7 @@ Object {
           "type": "depGraph",
         },
         Object {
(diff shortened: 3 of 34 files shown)


lib/extractor/docker-archive/index.ts
```diff
@@ -15,7 +16,13 @@ export function getImageIdFromManifest(
   manifest: DockerArchiveManifest,
 ): string {
   try {
-    return manifest.Config.split(".")[0];
+    const imageId = manifest.Config.split(".")[0];
+    const hashAlgorithmPrefix = `${HashAlgorithm.Sha256}:`;
```
Shesekino, line 20: what happens if they use a different hashing algorithm?

lib/extractor/docker-archive/index.ts
```diff
@@ -15,7 +16,13 @@ export function getImageIdFromManifest(
...
+    if (imageId.includes(":")) {
+      // imageId includes the algorithm prefix
+      return imageId;
+    }
+    return `${hashAlgorithmPrefix}${imageId}`;
```
Shesekino, line 25: I'd separate the prefix from the delimiter, so we'll have `${hashAlgorithm}:{imageId}` instead of including the `:`

lib/extractor/docker-archive/index.ts
```diff
@@ -15,7 +16,13 @@ export function getImageIdFromManifest(
   manifest: DockerArchiveManifest,
 ): string {
   try {
-    return manifest.Config.split(".")[0];
+    const imageId = manifest.Config.split(".")[0];
+    const hashAlgorithmPrefix = `${HashAlgorithm.Sha256}:`;
+    if (imageId.includes(":")) {
```
Shesekino, line 21: isn't it always supposed to include the algorithm according to the spec?

snaftaly: The behavior today is that we take the imageId from the manifest.json Config attribute's value, which is the imageSha without the prefix, with a '.json' suffix. i.e. today there is no prefix.
As [mentioned](https://snyk.slack.com/archives/C01EPSMGZML/p1618420140094600?thread_ts=1618396754.091000&cid=C01EPSMGZML) by @agatakrajewska, the documentation of docker save is not the best, and so I added this check to make sure we don't accidentally end up with having the prefix twice, although I don't have a real example for this.

lib/extractor/docker-archive/index.ts
snaftaly: That's a good question. From what I read [here](https://windsock.io/explaining-docker-image-ids/#:~:text=A%20Docker%20image's%20ID%20is,when%20an%20image%20is%20pushed), it looks like the value of manifest.Config is the imageId, which is a SHA256 hash of the image's JSON configuration object.
I didn't find a formal documentation, but added my  findings [here](https://snyk.slack.com/archives/C01EPSMGZML/p1618416730094000?thread_ts=1618396754.091000&cid=C01EPSMGZML).

Shesekino: your link looks like a blog post based on Docker, not OCI's image spec.
I'd stick to the spec - if the spec doesn't stick to a specific hashing algorithm, why should we?

lib/extractor/docker-archive/index.ts
Shesekino: I'm missing something, the spec for the manifest's config is described here: https://github.com/opencontainers/image-spec/blob/master/manifest.md
the ```config``` component is described as a ``` descriptor``` which is defined here: https://github.com/opencontainers/image-spec/blob/master/descriptor.md
its properties include a digest which is defined here: https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests
(sorry for the long link chain 😂)
but it says the digest must adhere to this grammar:
```
digest                ::= algorithm ":" encoded
algorithm             ::= algorithm-component (algorithm-separator algorithm-component)*
algorithm-component   ::= [a-z0-9]+
algorithm-separator   ::= [+._-]
encoded               ::= [a-zA-Z0-9=_-]+
```
so must always include the algorithm.

where is the discrepancy?

snaftaly: First I'd mention that we are talking here about docker-archive and not oci-archive. In OCI archive we indeed get the image ID from manifest.config.digest, which includes the algorithm prefix.
As @agatakrajewska mentioned [here](https://snyk.slack.com/archives/C01EPSMGZML/p1618420140094600?thread_ts=1618396754.091000&cid=C01EPSMGZML):
'docker save does not save into manifest v2, schema2, they use something custom based on v1'.
Therefore, in the docker archive, we get the imageId from the manifest.json 'Config' attribute value, which is actually the path of the config.json file in the archive. The name of this file is the imageId (without the algo prefix). 

From what I understand from the links I sent above, the algorithm itself is sha256, although this information is not mentioned in the docker archive itself. 

I think it would be best to have @agatakrajewska 's input here 🙏

Shesekino: okay, makes sense :) feel free to resolve

agatakrajewska: Sharon, you nailed it! I think the only uncertainty (is that a word?) is are we sure that it's always sha256 algorythm? But based on the docs you found I think it's pretty safe to make that assumption.

agatakrajewska (approved):
Great PR and lovely to see more test coverage added with the change 🙏 
I would also make sure we don't break any e2e in registry, that we don't compare imageId's anywhere 🤔

snaftaly:
> Great PR and lovely to see more test coverage added with the change 🙏
> I would also make sure we don't break any e2e in registry, that we don't compare imageId's anywhere 🤔

Hey Agata, I made a local check with Registry+DRA+this change, and it looks like the import flow is ok, and it even doesn't cause the creation of a new monitor state when the imageId contains the prefix. 
I believe it is safe to merge.

PR #341, opened 2021-05-03
fix: ignore gradle/caches path when extracting jars

RotemS:
- [x] Ready for review

#### What does this PR do?

When extracting `jar` files, excluding any path that contains `gradle/caches`

Scan an archive that has a `jar` that's located in `gradle/caches` or any path that includes that.

`fixtures/docker-archives/docker-save/java.tar` was modified to contain such a path that's excluded.


shaimendel (approved):

lib/inputs/java/static.ts
```diff
@@ -2,13 +2,15 @@ import * as path from "path";
 import { ExtractAction } from "../../extractor/types";
 import { streamToSha1 } from "../../stream-utils";
 
-const ignoredPaths = [path.normalize("/usr/lib")];
+const ignoredPaths = ["/usr/lib", "gradle/caches"];
```
dkontorovskyy, line 5: why not just `gradle/cache`? It will also catch `gradle/caches`

RotemS: No good reason :) Thanks for flagging, changed it

dkontorovskyy (approved):
LGTM

shaimendel (approved):

PR #342, opened 2021-05-03
Run 1464 go file parse issue

chen-keinan:
- [x] Ready for review

#### What does this PR do?
this pr catch the exception from elfy lib and do no throw it again , just return empty object as same would happen if magic number do not match elf , there is no reason the throw an exception to break layer file iterator.
its either we determine that the file is valid elf(go binary) or not.

customer image archive indexing failed due to exception throw when has bad elf file while parsing binary file


agatakrajewska (approved):

agatakrajewska:
Great change, thanks for fixing! 🙏 @chen-keinan Could you please assign the PR to yourself, it makes it easier later when browsing through them 🙏

PR #353, opened 2021-06-29
fix: Handle container test with invalid image [CAP-228]

iuliams:
- [x] Ready for review
- [x] Tests Updated

#### What does this PR do?
This PR checks if the image argument passed by the user via the CLI is an existing image and throws an error if it could not be analysed.

From the cli:
$ node dist/cli container test /does-not-exist
$ node dist/cli container test does-not-exist

See [CAP-228 ](https://snyksec.atlassian.net/browse/CAP-228)

### DIFF
--- a/lib/analyzer/image-inspector.ts
+++ b/lib/analyzer/image-inspector.ts
@@ -89,6 +89,11 @@ async function pullWithDockerBinary(
     ) {
       throw err;
     }
+
+    if (err.stderr && err.stderr.includes("invalid reference format")) {
+      throw new Error(`invalid image format`);
+    }
+
     return pullAndSaveSuccessful;
   }
 }
--- a/test/lib/analyzer/image-inspector.spec.ts
+++ b/test/lib/analyzer/image-inspector.spec.ts
@@ -4,6 +4,7 @@ import * as tmp from "tmp";
 import { v4 as uuidv4 } from "uuid";
 
 import { DockerPullResult } from "@snyk/snyk-docker-pull";
+import * as plugin from "../../../lib";
 import * as imageInspector from "../../../lib/analyzer/image-inspector";
 import { ArchiveResult } from "../../../lib/analyzer/types";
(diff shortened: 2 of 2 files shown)


agranado2k (dismissed):
Great job! LGTM! 🎉

lib/analyzer/image-inspector.ts
```diff
@@ -89,6 +89,11 @@ async function pullWithDockerBinary(
     ) {
       throw err;
     }
+
+    if (err.stderr && err.stderr.includes("invalid reference format")) {
```
michael-c-snyk, line 93: Could this be `if (err.stderr?.includes......` instead?

test/lib/analyzer/image-inspector.spec.ts
```diff
@@ -85,6 +86,17 @@ describe("extractImageDetails", () => {
     expect(imageName).toEqual(expected.imageName);
     expect(tag).toEqual(expected.tag);
   });
+  it("test unprovided image", async () => {
```
michael-c-snyk, line 89: Looks like the pattern for test naming begins "should eg line 112", wdyt about aligning these?

michael-c-snyk (commented):
Nice one, just added a couple of questions

test/lib/analyzer/image-inspector.spec.ts
iuliams: Aligned, however note that in the other tests "should produce the expected state" is not a very clear message. Could change it here if you find it like a good fit.

lib/analyzer/image-inspector.ts
iuliams: As it is now, it's consistent with the rest of the error checks I'd say... Wdyt?

michael-c-snyk: No strong feeling :)

michael-c-snyk (approved):

PR #355, opened 2021-07-07
fix: split auto detected user instructions and dockerfile analysis type

ahmed-agabani-snyk:
- [ ] Ready for review

#### What does this PR do?

Updates domain to reflect that AutoDetectedUserInstructions does not have a base image.
This is preparation for adding parsing errors to DockerFileAnalysis.

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -1,6 +1,10 @@
 import { DepGraphData } from "@snyk/dep-graph";
 
-import { DockerFileAnalysis, DockerFilePackages } from "./dockerfile/types";
+import {
+  DockerFileAnalysis,
+  DockerFileLayers,
+  DockerFilePackages,
+} from "./dockerfile/types";
 
 export enum ImageType {
   Identifier, // e.g. "nginx:latest"
@@ -71,6 +75,11 @@ export interface ScanResult {
   identity: Identity;
   /** Facts are the collection of things you found. */
   facts: Fact[];
+}
+
+export interface AutoDetectedUserInstructions {
+  dockerfilePackages: DockerFilePackages;
+  dockerfileLayers: DockerFileLayers;
 }
 
 export interface ContainerTarget {
--- a/lib/analyzer/types.ts
+++ b/lib/analyzer/types.ts
@@ -1,5 +1,4 @@
-import { DockerFileAnalysis } from "../dockerfile";
-import { ManifestFile } from "../types";
+import { AutoDetectedUserInstructions, ManifestFile } from "../types";
 import { AppDepsScanResultWithoutTarget } from "./applications/types";
 
 export interface AnalyzedPackage {
@@ -74,7 +73,7 @@ export interface StaticAnalysis {
   binaries: string[];
   imageLayers: string[];
   rootFsLayers?: string[];
-  autoDetectedUserInstructions?: DockerFileAnalysis;
+  autoDetectedUserInstructions?: AutoDetectedUserInstructions;
   applicationDependenciesScanResults: AppDepsScanResultWithoutTarget[];
   manifestFiles: ManifestFile[];
   imageLabels?: { [key: string]: string };
--- a/lib/extractor/docker-archive/index.ts
+++ b/lib/extractor/docker-archive/index.ts
@@ -3,7 +3,7 @@ import {
   getLayersFromPackages,
   getPackagesFromRunInstructions,
 } from "../../dockerfile/instruction-parser";
-import { HashAlgorithm } from "../../types";
+import { AutoDetectedUserInstructions, HashAlgorithm } from "../../types";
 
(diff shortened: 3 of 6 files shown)


lib/types.ts
```diff
@@ -73,6 +77,11 @@ export interface ScanResult {
   facts: Fact[];
 }
 
+export interface AutoDetectedUserInstructions {
```
Shesekino, line 80: Can we not call it "user instructions" please? I get that's how it's already called but I don't want us doubling down on it. this term doesn't exist

ahmed-agabani-snyk: Changing `AutoDetectedUserInstructions` -> `UserInstructions` seems ok in terms of domain.

Given that the interface would be no longer tied to the feature, I think it would make sense to change 
```typescript
export interface DockerFileAnalysis {
  baseImage?: string;
  dockerfilePackages: DockerFilePackages;
  dockerfileLayers: DockerFileLayers;
}
```
to
```typescript
import { UserInstructions } from "../types";

export interface DockerFileAnalysis extends UserInstructions {
  baseImage?: string;
}
```
because Dockerfile conceptually contains user instructions. WDYT @Shesekino?

Shesekino: "user instructions" is a term we made up, that's not a thing.
these are instructions, or commands, that were used to build the image

Shesekino: https://github.com/opencontainers/image-spec/blob/master/config.md
the spec refers to those as ```created_by```

```
created_by string, OPTIONAL

The command which created the layer.
```

Shesekino: (essentially I prefer something that doesn't include ```user``` in it)

ahmed-agabani-snyk: Trying to think of a name that represents this collection of data. [reference](https://github.com/snyk/snyk-docker-plugin/blob/master/test/lib/dockerfile.test.ts#L138-L153)

```typescript
        dockerfilePackages: {
          bash: {
            instruction:
              "RUN apk update && apk upgrade && apk add --update --no-cache nodejs bash",
          },
          nodejs: {
            instruction:
              "RUN apk update && apk upgrade && apk add --update --no-cache nodejs bash",
          },
        },
        dockerfileLayers: {
          UlVOIGFwayB1cGRhdGUgJiYgYXBrIHVwZ3JhZGUgJiYgYXBrIGFkZCAtLXVwZGF0ZSAtLW5vLWNhY2hlIG5vZGVqcyBiYXNo: {
            instruction:
              "RUN apk update && apk upgrade && apk add --update --no-cache nodejs bash",
          },
        },
```

Train of thought, this structure represents how dependencies were brought into the:
* `autoDetectedUserInstructions` case:  container image's last x hours of layer(s)
* `dockerfileAnalysis` case: dockerfile's current set of instructions

The dependencies introduced by both of these cases are driven exclusively by the end user, explicitly excluding the instructions inherited from the base image the end user is using.

The long name I would give this is "how dependencies where introduced into the layers created by the end user".

What would be the short name?

ahmed-agabani-snyk: This structure represents processed user/end user instructions where the processed results includes which instructions introduced each dependency.

I honestly think `UserInstructions` or `ProcessedUserInstructions` would be appropriate names.

As you pointed out, AutoDetected should be removed from the name because the method of producing the user instructions should not be captured in the interface.

Shesekino: I don't understand who is this user. These are the commands that created the layer

Shesekino: some reference https://snyk.slack.com/archives/C011VKVL015/p1614009101001800?thread_ts=1614008803.001500&cid=C011VKVL015
https://snyk.slack.com/archives/C01EPSKVC9L/p1613997226010300?thread_ts=1613992505.006200&cid=C01EPSKVC9L

Shesekino: I suggest someone from Capsule take over for finding the terminology we want to use - it would be their responsibility to pay back that tech debt eventually (or handle the consequences of not)

ahmed-agabani-snyk: > I don't understand who is this user. These are the commands that created the layer

I've believe i've just understood your viewpoint. These are the Dockerfile instructions or approximated Dockerfile used to build the top most layer.

Shesekino: Correct, for the most part, I think :)
Are they necessarily coming from a Dockerfile, or from the image's metadata? (didn't look at the code)
Reason I'm asking: Container images may be built without a Dockerfile

ahmed-agabani-snyk: They are coming from both depending on the flow that is being used.

* For auto detected user instructions flow, they come from the latest 5 hours worth of `history.created_by` https://github.com/opencontainers/image-spec/blob/master/config.md)
* For the dockerfile flow, they come from the run statements of the dockerfile the user provided

michael-c-snyk (approved):

PR #361, opened 2021-08-12
feat: allow base image to be partially missing

michael-c-snyk:
- [ ] Ready for review

#### What does this PR do?
Allow base image to be partially missing

In some cases where the base image in the Dockerfile's FROM instruction contains arguments that are substituted at build-time, we would incorrectly detect the base image as /image:tag which is a bad user experience. We fix this by ensuring that we report the base image only if we can read every variable in the FROM instruction, while keeping the old handling for package instruction parsing.

Co-authored-by:@ahmed-agabani-snyk


ahmed-agabani-snyk (approved):

agranado2k (approved):
LGTM! 🎉

PR #362, opened 2021-08-16
refactor: jar extract action to return a Buffer

shaninja:
- [x] Ready for review

#### What does this PR do?
Return a Buffer rather than sha from the extract actions, and only convert to sha during later processing.

This is required before will are able to extend the later processing with unpacking the jar within the next task

can be reviewed commit-by-commit. main changes are in first commit, and later commits are more refactoring and tests.

run a scan on image with jars, note that output should be the same on master and on this branch.

Part of "UberJars" work to support nested jars.

CAP-409

### DIFF
--- a/lib/inputs/index.ts
+++ b/lib/inputs/index.ts
@@ -1,4 +1,5 @@
 import {
+  FilePathToBuffer,
   FilePathToContent,
   FilePathToElfContent,
 } from "../analyzer/applications/types";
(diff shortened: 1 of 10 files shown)


lib/inputs/index.ts
```diff
@@ -17,7 +21,7 @@ export function getFileContent(
       if (actionName !== searchedAction) {
         continue;
       }
-      if (!(typeof extractedLayers[filePath][actionName] === "string")) {
+      if (!isStringType) {
```
agranado2k: I curious about how this `isStringType` works. Does it implicit pass the `actionName` const? Do you think it can raise some doubts for someone who will read the code in the future, like me? :)

test/lib/buffer-utils.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { Buffer } from "buffer";
+import * as crypto from "crypto";
+import { bufferToSha1 } from "../../lib/buffer-utils";
+
+describe("buffer-utils", () => {
+  describe("bufferToSha1", () => {
```
agranado2k, line 6: I guess this test i related with the previous commit, right? Do you think it will be better merge both commits in one in this case?

agranado2k (dismissed):
LGTM! Great job! 🎉

lib/inputs/index.ts
shaninja: :O I don't think it did work! but probably I haven't noticed because it was fixed immediately in a later commit. let me check if any tests have failed over that (I might have not noticed due to the jest tests not running on `npm run test`, to make sure that this part is covered well, and then I'll fix this commit.

shaninja: https://github.com/snyk/snyk-docker-plugin/pull/362/commits/2e777496cb0f3126ac283599264984377c456abe

^^ this new commit fixes this issue, and adds test to lib/inputs/index.ts (which i tested and they failed prior to the change)

agranado2k (approved):
LGTM! 🎉 
Great job! :)

PR #367, opened 2021-08-25
Feat/unpacking jar from buffer

shaninja:
- [ ] Ready for review

#### What does this PR do?
this is code that unpacks _the first level_ of jars, and checks if there are jars within them.
if there are jars within the jar (aka "fat-jar") we only return the children jars to be checked for vulns. 

This is not ready to go out yet!! we want to hide it behind a CLI flag.
also, in a soon but later iteration, we're working on performance improvements.

we'd appreciate your early feedback while we work on adding the CLI flag.


Shesekino:
image

lib/types.ts
```diff
@@ -163,6 +163,16 @@ export interface PluginOptions {
 
   /** Whether to enable application dependencies scanning. The default is "false" */
   "app-vulns": boolean | string;
+
+  /**
+   * Whether to unpack jars and look for jars within them.
+   * If jar contains other jars (AKA fat-jar or shaded-jar), we send back only the children jars, and don't look for vulns in the parent.
+   * Currently only one level deep.
+   * Must always come with app-vulns
```
Shesekino, line 171: Is this enforced somehow?

shaninja: hmm, only in the sense that you wont even get to the desired code if you don't have the app-vulns flag:
https://github.com/snyk/snyk-docker-plugin/pull/367/files#diff-ef7473bc4617398b357238645eb8280929cc7617ebf63ad15e8fb1ea8d82cf34L149
https://github.com/snyk/snyk-docker-plugin/pull/367/files#diff-ef7473bc4617398b357238645eb8280929cc7617ebf63ad15e8fb1ea8d82cf34L79

do you believe it's not enough?

also made sure it's covered in a test:
https://github.com/snyk/snyk-docker-plugin/pull/367/files#diff-b0203a8cdef34bc7279d600248ec84804e0b90001ca61141bb95f2bb4735bb51R106

Shesekino: My point is having as few surprises as possible for whoever consumes this library.
If I provide the new flag, but not appVulns, and I see nothing, I'm wondering what's going on.
If I provide the new flag, but not appVulns, and I see an error, I understand I've done something wrong

shaninja: 👍 gotcha

shaninja: added a commit https://github.com/snyk/snyk-docker-plugin/pull/367/commits/32be93a3500ee794c3d4ad5124ec18867f62eb82

I will squash all commits before merging. also will get @garethr 's feedback on the error msg.

PR #370, opened 2021-09-12
test: falsely identify deleted files

Shesekino:
- [x] Ready for review

#### What does this PR do?

This commit adds a test that asserts how our scanner behaves when files have been "deleted" from a layer.

It adds an archive fixture built from the Dockerfile:
```
FROM busybox
COPY node /
RUN rm /node
```
(saved as an OCI archive)

This archive is then scanned, and even thogh the node "binary" (actually an empty file) was deleted from the image, we're still reporting its SHA, potentially reporting false positives.


shaninja (approved):
I also documented this issue in JIRA for the team to prioritize.

thank you 🙇

PR #375, opened 2021-09-28
fix: Release analyzer should return the right code

mladkau:
- [x] Ready for review

#### What does this PR do?
Fixes the detection of RedHat OS in RedHat 6 images.

Yes

https://snyk.zendesk.com/agent/tickets/14148

### DIFF
--- a/lib/analyzer/os-release/release-analyzer.ts
+++ b/lib/analyzer/os-release/release-analyzer.ts
@@ -76,7 +76,10 @@ export async function tryRedHatRelease(
   if (!idRes || !versionRes) {
     throw new Error("Failed to parse /etc/redhat-release");
   }
-  const name = idRes[1].replace(/"/g, "").toLowerCase();
+  const name =
+    text.indexOf("Red Hat") !== -1
+      ? "rhel"
+      : idRes[1].replace(/"/g, "").toLowerCase();
   const version = versionRes[1].replace(/"/g, "");
   return { name, version, prettyName: "" };
 }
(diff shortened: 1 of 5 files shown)


test/lib/analyzer/os-release-detector.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+    }
+  }
+
+  return fac;
+};
+
+describe("os release parsing", () => {
+  it("should parse os release files correctly", async () => {
+    const examples = {
+      "alpine:2.6": {
```
mladkau, line 43: I salvaged these from the old `test/lib/analyzer/os-release-detector.test.ts` file.

lib/analyzer/os-release/release-analyzer.ts
```diff
@@ -76,7 +76,10 @@ export async function tryRedHatRelease(
   if (!idRes || !versionRes) {
     throw new Error("Failed to parse /etc/redhat-release");
   }
-  const name = idRes[1].replace(/"/g, "").toLowerCase();
+  const name =
+    text.indexOf("Red Hat") !== -1
```
mladkau, line 80: Looking at the present parsing logic I cannot see how this could ever have worked for RedHat before version 7 (I think the parsing logic might have only been tested on CentOS). Since we need the value `rhel` for the os name when we have an actual RedHat image I couldn't think a better fix.

agranado2k (approved):
LGTM!

lib/analyzer/os-release/release-analyzer.ts
agatakrajewska: I think this is right, I also can't think of a better way. Maybe they've also changed the older images like centos? Who knows...

agatakrajewska (approved):

PR #377, opened 2021-10-19
fix: Swap async to sync tmp dir removal

agatakrajewska:
- [x] Ready for review

#### What does this PR do?
We're getting intermittent scan failures, where we can't remove the tmp folder cause it has not yet been created. This change might help solving the problem.
Also taking this opportunity to simplify the code, new versions of node now support removal of the whole directory with contents recursively and `force` option will not throw when dir does not exists.

### DIFF
--- a/lib/analyzer/image-inspector.ts
+++ b/lib/analyzer/image-inspector.ts
@@ -25,8 +25,14 @@ async function getInspectResult(
 
 function cleanupCallback(imageFolderPath: string, imageName: string) {
   return () => {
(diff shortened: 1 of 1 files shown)


shaninja (commented):
🤞 that it would fix the issue (I don't have insight into that, not sure). but nice change regardless, much cleaner.

shaninja (dismissed):

dkontorovskyy (changes requested):
Blocking for a reason that rmdirSync is not supported in Node 10

PR #381, opened 2021-12-01
feat: Add image creation time fact [MAGMA-1271]

snaftaly:
- [X] Ready for review

#### What does this PR do?

This PR adds image creation time fact.

I'd suggest to start in extractImageContent function

As part of the rule-based auto-import from container registries, we would like to base some of the import rules on the image creation time, found in the image config. 
Since the image creation time is not a required field in the [spec](https://github.com/opencontainers/image-spec/blob/main/config.md#properties), we would like to examine what is the
percentage of images that have this information missing. Adding this attribute as a fact will allow us to collect this information.

https://snyksec.atlassian.net/browse/MAGMA-1271

### DIFF
--- a/lib/response-builder.ts
+++ b/lib/response-builder.ts
@@ -77,6 +77,14 @@ async function buildResponse(
     additionalOsDepsFacts.push(imageLabels);
   }
 
+  if (depsAnalysis.imageCreationTime) {
+    const imageCreationTimeFact: facts.ImageCreationTimeFact = {
+      type: "imageCreationTime",
+      data: depsAnalysis.imageCreationTime,
+    };
+    additionalOsDepsFacts.push(imageCreationTimeFact);
+  }
+
   if (
     depsAnalysis.rootFsLayers &&
     Array.isArray(depsAnalysis.rootFsLayers) &&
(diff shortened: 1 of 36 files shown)


lib/response-builder.ts
```diff
@@ -77,6 +77,14 @@ async function buildResponse(
     additionalOsDepsFacts.push(imageLabels);
   }
 
+  if (depsAnalysis.imageCreationTime) {
+    const imageCreationTimeFact: facts.ImageCreationTimeFact = {
+      type: "imageCreationTime",
+      data: depsAnalysis.imageCreationTime,
+    };
+    additionalOsDepsFacts.push(imageCreationTimeFact);
```
snaftaly, line 85: Does it make sense to add it to this array of facts (it's not really an os-deps fact, but I see that almost all facts are added to this array). Or maybe this array can be renamed to additionalFacts?

agatakrajewska: Yes, it looks like this array has been abused 😆 If you fancy renaming that would be awesome 🙇‍♀️

agatakrajewska (approved):

PR #386, opened 2021-12-16
feat: add alias nested-jars-depth in addition to shaded-jars-depth

shaninja:
we realized that "shaded-jars" to speak about "nested jars" (jars inside jar)  is not the correct term.

We don't want to break backwards compatibility, so adding the alias. in the future we will only document
--nested-jars-depth, and will ask those who use --shaded to switch.

> **uber jar** is also known as **fat jar** i.e. **jar with dependencies.**There are three common methods for constructing an uber jar:
>

> **Unshaded:** Unpack all JAR files, then repack them into a single JAR. Works with Java's default class loader. Tools [maven-assembly-plugin](http://maven.apache.org/plugins/maven-assembly-plugin/)
>

> **Shaded:** Same as unshaded, but rename (i.e., "shade") all **packages** of all dependencies. Works with Java's default class loader. Avoids some (not all) dependency version clashes. Tools [maven-shade-plugin](http://maven.apache.org/plugins/maven-shade-plugin/)
>

> **JAR of JARs:** The final JAR file contains the other JAR files embedded within. Avoids dependency version clashes. All resource files are preserved. Tools: [Eclipse JAR File Exporter](http://help.eclipse.org/luna/index.jsp?topic=%2Forg.eclipse.jdt.doc.user%2Freference%2Fref-export-jar.htm)
>

- [x] Ready for review

### DIFF
--- a/test/system/application-scans/java.spec.ts
+++ b/test/system/application-scans/java.spec.ts
@@ -28,142 +28,15 @@ describe("jar binaries scanning", () => {
       digest: expect.any(String),
     };
 
-    describe("with all needed CLI flags (app-vulns and shaded-jars-depth)", () => {
-      beforeAll(async () => {
-        // Arrange
-        fixturePath = getFixture(
-          "docker-archives/docker-save/java-uberjar.tar",
-        );
-        const imageNameAndTag = `docker-archive:${fixturePath}`;
-
-        // Act
-        pluginResult = await scan({
-          path: imageNameAndTag,
-          "app-vulns": true,
-          "shaded-jars-depth": "1",
-        });
-
-        fingerprints = pluginResult.scanResults[1].facts[0].data.fingerprints;
-      });
-
-      it("should return two results", async () => {
-        expect(fingerprints).toHaveLength(2);
-      });
-
-      it("should return nested (second-level) jar in the result", async () => {
-        expect(fingerprints).toContainEqual(expect.objectContaining(nestedJar));
-      });
-
-      it("should not return a first-level jar that have nested jars in it (uber jar)", async () => {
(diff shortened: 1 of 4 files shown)


test/system/application-scans/java.spec.ts
```diff
@@ -28,142 +28,368 @@ describe("jar binaries scanning", () => {
...
 
-    describe("with all needed CLI flags (app-vulns and shaded-jars-depth)", () => {
-      beforeAll(async () => {
-        // Arrange
-        fixturePath = getFixture(
-          "docker-archives/docker-save/java-uberjar.tar",
-        );
-        const imageNameAndTag = `docker-archive:${fixturePath}`;
+    // TODO: deprecate --shaded-jars-depth and leave only --nested-jars-depth
+    describe("--shaded-jars-depth", () => {
```
shaninja, line 32: everything is nested behind this new describe. "Hide whitespace" for a better view.

test/system/application-scans/java.spec.ts
```diff
@@ -28,142 +28,368 @@ describe("jar binaries scanning", () => {
...
+
+        expect(fingerprints).toHaveLength(1);
+        expect(fingerprints).toContainEqual(
+          expect.objectContaining(level2JarFingerprint),
+        );
         expect(fingerprints).not.toContainEqual(
-          expect.objectContaining(nestedJar),
+          expect.objectContaining(deepestLevelJarFingerprint),
         );
       });
```
shaninja, line 384: This is a new test. all tests below it are duplicates of --shaded-jars-depth.

iuliams (approved):

PR #390, opened 2021-12-20
fix: bring back option for nested-jars-depth=0

shaninja:
- [x] Ready for review

#### What does this PR do?

Allowing the CLI users to opt-out and use `--nested-jars-depth=0` to not unpack any levels of JARs.

### DIFF
--- a/test/system/application-scans/java.spec.ts
+++ b/test/system/application-scans/java.spec.ts
@@ -121,15 +121,20 @@ describe("jar binaries scanning", () => {
         );
         const imageNameAndTag = `docker-archive:${fixturePath}`;
 
-        it("should throw if shaded-jars-depth flag is set to 0", async () => {
-          // Act + Assert
-          await expect(
-            scan({
-              path: imageNameAndTag,
-              "app-vulns": true,
-              "shaded-jars-depth": "0",
-            }),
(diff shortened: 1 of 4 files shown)


test/system/application-scans/java.spec.ts
```diff
@@ -482,14 +487,19 @@ describe("jar binaries scanning", () => {
         const imageNameAndTag = `docker-archive:${fixturePath}`;
 
         it("should not unpack jars if nested-jars-depth flag is set to 0", async () => {
```
shaninja, line 489: I wonder if there's a way with `it.each` that I can run all the tests once with --nested and once with --shaded, would it be easier to maintain, and how hard it would be to break apart once we get rid of --shaded 🤔

michael-c-snyk (approved):

test/system/application-scans/java.spec.ts
shaninja: nope, can't really. you cannot nest `describe` or even `test` inside an it.each, so this structure wont work.

shaninja: https://github.com/snyk/snyk-docker-plugin/pull/392

PR #398, opened 2022-01-18
feat: scan image config for oci archives to get rootFs, platform, etc.

ivanstanev:
- [x] Ready for review

#### What does this PR do?

The OCI Archive scanning was not collecting the image config, so we were missing lots of good information from the image, such as root fs layers, auto-detected user instructions, platform, and image labels. The config is now properly detected and we can read this data.

### DIFF
--- a/lib/extractor/oci-archive/layer.ts
+++ b/lib/extractor/oci-archive/layer.ts
@@ -7,9 +7,9 @@ import { extract, Extract } from "tar-stream";
 import { streamToJson } from "../../stream-utils";
 import { extractImageLayer } from "../layer";
 import {
-  DockerArchiveImageConfig,
   ExtractAction,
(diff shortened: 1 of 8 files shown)


lib/extractor/oci-archive/layer.ts
```diff
@@ -28,12 +29,14 @@ export async function extractArchive(
 ): Promise<{
   layers: ExtractedLayers[];
   manifest: OciArchiveManifest;
+  imageConfig: DockerArchiveImageConfig;
```
agatakrajewska, line 32: Could we maybe refactor and rename that interface just to eg `ImageConfig` if we are now using it for the OCI archive? 🙇‍♀️

agatakrajewska (approved):

PR #400, opened 2022-01-18
fix: allow oci-archives with empty layers to be scanned

ivanstanev:
- [x] Ready for review

#### What does this PR do?

Images like busybox, where we cannot detect any facts, were not able to be scanned until now.
This change ensures that we can do this.


RotemS (approved):
LGTM 🚀

PR #402, opened 2022-01-21
Feat/memory improvements

ivanstanev:
- [x] Ready for review

#### What does this PR do?

- optimize memory allocations in streamToString()
- optimize memory allocations in streamToJson()

### DIFF
--- a/lib/stream-utils.ts
+++ b/lib/stream-utils.ts
@@ -15,14 +15,14 @@ export async function streamToString(
   streamSize?: number,
   encoding: SupportedEncodings = "utf8",
 ): Promise<string> {
-  const chunks: string[] = [];
+  const chunks: Buffer[] = [];
   return new Promise((resolve, reject) => {
     stream.on("end", () => {
-      resolve(chunks.join(""));
+      resolve(Buffer.concat(chunks).toString(encoding));
     });
     stream.on("error", (error) => reject(error));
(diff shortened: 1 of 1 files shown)


lib/stream-utils.ts
```diff
@@ -15,14 +15,14 @@ export async function streamToString(
...
+  const chunks: Buffer[] = [];
   return new Promise((resolve, reject) => {
     stream.on("end", () => {
-      resolve(chunks.join(""));
+      resolve(Buffer.concat(chunks).toString(encoding));
     });
     stream.on("error", (error) => reject(error));
-    stream.on("data", (chunk) => {
-      chunks.push(chunk.toString(encoding));
+    stream.on("data", (chunk: Buffer) => {
```
agatakrajewska, line 24: Oh nice! How do you know this change improves memory allocation?

agatakrajewska (approved):

lib/stream-utils.ts
ivanstanev: Haven't had the chance to go back to this code yet, but once I do I will try to collect some performance data and share it here 😄 

I noticed that in https://github.com/snyk/kubernetes-monitor a huge amount of allocations were done in this function. Looking at some code on various websites I notice that this was the suggested way to collect data and to then combine it to a single string. I'll verify if it indeed improves the allocations!

PR #403, opened 2022-01-24
fix: remove unnecessary async keyword from extractImageDetails

SteveShani:
- [x] Ready for review

#### What does this PR do?

Removes the async keywork from the function extractImageDetails which doesn't use await inside it.


RotemS (dismissed):
LGTM 💯

Shesekino:
Nice find!
Worth pointing out that it's more of a `fix: ` than a `chore: ` since it changes actual behaviour (not just renaming etc.)

shaninja (approved):

RotemS (approved):

PR #407, opened 2022-02-01
fix: Fixed invalid jar fingerprints for maven projects

pecodez:
- [ ] Ready for review

#### What does this PR do?
Refactored the recursion of unpacking nested JARs to avoid mutating objects
which was difficult to follow and caused incorrect dependency allocation
in the fingerprints facts which resulted in an invalid dep-graph.

### DIFF
--- a/lib/analyzer/applications/java.ts
+++ b/lib/analyzer/applications/java.ts
@@ -3,7 +3,7 @@ import * as path from "path";
 import { bufferToSha1 } from "../../buffer-utils";
 import { JarFingerprintsFact } from "../../facts";
 import { JarFingerprint } from "../types";
-import { JarBuffer, JarCoords } from "./types";
+import { JarBuffer, JarCoords, JarInfo } from "./types";
 import { AppDepsScanResultWithoutTarget, FilePathToBuffer } from "./types";
 
 function groupJarFingerprintsByPath(input: {
@@ -12,12 +12,13 @@ function groupJarFingerprintsByPath(input: {
   [path: string]: JarBuffer[];
 } {
   const jarFingerprints: JarBuffer[] = Object.entries(input).map(
-    ([filePath, digest]) => {
+    ([filePath, buffer]) => {
       return {
         location: filePath,
-        digest,
+        buffer,
         coords: null,
         dependencies: [],
+        nestedJars: [],
       };
     },
   );
@@ -77,11 +78,14 @@ function getFingerprints(
   desiredLevelsOfUnpacking: number,
   jarBuffers: JarBuffer[],
 ): JarFingerprint[] {
-  return unpackJars(jarBuffers, desiredLevelsOfUnpacking);
+  const fingerprints = new Set([
+    ...unpackJars(jarBuffers, desiredLevelsOfUnpacking),
+  ]);
+  return Array.from(fingerprints);
 }
 
 /**
- * Recursively unpacks JARs and attempts to add coords to the
+ * Unpacks a JAR and attempts to add coords to the
  * package and it's dependencies.
  *
  * JARs will always be unpacked one level more than requested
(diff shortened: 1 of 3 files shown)


lib/analyzer/applications/java.ts
```diff
@@ -172,39 +169,39 @@ function unpackJarsTraverse({
...
 }
 
+/**
+ * Manages the unpacking an array of JarBuffer objects and returns the resulting
+ * fingerprints. Recursion to required depth is handled here when the returned
+ * info from each JAR that is unpacked has nestedJars.
+ *
+ * @param {JarBuffer[]} jarBuffers
+ * @param {number} desiredLevelsOfUnpacking
+ * @param {number} unpackedLevels
```
noaleibo1, line 198: Curious about what this annotation means. Where is this loaded from? Some configuration file?

noaleibo1 (commented):
Understood very little of this PR obviously, and even the part I did understand I am not sure about  😅 
Are we basically limiting the number of levels we dive into, to avoid the "overflow" error? If so, to what number of levels? And where is it configured?

lib/analyzer/applications/java.ts
pecodez: Yes, we do apply a limit to the level of unpacking. By default, this is 1 level but controlled with the `--nested-jar-depth` flag to allow the customer control. We actually always unpack 1 level deeper than the `desiredLevelOfUnpacking` as that is how we find shaded JARs and coords for custom JARs. We only generate fingerprints for JARs found up to the `desiredLevelOfUnpacking` though.

lib/analyzer/applications/java.ts
```diff
@@ -218,29 +215,41 @@ function unpackJars(
...
+      dependencies: jarInfo.dependencies,
+      ...jarInfo.coords,
     });
+
+    if (jarInfo.nestedJars.length > 0) {
+      // this is an uber/fat JAR so we need to unpack the nested JARs to
+      // analyse them for coords and further nested JARs (depth flag allowing)
+      fingerprints.push(
+        ...unpackJars(
+          jarInfo.nestedJars as JarBuffer[],
```
shaninja, line 247: does it really have all the properties of JarBuffer, or is it a partial? would it make more sense to change line 202 to `Partial<JarBuffers>[]` (if that's a thing...)? not sure I fully follow and understand the source of the type difference...

test/system/application-scans/java.spec.ts
```diff
@@ -56,16 +56,22 @@ describe("jar binaries scanning", () => {
               pluginResult.scanResults[1].facts[0].data.fingerprints;
           });
 
-          it("should return two results", () => {
-            expect(fingerprints).toHaveLength(2);
+          it("should return three results", () => {
```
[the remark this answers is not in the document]
pecodez: I agree, it isn't clear. I will update the test name to be clearer

lib/analyzer/applications/java.ts
pecodez: This was super unclear too. I've reduced the scope of the `JarBuffer` type and extended the `JarInfo` type. Now `unpackJars` method takes a `JarBuffer[]` and `unpackJar` method returns `JarInfo` which has the decorated information needed to create a `Fingerprint` type.

test/system/application-scans/java.spec.ts
```diff
@@ -251,7 +257,7 @@ describe("jar binaries scanning", () => {
             fingerprints =
               pluginResult.scanResults[1].facts[0].data.fingerprints;
 
-            expect(fingerprints).toHaveLength(1);
+            expect(fingerprints).toHaveLength(3);
```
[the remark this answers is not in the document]
pecodez: The assertions have not actually changed; they both previously expected 1 fingerprint matching the deepest level. I believe this test case was more about checking things not breaking when specifying a depth greater than the actual levels of nested JARs and so have renamed the test to be specific.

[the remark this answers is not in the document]
shaninja: ah, true. I think I got confused when reading the previous assertion.

shaninja (approved):

noaleibo1 (approved):

PR #417, opened 2022-05-02
php app vulns [MYC-82]

yaronschwimmer:
- [x] Ready for review

#### What does this PR do?
Adds support for scanning `composer` (PHP) manifest files directly in the plugin.

Using the CLI / DRA, run a scan on an app+os image containing PHP manifest files. Both should result with the same vulns for both app and OS.

[MYC-82](https://snyksec.atlassian.net/browse/MYC-82)

### DIFF
--- a/lib/analyzer/applications/php.ts
+++ b/lib/analyzer/applications/php.ts
@@ -1,0 +1,99 @@
+import { legacy } from "@snyk/dep-graph";
+import * as path from "path";
+import * as lockFileParser from "@snyk/composer-lockfile-parser";
+import { DepGraphFact, TestedFilesFact } from "../../facts";
+
+import { AppDepsScanResultWithoutTarget, FilePathToContent } from "./types";
+import { DepTreeDep } from "../../types";
+
+interface ManifestLockPathPair {
+  manifest: string;
+  lock: string;
+}
+const PACKAGE_MANAGER_TYPE = "composer";
+
+export async function phpFilesToScannedProjects(
+  filePathToContent: FilePathToContent,
+): Promise<AppDepsScanResultWithoutTarget[]> {
+  const scanResults: AppDepsScanResultWithoutTarget[] = [];
+
+  const filePairs = findManifestLockPairsInSameDirectory(filePathToContent);
+
+  const shouldIncludeDevDependencies = false;
+
+  for (const pathPair of filePairs) {
+    const parserResult = await lockFileParser.buildDepTree(
+      filePathToContent[pathPair.lock],
+      filePathToContent[pathPair.manifest],
+      pathPair.manifest,
+      {},
+      shouldIncludeDevDependencies,
+    );
+
+    const depGraph = await legacy.depTreeToGraph(
+      parserResult as DepTreeDep,
+      PACKAGE_MANAGER_TYPE,
+    );
+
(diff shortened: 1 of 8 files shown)


lib/analyzer/applications/php.ts
```diff
@@ -0,0 +1,99 @@
...
+  for (const pathPair of filePairs) {
+    const parserResult = await lockFileParser.buildDepTree(
+      filePathToContent[pathPair.lock],
+      filePathToContent[pathPair.manifest],
+      pathPair.manifest,
+      {},
+      shouldIncludeDevDependencies,
+    );
+
+    const depGraph = await legacy.depTreeToGraph(
```
RotemS, line 33: Do we have a type for this?

yaronschwimmer: Yes. It's `DepGraph` from `@snyk/dep-graph`

yaronschwimmer: Do you feel it's missing an explicit type declaration?

lib/analyzer/applications/php.ts
```diff
@@ -0,0 +1,99 @@
...
+import { DepGraphFact, TestedFilesFact } from "../../facts";
+
+import { DepTreeDep } from "../../types";
+import { AppDepsScanResultWithoutTarget, FilePathToContent } from "./types";
+
+interface ManifestLockPathPair {
+  manifest: string;
+  lock: string;
+}
+const PACKAGE_MANAGER_TYPE = "composer";
```
noaleibo1, line 13: Is this how we usually add constants? Would it be an overkill to have it as an environment variable? Or maybe as part of some config/consts file?

lib/analyzer/applications/php.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export async function phpFilesToScannedProjects(
+  filePathToContent: FilePathToContent,
+): Promise<AppDepsScanResultWithoutTarget[]> {
+  const scanResults: AppDepsScanResultWithoutTarget[] = [];
+
+  const filePairs = findManifestLockPairsInSameDirectory(filePathToContent);
+
+  const shouldIncludeDevDependencies = false;
```
noaleibo1, line 22: Same question as above - would be want to have this defined somewhere where it'll be easier to find/change?
I hope it doesn't sound like I'm trying to be annoying :)

lib/analyzer/applications/php.ts
yaronschwimmer: This is mostly to avoid putting this string directly when it's used in a function call. It's not expected to change, but I believe this way is a bit nicer and easier to read or extend in the future. It will be definitely be an overkill to add as an environment variable 🙂

lib/analyzer/applications/php.ts
yaronschwimmer: I don't expect it to change often if at all, so I think this way is easier to maintain than an env variable.

yaronschwimmer: also this file was heavily, um, inspired by the [node](https://github.com/snyk/snyk-docker-plugin/blob/main/lib/analyzer/applications/node.ts#L31) implementation

noaleibo1 (approved):

PR #422, opened 2022-05-26
feat: detect jar files in war files [MYC-109]

yaronschwimmer:
- [X] Ready for review

#### What does this PR do?
We want to be able to detect and scan WAR files at the top level, and
recursively unpack and detect JAR files in them, the same way we do with
top-level JAR files.

[MYC-109](https://snyksec.atlassian.net/browse/MYC-109)


lib/inputs/java/static.ts
```diff
@@ -3,11 +3,13 @@ import { ExtractAction } from "../../extractor/types";
 import { streamToBuffer } from "../../stream-utils";
 
 const ignoredPaths = ["/usr/lib", "gradle/cache"];
+const javaArchiveFileFormats = [".jar", ".war"];
 
 function filePathMatches(filePath: string): boolean {
   const dirName = path.dirname(filePath);
+  const fileExtension = filePath.slice(-4);
```
noaleibo1, line 10: should we consider breaking the the file path by ".", and looking at the last element of the array? I might be misunderstanding, but what about the option that the file ends with something like ".warwar"? I know this might actually be an option, but maybe it'd be better to avoid such case?

yaronschwimmer: splitting by "." is a valid option as well, but ".warwar" is not something we want to scan, and will not pass the condition in line 12

noaleibo1: You are right, and anyways I don't feel strongly about this.

noaleibo1 (approved):

PR #424, opened 2022-06-01
feat: support for image with RPM>=4.16

RotemS:
- [x] Ready for review

#### What does this PR do?

As of rpm>=4.16 the installed packages are now stored in an SQLite DB instead of a Berkeley DB.
This adds support for these images to return the correct list of the image's dependencies.

### DIFF
--- a/lib/analyzer/package-managers/rpm.ts
+++ b/lib/analyzer/package-managers/rpm.ts
@@ -1,3 +1,4 @@
+import { PackageInfo } from "@snyk/rpm-parser/lib/rpm/types";
 import { AnalysisType, AnalyzedPackage, ImageAnalysis } from "../types";
 
 export function analyze(
@@ -32,4 +33,29 @@ function parseLine(text: string, pkgs: AnalyzedPackage[]) {
     };
     pkgs.push(pkg);
   }
+}
+
+export function mapRpmSqlitePackages(
+  targetImage: string,
+  rpmPackages: PackageInfo[],
+): ImageAnalysis {
+  let analysis;
+
+  if (rpmPackages) {
+    analysis = rpmPackages.map((pkg) => {
+      return {
+        Name: pkg.name,
+        Version: pkg.version,
+        Source: undefined,
+        Provides: [],
+        Deps: {},
+        AutoInstalled: undefined,
+      };
+    });
+  }
+  return {
+    Image: targetImage,
+    AnalyzeType: AnalysisType.Rpm,
+    Analysis: analysis as AnalyzedPackage[],
+  };
 }
--- a/lib/inputs/rpm/static.ts
+++ b/lib/inputs/rpm/static.ts
@@ -1,11 +1,18 @@
-import { getPackages } from "@snyk/rpm-parser";
+import { getPackages, getPackagesSqlite } from "@snyk/rpm-parser";
+import { PackageInfo } from "@snyk/rpm-parser/lib/rpm/types";
+import { IParserSqliteResponse } from "@snyk/rpm-parser/lib/types";
 import * as Debug from "debug";
+import { writeFile as writeFileFs } from "fs";
 import { normalize as normalizePath } from "path";
+import { fileSync } from "tmp";
+import { promisify } from "util";
 import { getContentAsBuffer } from "../../extractor";
 import { ExtractAction, ExtractedLayers } from "../../extractor/types";
 import { streamToBuffer } from "../../stream-utils";
 
 const debug = Debug("snyk");
+
+const writeFile = promisify(writeFileFs);
 
 export const getRpmDbFileContentAction: ExtractAction = {
   actionName: "rpm-db",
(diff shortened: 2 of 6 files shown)


lib/analyzer/package-managers/rpm.ts
```diff
@@ -33,3 +34,36 @@ function parseLine(text: string, pkgs: AnalyzedPackage[]) {
...
+    });
+  }
+  return {
+    Image: targetImage,
+    AnalyzeType: AnalysisType.Rpm,
+    Analysis: analysis as AnalyzedPackage[],
+  };
+}
+
+/**
```
yaronschwimmer, line 63: this comment can be removed

lib/inputs/rpm/static.ts
```diff
@@ -1,12 +1,19 @@
-import { getPackages } from "@snyk/rpm-parser";
+import { getPackages, getPackagesSqlite } from "@snyk/rpm-parser";
+import { PackageInfo } from "@snyk/rpm-parser/lib/rpm/types";
+import { IParserSqliteResponse } from "@snyk/rpm-parser/lib/types";
 import * as Debug from "debug";
+import * as fs from "fs";
```
yaronschwimmer, line 5: maybe just import `writeFile` instead of everything?

lib/inputs/rpm/static.ts
```diff
@@ -38,3 +45,42 @@ export async function getRpmDbFileContent(
...
+  try {
+    const fileObj = tmp.fileSync();
+    await writeFile(fileObj.fd, rpmDb);
+
+    const results: IParserSqliteResponse = await getPackagesSqlite(
+      fileObj.name,
+    );
+
+    fileObj.removeCallback(); // removing the temp file created after processing
+    return results.response;
```
yaronschwimmer, line 69: if we only use `response` out of results, why do we return an object and not whatever's in the response?

lib/stream-utils.ts
```diff
@@ -93,3 +95,14 @@ export async function streamToJson<T>(stream: Readable): Promise<T> {
     });
   });
 }
+
+export async function streamToFile(stream: Readable): Promise<tmp.FileResult> {
+  try {
+    const dataBuffer = await streamToBuffer(stream);
+    const fileObj = tmp.fileSync();
+    writeFileSync(fileObj.fd, dataBuffer);
```
yaronschwimmer, line 103: can we use the promisified version of `writeFile` instead like in`lib/inputs/rpm/static.ts`?

yaronschwimmer: Also, do even use this function?

test/system/package-managers/rpm.spec.ts
```diff
@@ -18,4 +19,13 @@ describe("rpm package manager tests", () => {
...
   });
+
+  it("should correctly analyze an rpm image with sqlite DB", async () => {
+    const image = "amazonlinux:2022.0.20220504.1";
+    const pluginResult = await scan({
+      path: image,
+    });
+
+    expect(pluginResult).toMatchSnapshot();
+  });
```
noaleibo1, line 30: I see that the tests we have for RPM (actually for all package managers) are quite pretty short. Are these the only testing we do? Do we want to have more test cases? There might not be interesting test cases, just asking.
For example, do we want to test that we can handle the case of a broken sqlite file? Or this not a likely thing to encounter?

lib/inputs/rpm/static.ts
RotemS: Sorry this was outdated, handling the error as well

lib/stream-utils.ts
RotemS: Removed :)

lib/inputs/rpm/static.ts
```diff
@@ -38,3 +45,42 @@ export async function getRpmDbFileContent(
...
+  const rpmDb = getContentAsBuffer(
+    extractedLayers,
+    getRpmSqliteDbFileContentAction,
+  );
+  if (!rpmDb) {
+    return [];
+  }
+
+  try {
+    const fileObj = tmp.fileSync();
```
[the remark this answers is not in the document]
RotemS: No it isn't. We need it here specifically because we don't analyze it as plain text as we need to open the sqlite db - for that we need a file.
Will change the name :)

test/system/package-managers/rpm.spec.ts
RotemS: We have a couple of more tests related to the data in `rpm-parser`.
I think it's extremely unlikely - if the file is corrupted than the OS will be broken I believe

lib/inputs/rpm/static.ts
```diff
@@ -38,3 +45,42 @@ export async function getRpmDbFileContent(
...
+    const results: IParserSqliteResponse = await getPackagesSqlite(
+      fileObj.name,
+    );
+
+    fileObj.removeCallback(); // removing the temp file created after processing
+    return results.response;
+  } catch (error) {
+    debug(
+      `An error occurred while analysing RPM packages: ${JSON.stringify(
+        error,
```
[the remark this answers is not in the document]
RotemS: This should be only when using the `debug` flag in the cli.

yaronschwimmer (approved):
🤩

lib/inputs/rpm/static.ts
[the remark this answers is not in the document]
noaleibo1: Alright, thanks!

PR #428, opened 2022-06-22
Refactor/code safari nested jars

rotemrip:
- [v] Ready for review

#### What does this PR do?
Rewriting the functionality of comments gathered by the team to make the code more readable.
Replacing reduce function, adding documentation, and renaming function.

[jira](https://snyksec.atlassian.net/browse/MYC-113)

#### Additional questions
2 more todo comments need to be handled - both levels of unpacking and measuring recursion performance.

### DIFF
--- a/lib/analyzer/applications/java.ts
+++ b/lib/analyzer/applications/java.ts
@@ -3,40 +3,34 @@ import * as path from "path";
 import { bufferToSha1 } from "../../buffer-utils";
 import { JarFingerprintsFact } from "../../facts";
 import { JarFingerprint } from "../types";
-import { JarBuffer, JarCoords, JarInfo } from "./types";
+import { AggregatedJars, JarBuffer, JarCoords, JarInfo } from "./types";
 import { AppDepsScanResultWithoutTarget, FilePathToBuffer } from "./types";
 
-function groupJarFingerprintsByPath(input: {
+/**
+ * @param {{[fileName: string]: Buffer}} fileNameToBuffer fileName
+ * @returns {AggregatedJars}
+ */
+function groupJarFingerprintsByPath(fileNameToBuffer: {
   [fileName: string]: Buffer;
-}): {
-  [path: string]: JarBuffer[];
-} {
-  const jarFingerprints: JarBuffer[] = Object.entries(input).map(
-    ([filePath, buffer]) => {
-      return {
-        location: filePath,
-        buffer,
-        coords: null,
-        dependencies: [],
-        nestedJars: [],
-      };
-    },
-  );
-
-  const resultAggregatedByPath = jarFingerprints.reduce(
-    (jarsAggregatedByPath, jarFingerprint) => {
-      const location = path.dirname(jarFingerprint.location);
-      jarsAggregatedByPath[location] = jarsAggregatedByPath[location] || [];
-      jarsAggregatedByPath[location].push(jarFingerprint);
-      return jarsAggregatedByPath;
-    },
-    {},
-  );
+}): AggregatedJars {
+  const resultAggregatedByPath: AggregatedJars = {};
+  Object.keys(fileNameToBuffer).forEach((filePath) => {
+    const location = path.dirname(filePath);
+    const jarFingerprint: JarInfo = {
+      location: filePath,
+      buffer: fileNameToBuffer[filePath],
+      coords: null,
+      dependencies: [],
+      nestedJars: [],
+    };
+    resultAggregatedByPath[location] = resultAggregatedByPath[location] || [];
+    resultAggregatedByPath[location].push(jarFingerprint);
+  });
 
   return resultAggregatedByPath;
 }
 
-export async function jarFilesToScannedProjects(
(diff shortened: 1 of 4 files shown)


yaronschwimmer (commented):
Please fix the following:
* Remove all the `//todo` lines
* Change the commit messages to adhere to the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#specification:~:text=any%20type.-,types%20other%20than,-fix%3A%20and) rules. (Note to self: add this to the linting step). Also remove the dummy commit please.

lib/analyzer/applications/java.ts
```diff
@@ -46,6 +40,7 @@ export async function jarFilesToScannedProjects(
 
   for (const path in mappedResult) {
     if (!mappedResult.hasOwnProperty(path)) {
+      // todo is there a case that this condition will be true?
```
noaleibo1, line 43: Generally speaking, I try to avoid adding `todo`s to the code. Instead, I write a jira ticket with the todo. This will usually help get it prioritised, and is easier to track.
In this specific case, if you don't have the capacity to try to answer the question, I'd consider adding a ticket to research/add a test for what happens if the condition is true.

lib/analyzer/applications/java.ts
```diff
@@ -3,40 +3,34 @@ import * as path from "path";
...
 import { JarFingerprintsFact } from "../../facts";
 import { JarFingerprint } from "../types";
-import { JarBuffer, JarCoords, JarInfo } from "./types";
+import { AggregatedJars, JarBuffer, JarCoords, JarInfo } from "./types";
 import { AppDepsScanResultWithoutTarget, FilePathToBuffer } from "./types";
 
+/**
+ * @param {{[fileName: string]: Buffer}} input fileName
+ * @returns {AggregatedJars}
+ */
```
noaleibo1, line 12: Is this a known syntax for documentation? I'm pretty new to Type Script, interesting to know!
I'm guessing you only added it here because this is the main function who call all other helper functions in the file?

lib/analyzer/applications/java.ts
```diff
@@ -3,40 +3,34 @@ import * as path from "path";
...
 function groupJarFingerprintsByPath(input: {
```
noaleibo1, line 13: Hmm what do you think about having a more meaningful name than `input`? How about something like `fileNameToBuffer`?

lib/analyzer/applications/java.ts
```diff
@@ -3,40 +3,34 @@ import * as path from "path";
...
-      const location = path.dirname(jarFingerprint.location);
-      jarsAggregatedByPath[location] = jarsAggregatedByPath[location] || [];
-      jarsAggregatedByPath[location].push(jarFingerprint);
-      return jarsAggregatedByPath;
-    },
-    {},
-  );
+}): AggregatedJars {
+  const resultAggregatedByPath: AggregatedJars = {};
+  Object.keys(input).forEach((filePath) => {
```
noaleibo1, line 17: Good job on removing `reduce`! Is there a reason you chose to replace `map` with `forEach`? Asking because it seemed more convenient to use map when when I looked at `buffer`.

noaleibo1 (commented):
Congrads on your first PR! 🥳

noaleibo1:
A small comment - I still see the `dummy` commit, did you have trouble removing it after we spoke? We can zoom about it again if you want.

lib/analyzer/applications/java.ts
rotemrip: The addition is similar to existing documentation in the file and project

lib/analyzer/applications/java.ts
rotemrip: Changes

lib/analyzer/applications/java.ts
rotemrip: It seems that in this case, it acts the same, map allocates a new struct and in this case, I thought it is unnecessary (would love to hear if I'm missing anything)

yaronschwimmer (dismissed):
🎉 🎉 🎉

lib/analyzer/applications/java.ts
noaleibo1: But did you add it just for this function?

lib/analyzer/applications/java.ts
noaleibo1: It just seemed a little more readable to me. Especially since you are looking at the value of `buffer`.

lib/analyzer/applications/java.ts
rotemrip: Yes, I thought it would be more readable but maybe if there is no additional usage should I remove it?

noaleibo1: I was just wondering about why you added it just to this function. It's no big deal. You can go ahead and merge!

noaleibo1 (dismissed):

yaronschwimmer (dismissed):

yaronschwimmer (approved):

PR #434, opened 2022-07-11
chore: go binary parsing regression test

yaronschwimmer:
- [X] Ready for review

#### What does this PR do?
Periodically run a CircleCI workflow that:

Builds a Go binary for a test Go application ->
Copies the binary into the go binaries fixtures folder ->
Runs the Go binary parsing unit test

Trying to avoid future issues like [MYC-131](https://snyksec.atlassian.net/browse/MYC-131)


noaleibo1 (approved):
Looks great! Just a question - do you think we should somehow document what this test is doing? I'm just thinking that someone new might not know to make sense out of this.
But maybe I'm just overthinking and it's too cumbersome to document our CI 😅🤷‍♀️

yaronschwimmer:
> Looks great! Just a question - do you think we should somehow document what this test is doing? I'm just thinking that someone new might not know to make sense out of this. But maybe I'm just overthinking and it's too cumbersome to document our CI 😅🤷‍♀️

That's a good idea, though I'm not sure where to put such a documentation 🙂

PR #435, opened 2022-07-19
No experimental check for platform setting [MYC-117]

tommyknows:
- [x] Ready for review

#### What does this PR do?

Mainly removes the check if Docker experimental is enabled if the user specifies a platform. It's not needed anymore as the `--platform` flag for `docker pull` has been graduated to stable. Also some clean-up around that feature.

https://snyksec.atlassian.net/browse/MYC-117

### DIFF
--- a/lib/docker.ts
+++ b/lib/docker.ts
@@ -51,24 +51,16 @@ class Docker {
     return await dockerPull.pull(registry, repo, tag, opt);
   }
 
-  public async isDockerExperimentalEnabled(): Promise<boolean> {
-    const output = await subProcess.execute("docker", ["system", "info"]);
-    return output.stdout.includes("Experimental: true");
-  }
-
-  public async pullCli(targetImage: string, options?: DockerOptions) {
-    let platformOption: string = "";
+  public async pullCli(
+    targetImage: string,
+    options?: DockerOptions,
+  ): Promise<subProcess.CmdOutput> {
+    const opts: string[] = ["pull", targetImage];
     if (options?.platform) {
-      if (await this.isDockerExperimentalEnabled()) {
-        platformOption = `--platform=${options.platform}`;
-      } else {
-        throw new Error(
-          '"--platform" is only supported on a Docker daemon with experimental features enabled',
-        );
-      }
+      opts.push(`--platform=${options.platform}`);
     }
 
-    return subProcess.execute("docker", ["pull", platformOption, targetImage]);
+    return subProcess.execute("docker", opts);
   }
 
   public async save(targetImage: string, destination: string) {
--- a/test/lib/analyzer/image-inspector.spec.ts
+++ b/test/lib/analyzer/image-inspector.spec.ts
@@ -153,14 +153,14 @@ describe("getImageArchive", () => {
     it("should produce the expected state", async () => {
       const customPath = tmp.dirSync().name;
       const imageSavePath = path.join(customPath, uuidv4());
-      const dockerPullSpy = jest.spyOn(Docker.prototype, "pull");
+      const registryPullSpy = jest.spyOn(Docker.prototype, "pull");
 
       const archiveLocation: ArchiveResult = await imageInspector.getImageArchive(
         targetImage,
         imageSavePath,
       );
 
-      expect(dockerPullSpy).not.toHaveBeenCalled();
+      expect(registryPullSpy).not.toHaveBeenCalled();
       expect(archiveLocation.path).toEqual(
         path.join(imageSavePath, "image.tar"),
       );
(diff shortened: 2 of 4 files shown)


yaronschwimmer:
[MYC-131](https://snyksec.atlassian.net/browse/MYC-131) is definitely not the relevant ticket. Is there a relevant ticket?

tommyknows:
> [MYC-131](https://snyksec.atlassian.net/browse/MYC-131) is definitely not the relevant ticket. Is there a relevant ticket?

Sorry, that was a copy-paste issue 😅 Yes, [MYC-117](https://snyksec.atlassian.net/browse/MYC-117). PR description updated as well. 

I'm not sure about the second commit, I marked it as a `BREAKING CHANGE: ...`, but it does not appear in the expected release notes? 🤔

yaronschwimmer:
> > [MYC-131](https://snyksec.atlassian.net/browse/MYC-131) is definitely not the relevant ticket. Is there a relevant ticket?
> 
> Sorry, that was a copy-paste issue 😅 Yes, [MYC-117](https://snyksec.atlassian.net/browse/MYC-117). PR description updated as well.
> 
> I'm not sure about the second commit, I marked it as a `BREAKING CHANGE: ...`, but it does not appear in the expected release notes? 🤔

I don't think it's a breaking change, as it doesn't change anything in the API of the library. The expected release notes is not great with identifying this I think. Why did you consider this a breaking change? 
Also, I wonder why there are 3 commits in this rather small PR. Does each of these commits stand on its own?

lib/docker.ts
```diff
@@ -51,24 +51,13 @@ class Docker {
...
   }
 
-  public async isDockerExperimentalEnabled(): Promise<boolean> {
-    const output = await subProcess.execute("docker", ["system", "info"]);
-    return output.stdout.includes("Experimental: true");
-  }
-
   public async pullCli(targetImage: string, options?: DockerOptions) {
-    let platformOption: string = "";
+    const opts: string[] = ["pull", targetImage];
```
yaronschwimmer, line 55: I'm not sure how smart Docker is with adding the option _after_ the image name (it might, I just don't know and want to make sure it works)

lib/docker.ts
```diff
@@ -51,24 +51,13 @@ class Docker {
...
     if (options?.platform) {
-      if (await this.isDockerExperimentalEnabled()) {
-        platformOption = `--platform=${options.platform}`;
-      } else {
-        throw new Error(
-          '"--platform" is only supported on a Docker daemon with experimental features enabled',
-        );
-      }
+      opts.push(`--platform=${options.platform}`);
```
yaronschwimmer, line 57: I'm not sure how smart Docker is with adding the option after the image name (it might be ok, I just don't know and want to make sure it works)

yaronschwimmer:
> > Why did you consider this a breaking change?
> 
> It removes a function that had been marked as `public`. Even though I can't find any other usages, this is a breaking change, right? 😅
> 
> > Also, I wonder why there are 3 commits in this rather small PR. Does each of these commits stand on its own?
> 
> They certainly could. Whether it would make sense or not... 🤷‍♂️
> 
> I can also squash them into a single commit. Is there some guidance on when to split up / squash specific commits / changes? I put them in three separate commits to more easily describe them, but obviously I can also do the same in a single, larger commit message :)

Well if nothing uses it internally, and it's not exported as part of the library's API (essentially everything that's exported in `lib/index.ts`), and it doesn't change the output of any exported function, then it's not a breaking change IMO.
As for the commits, there are no strict guidelines. If they make sense as separate commits (in the sense that you can theoretically revert one without having to revert the others), then leave it as it is.

lib/docker.ts
tommyknows: It does work, tested that locally (honestly, this should work for every sensible CLI imo, but that's another topic 😄). 

Not sure if it makes sense to test that though? And if yes, how?

tommyknows:
> Well if nothing uses it internally, and it's not exported as part of the library's API (essentially everything that's exported in lib/index.ts), and it doesn't change the output of any exported function, then it's not a breaking change IMO.

Ah I did not know that, I thought a `public ...` was...public :-) Will need to read up on that and change the commit.

> As for the commits, there are no strict guidelines. If they make sense as separate commits (in the sense that you can theoretically revert one without having to revert the others), then leave it as it is.

Theoretically with some merge conflicts you probably could, but it would probably not make too much sense :smile:

tommyknows:
I dropped / squashed the breaking-change commit into the `remove docker experimental check` commit, but decided to keep the refactor-commit separate.

yaronschwimmer (approved):
Congratulations on your first PR at Snyk 🎉

test/lib/analyzer/image-inspector.spec.ts
```diff
@@ -182,6 +182,82 @@ describe("getImageArchive", () => {
...
+      expect(dockerPullCliSpy).toHaveBeenCalled();
+
+      const imageExistsOnDisk: boolean = fs.existsSync(
+        path.join(imageSavePath, "image.tar"),
+      );
+      expect(imageExistsOnDisk).toBe(false);
+    });
+  });
+
+  describe("from remote registry with experimental platform feature binary", () => {
```
tommyknows, line 223: This and the above new tests are pretty much the exact same apart from the returned errors.
Should I use a loop instead for this? 
Or only test one of these?

tommyknows:
Updated the PR with the things we discussed in Slack.

Note that this doesn't really change behaviour in cases of errors: we still *do not* fall back to pulling with the pull library. 
In the two cases that I've added it would not make sense anyway as the pull library does not support a platform setting anyways.

tommyknows:
@yaronschwimmer may I ask for another review please? Thanks!

yaronschwimmer (approved):
🎉

PR #439, opened 2022-08-07
feat: scan poetry files in images [MYC-148]

yaronschwimmer:
- [X] Ready for review

#### What does this PR do?
Given a `pyproject.toml` manifest file and a `poetry.lock` file, both in
the same directory, we will scan for application dependencies in the
project they represent.

https://snyksec.atlassian.net/browse/MYC-148

#### Screenshots
image


lib/analyzer/applications/python.ts
```diff
@@ -0,0 +1,90 @@
...
+        manifest: path.join(directoryPath, "pyproject.toml"),
+        lock: path.join(directoryPath, "poetry.lock"),
+      });
+    }
+  }
+
+  return manifestLockPathPairs;
+}
+
+// assumption: we only care about manifest+lock files if they are in the same directory
```
noaleibo1, line 76: I wonder if we should document this anywhere else? Or is this simply some technical edge case, and not necessarily a "limitation"?

test/system/application-scans/python.spec.ts
```diff
@@ -0,0 +1,55 @@
+import { scan } from "../../../lib";
+import { getFixture } from "../../util";
+
+describe("python application scans", () => {
```
noaleibo1, line 4: What do you think about splitting the tests for `poetry`, `pip` and `pipenv` to different files? They don't actually have anything in common, and that would enable better run times for the tests.

noaleibo1 (commented):
Looks good to me :)
Question - should we be mindful of the poetry version we are implementing support for? This thought came to my mind after the whole story we have with Go.

test/system/application-scans/python.spec.ts
yaronschwimmer: Sure. Once we'll have more package managers we support, we can split the tests to different files

lib/analyzer/applications/python.ts
yaronschwimmer: If these two files are not in the same directory, they might be related to two different projects. We do that for all other manifest+lockfile combos (nodejs, php)

yaronschwimmer:
> Looks good to me :) Question - should we be mindful of the poetry version we are implementing support for? This thought came to my mind after the whole story we have with Go.

I really hope not, as I don't know anything about this :)
Also, that's an advantage of using a parsing library, unlike writing your own parsing code

lib/analyzer/applications/python.ts
noaleibo1: What I had in my mind when writing this comment is that if we document this somewhere it might spare questions from support/customer success in the future if they encounter such case. I understand from your answer that it is the same for all apps right? Do you know if that is documented somewhere? Again, it might be trivial and I'm out of context here :)

test/system/application-scans/python.spec.ts
noaleibo1: I actually think it's better to do it now if it is your intention. It would make the next PR easier to review, and generally allows the reviewer to better understand the test file's purpose.

lib/analyzer/applications/python.ts
```diff
@@ -0,0 +1,90 @@
...
+  return scanResults;
+}
+
+function findManifestLockPairsInSameDirectory(
+  filePathToContent: FilePathToContent,
+): ManifestLockPathPair[] {
+  const fileNamesGroupedByDirectory = groupFilesByDirectory(filePathToContent);
+  const manifestLockPathPairs: ManifestLockPathPair[] = [];
+
+  for (const directoryPath of Object.keys(fileNamesGroupedByDirectory)) {
```
tommyknows, line 55: Not related to this PR, but a general TS question: is there no "nice way" of looping over the keys & values of a map? 
Something like Go's `for k, v := range m { ... }`.

yaronschwimmer: There's nothing similar in JavaScript. The closest thing are [for..in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) loops, but they are generally not recommended

PR #445, opened 2022-08-25
test: fix regression test shell issues

tommyknows:
`==` is a bashism, `=` is the shell equivalent. Interestingly, when run
on MacOS, `==` works as well, which is why local testing didn't bring up
this issue.

Further, `@*` concatenates the array of arguments into a single variable
seperated by the `IFS`, which is a space. So the end result should be
the same.

These changes have been tested in the same Docker image that is used in
CI.

Also, the `-u` flag has been added to the `set` command so that the script fails on any error. This is useful because so far, CI failed due to `golang:latest-alpine` not being a valid image. This image name was only created because the if-check didn't succeed due to the syntax error.

With the `-u` flag, this would have exited earlier and would have made the issue more obvious as well.

- [x] Ready for review

### DIFF
--- a/.circleci/build-go-binary-latest.sh
+++ b/.circleci/build-go-binary-latest.sh
@@ -1,6 +1,6 @@
 #! /bin/bash
 
-set -ex
+set -eux
 
 cd ~/snyk-docker-plugin/test/fixtures/go-binaries/source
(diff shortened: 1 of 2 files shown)


RotemS (approved):
LGTM

PR #450, opened 2022-09-08
fix: allow package version to be non-semver

yaronschwimmer:
- [X] Ready for review

#### What does this PR do?
When comparing metadata files versions, we want to avoid crashing when a version is not a semver valid version. We now coerce the version so we'll be able to compare it safely.


lib/python-parser/metadata-parser.ts
```diff
@@ -29,7 +30,12 @@ export function getPackageInfo(fileContent: string): PythonPackage {
       }
     }
   }
-  return { name: name.toLowerCase(), version, dependencies } as PythonPackage;
+  const validVersion = semver.coerce(version);
+  return {
+    name: name.toLowerCase(),
+    version: validVersion ? validVersion.toString() : version,
```
tommyknows, line 36: Does it make sense to fallback to the original version if the semver library can't parse it? 
Or should we potentially throw an error instead? What's the failure mode here? :)

tommyknows: That sounds to me like you should throw a (custom?) error and handle that case in the calling code to discard this file? 😄

tommyknows: I guess that's the JS way :P

yaronschwimmer: Both ways work :)

yaronschwimmer: I went with your suggestion

lib/python-parser/metadata-parser.ts
```diff
@@ -29,7 +30,17 @@ export function getPackageInfo(fileContent: string): PythonPackage {
...
-  return { name: name.toLowerCase(), version, dependencies } as PythonPackage;
+  const validVersion = semver.coerce(version);
+  if (!validVersion) {
+    throw new PythonInvalidVersionError(
+      `version ${version} is not compatible with semver and cannot be compared`,
+    );
+  }
+  return {
+    name: name.toLowerCase(),
+    version: validVersion ? validVersion.toString() : version,
```
tommyknows, line 41: I guess we don't need that check then, if we throw an error :)

test/unit/python-metadata-parser.spec.ts
```diff
@@ -72,4 +71,55 @@ describe("python metadata parser", () => {
...
+      Requires-Dist: colorama ; platform_system == "Windows"
+      Requires-Dist: importlib-metadata ; python_version < "3.8"
+    `;
+    const packageResult = getPackageInfo(fileContent);
+    expect(packageResult.name).toEqual("click");
+    expect(packageResult.version).toEqual("8.3.0");
+    expect(packageResult.dependencies).toHaveLength(2);
+  });
+
+  it("falis to parse package metadata when version can't be coerced to semver", () => {
```
tommyknows, line 101: NiT: `falis` typo :)

lib/python-parser/metadata-parser.ts
yaronschwimmer: oh, you're right. forgot to delete

tommyknows (approved):

PR #451, opened 2022-09-08
fix: handle Go binaries from the Go distributions

tommyknows:
- [x] Ready for review

#### What does this PR do?

Binaries that are built in the Go distribution (e.g. the Go compiler, `gofmt`, `vet` and many more) do not have a `mod` directive like "normal" Go binaries. This seems to be a special case for all binaries from the Go source / standard library. So far, our parsing logic couldn't handle the missing `mod` directive, which means that all these binaries currently have the same project name `-compiler=gc`.

This commit fixes that behaviour by naming the binaries with a fixed prefix `go-stdlib@` plus the path that we can extract from the `path` directive in the build metadata. This results in names like `go-stdlib@cmd/vet`.

Note that this is only relevant for the CLI output, a `snyk container monitor` does not seem to use the project name from the CLI, but rather the path.


lib/go-parser/go-binary.ts
```diff
@@ -175,8 +175,17 @@ export function extractModuleInformation(
     throw Error("binary contains empty module info");
   }
 
-  const [, mainModuleLine, ...versionsLines] = mod.split("\n");
-  const [, name] = mainModuleLine.split("\t");
+  const [pathDirective, mainModuleLine, ...versionsLines] = mod.split("\n");
+  let [modDirective, name] = mainModuleLine.split("\t");
+  if (modDirective !== "mod") {
```
yaronschwimmer, line 180: do we even care about these binaries? do we have any vulns for the stdlib packages of Go?

tommyknows: My initial thought was that sure, if the Go stdlib has a vulnerability we should also report these binaries so that customers could upgrade their Go env.

But looking into this, we don't return the stdlib packages as a dependency in the depgraph, so no we probably don't 😢 

(Though I think it would make sense to also return stdlib packages...)

tommyknows: Technically, these binaries could also have vulnerabilities in their "own" code (which is not part of the standard library), which we might flag? I don't know :)

yaronschwimmer: I don't know if it's something we have in our DB. It's a good question, and I think we need to answer it first

tommyknows: I think it would be good to show these binaries anyway, to have a more complete picture of what is in a container? With that in mind, it would be secondary whether or not we support vulns for those.

tommyknows: Talked to a few different parties and we only support vulnerabilities in the Go standard library if we find the Go distribution (like the actual apk / deb / whatever package) in the container.

We don't support it from a scanned binary / SCM as of now. I'd argue that supporting this would make a lot of sense...

Either way, I still think my argument from above stands. What do you think?

tommyknows:
Also, do you have any idea how I can fix the lint error (properly)? 😅 
```
lib/go-parser/go-binary.ts:179:8
ERROR: 179:8  prefer-const  Identifier 'modDirective' is never reassigned; use 'const' instead of 'let'.
```

```ts
let [modDirective, name] = mainModuleLine.split("\t");
if (modDirective !== "mod") {
    name = "whatever"
}
```

One is re-assigned, the other one is not.
Obviously I could do:
```ts
let modDirective = ""
let name = ""
[modDirective, name] = maidModuleLine.split("\t")
```
but...that's not better 😅

yaronschwimmer (approved):

PR #456, opened 2022-09-14
fix: skip CGo binaries without PCLNTable

tommyknows:
- [ ] Ready for review

#### What does this PR do?

Some CGo-built binaries might not contain a pcln table, as is described in [Building a better Go linker](https://docs.google.com/document/d/1D13QhciikbdLtaI67U6Ble5d_1nsI4befEd6_k1z91U/view):

> For “cgo” binaries, which may make arbitrary use of C libraries, the
Go linker links all of the Go code into a single native object file and then invokes the system linker to produce the final binary.

One example and the reason why we have found this issue is containerd, that is built with CGo enabled.

I do not know how exactly to build a Go binary without a PCLN table, but somehow it must be possible :-)

So far, we have thrown an error and thus exited the scan if the PCLN table was missing in a Go binary, because we did not expect this case to exist. Now that we know it exists, this commit fixes that behaviour and simply skips the binary in the scan.

### DIFF
--- a/lib/go-parser/go-binary.ts
+++ b/lib/go-parser/go-binary.ts
@@ -18,8 +18,14 @@ export class GoBinary {
     const pclnTab = goElfBinary.body.sections.find(
       (section) => section.name === ".gopclntab",
     );
+
+    // some CGo built binaries might not contain a pclnTab, which means we
+    // cannot scan the files.
+    // TODO: from a technical perspective, it would be enough to only report the
+    // modules, as the only remediation path is to upgrade a full module
+    // anyways. From a product perspective, it's not clear (yet).
     if (pclnTab === undefined) {
-      throw Error("no pcln table found in Go binary");
+      throw Error("no pcln table present in Go binary");
     }
 
     this.matchFilesToModules(new LineTable(pclnTab.data).go12MapFiles());
--- a/lib/go-parser/index.ts
+++ b/lib/go-parser/index.ts
@@ -1,3 +1,4 @@
+import * as Debug from "debug";
 import * as elf from "elfy";
 import { eventLoopSpinner } from "event-loop-spinner";
 import * as path from "path";
@@ -10,6 +11,8 @@ import {
 import { ExtractAction } from "../extractor/types";
 import { DepGraphFact } from "../facts";
 import { GoBinary, readRawBuildInfo } from "./go-binary";
+
+const debug = Debug("snyk");
 
 const ignoredPaths = [
   path.normalize("/boot"),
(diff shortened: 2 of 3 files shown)


lib/go-parser/index.ts
```diff
@@ -151,23 +151,32 @@ export async function goModulesToScannedProjects(
...
+        },
+      });
+    } catch (err) {
+      if (err instanceof MissingPCLNTableError) {
+        // Depending on how a CGo binary is built, it might not have the PCLN
+        // table linked into the binary. In such cases, we don't want to fail,
+        // we simply don't want to scan the binary.
+        continue;
+      }
+      throw err;
```
yaronschwimmer, line 178: is this being caught somewhere? I'm not sure it's a good idea to fail the entire scan in case of an error. I would prefer catching that error, print a debug message, and move on. IMO it's better to have _some_ scan results than break the entire flow because one binary had an unexpected error

tommyknows: >  IMO it's better to have some scan results than break the entire flow because one binary had an unexpected error

Yeah I absolutely agree on that. I'll add a debug-message and remove the throw. I guess we also don't need the new class then.

I'll keep the linter adjustment though, that has been bothering me for a while now :D

lib/go-parser/index.ts
```diff
@@ -151,23 +154,26 @@ export async function goModulesToScannedProjects(
...
+      };
+      scanResults.push({
+        facts: [depGraphFact],
+        identity: {
+          type: DEP_GRAPH_TYPE,
+          targetFile: filePath,
+        },
+      });
+    } catch (err) {
+      debug(`Go binary scan for file ${filePath} failed: ${err}`);
```
noaleibo1, line 175: Would it make sense to add some test to check this?

lib/go-parser/go-binary.ts
```diff
@@ -18,8 +18,14 @@ export class GoBinary {
...
     );
+
+    // some CGo built binaries might not contain a pclnTab, which means we
+    // cannot scan the files.
+    // TODO: from a technical perspective, it would be enough to only report the
+    // modules, as the only remediation path is to upgrade a full module
+    // anyways. From a product perspective, it's not clear (yet).
     if (pclnTab === undefined) {
-      throw Error("no pcln table found in Go binary");
+      throw Error(`no pcln table present in Go binary`);
```
yaronschwimmer, line 28: just nitpicking, but why did you use back-ticks instead of quotes? (I also wonder why the linter was ok with it)

lib/go-parser/index.ts
yaronschwimmer: +1

lib/go-parser/go-binary.ts
tommyknows: No reason actually, why? Is there a difference? 😅

lib/go-parser/index.ts
tommyknows: I could use a binary that we know fails, but looking at those, the smallest one is 7MB...and I'd need to check that into the git repo.

noaleibo1: I don't mind adding a test in another PR.

lib/go-parser/go-binary.ts
yaronschwimmer: Back-ticks are usually used as [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). They can be used as ordinary strings, but for the sake of code uniformity we use one type of quotes. This is why the formatter _will_ replace single quotes with double quotes, even though they're both valid in JS

yaronschwimmer (approved):

lib/go-parser/go-binary.ts
tommyknows: Ah okay, that makes perfect sense.
I adjusted the quotes now.

PR #457, opened 2022-09-14
feat: support deleted files

rotemrip:
- [ V] Ready for review
- [ V] Follows CONTRIBUTING rules

#### What does this PR do?
Avoid checking Python poetry+pip, PHP, node, and Java files that have been deleted.

lib/extractor/index.ts

Added: test/system/deleted-files.spec.ts

https://snyksec.atlassian.net/browse/MYC-147

### DIFF
--- a/lib/extractor/index.ts
+++ b/lib/extractor/index.ts
@@ -75,17 +75,35 @@ function layersWithLatestFileModifications(
   layers: ExtractedLayers[],
 ): ExtractedLayers {
   const extractedLayers: ExtractedLayers = {};
+  const removedFilesToIgnore: Set<string> = new Set();
+
   // TODO: This removes the information about the layer name, maybe we would need it in the future?
   for (const layer of layers) {
     // go over extracted files products found in this layer
     for (const filename of Object.keys(layer)) {
-      // file was not found
-      if (!Reflect.has(extractedLayers, filename)) {
+      // if finding a deleted file - trimming to its original file name for excluding it from extractedLayers
+      if (isDeletedFile(filename)) {
+        removedFilesToIgnore.add(filename.replace(/.wh./, ""));
+      }
+
+      // not adding the original file to extractedLayers
+      // and removing it from the set since it can be found in consecutive layers
+      if (removedFilesToIgnore.has(filename)) {
+        removedFilesToIgnore.delete(filename);
+      } else if (
+        // file was not found + avoid adding deleted files with .wh.
+        !Reflect.has(extractedLayers, filename) &&
+        !isDeletedFile(filename)
+      ) {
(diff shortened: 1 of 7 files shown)


lib/extractor/index.ts
```diff
@@ -75,19 +75,34 @@ function layersWithLatestFileModifications(
...
   for (const layer of layers) {
     // go over extracted files products found in this layer
     for (const filename of Object.keys(layer)) {
-      // file was not found
-      if (!Reflect.has(extractedLayers, filename)) {
+      if (isDeletedFile(filename)) {
+        filesToRemove.add(filename.replace(/.wh./, ""));
+      }
+
+      if (filesToRemove.has(filename)) {
```
agatakrajewska, line 88: It is not quite clear from looking at the code that you are working with both the deleted and not deleted file, it would be nice to either extract a method or add some comment :-)

rotemrip: I have tried to add some comments,
since it's located between loops adding another function isn't ideal for the returned value

test/system/deleted-files.spec.ts
```diff
@@ -0,0 +1,66 @@
+import * as path from "path";
+import * as plugin from "../../lib";
+
+describe("scan result should not contain any deleted file", () => {
```
agatakrajewska, line 4: The test description from `describe` and then `it` test below in line 15 should form a sentence describing a test case. Now we have: 
`describe("scan result should not contain any deleted file"` `it("scan result don't bring deleted facts"`
and ideally we would want to have something more like: 
`describe("deleted files"` `it("should not exist in scan result"` - so the `describe` tells us what the whole suite does and then we can keep adding more tests for the deleting logic. 
Hope that makes sense? :-)

test/system/static.test.ts
```diff
@@ -308,10 +308,10 @@ test("able to scan opensuse/leap images", async (t) => {
  *
  * Even though the node "binary"
  * (actually just an empty file to keep the fixture small)
- * has been removed, we still report it exists.
+ * has been removed, we used to report it exists.
  */
 
-test("bug exists: false positive for deleted files", async (t) => {
+test("bug fixed: used not to recognize deleted files", async (t) => {
```
agatakrajewska, line 314: I don't think we need that test any more, can remove, we have the functionality covered now.

tommyknows:
I still think we should have a test for deleted directories as well, e.g. a test based on a Dockerfile like:
```
FROM busybox
RUN mkdir /app
COPY <whatever> /app/
RUN rm -rf /app
```

Because I don't think the code would properly catch that as of now.

yaronschwimmer (approved):

PR #460, opened 2022-09-27
fix: amending condition for 'manifest unknown' errors

pecodez:
- [ ] Ready for review

#### What does this PR do?

This commit amends the condition used to check against `err.stderr` from the docker daemon, based on the latest response (assume it must have changed), when the manifest for a given image or tag cannot be found. It ensures that the errors is not treated as a generic error and will not fall back to use `snyk-docker-pull` and fail with invalid credentials error when using the docker binary.

https://snyk.zendesk.com/agent/tickets/32820

### DIFF
--- a/lib/analyzer/image-inspector.ts
+++ b/lib/analyzer/image-inspector.ts
@@ -76,7 +76,14 @@ async function pullWithDockerBinary(
       throw new Error(`Operating system is not supported`);
     }
 
-    if (err.stderr && err.stderr.includes("no matching manifest for")) {
+    const unknownManifestConditions = [
+      "no matching manifest for",
+      "manifest unknown",
+    ];
+    if (
+      err.stderr &&
+      unknownManifestConditions.some((value) => err.stderr.includes(value))
+    ) {
       if (platform) {
         throw new Error(`The image does not exist for ${platform}`);
       }
--- a/test/windows/plugin.spec.ts
+++ b/test/windows/plugin.spec.ts
@@ -78,8 +78,8 @@ describe("windows scanning", () => {
     ]);
   });
 
-  it("can static scan for Identifier type image (nginx:1.19.0)", async () => {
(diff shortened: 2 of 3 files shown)


test/windows/plugin.spec.ts
```diff
@@ -78,8 +78,8 @@ describe("windows scanning", () => {
     ]);
   });
 
-  it("can static scan for Identifier type image (nginx:1.19.0)", async () => {
-    const imageNameAndTag = "nginx:1.19.0";
```
pecodez, line 82: ngnix:1.19.0 is expected to fail here but the image does exist; changing to version 1.19.11 that does not exist and never will

yaronschwimmer: So how did this test pass until now? 🤔

lib/analyzer/image-inspector.ts
```diff
@@ -76,7 +76,7 @@ async function pullWithDockerBinary(
       throw new Error(`Operating system is not supported`);
     }
 
-    if (err.stderr && err.stderr.includes("no matching manifest for")) {
+    if (err.stderr && err.stderr.includes("manifest unknown")) {
```
yaronschwimmer, line 79: if the error message has changed in recent versions, shouldn't we make sure we're also backwards compatible and support the old message as well?

test/windows/plugin.spec.ts
pecodez: That is something I am unsure of too, it should not of passed. The test is clearly to check that it fails when the image does not exist, this test looks like it was written before 1.19.0 existed but it should have been failing for a while now. I would like to understand the logic of why it was not failing previously but for now this seems like a valid change.

lib/analyzer/image-inspector.ts
pecodez: Good shout 👍🏻

yaronschwimmer (approved):

PR #462, opened 2022-11-14
Fix/do not trim go versions

ahmed-agabani-snyk:
#### What does this PR do?

- Use full version of go modules in dep-graph

Snyk expects Go Module versions to have one of the following formats:
- `3.0.0` / `3.0.0-19700101000000-0123abcd5678`
- `v3.0.0` / `v30.0-19700101000000-0123abcd5678`

The `v` prefix variant is chosen to align with Snyk Open Source and ensure reporting filters stay consistent between Snyk product lines.

#### Screenshots

Before

image

After

image


lib/go-parser/go-module.ts
```diff
@@ -13,16 +13,4 @@ export class GoModule {
...
     return this.name + "@" + this.version;
   }
-
-  public snykNormalisedVersion(): string {
-    // Versions in Go have leading 'v'
-    let version = this.version.substring(1);
-    // In versions with hash, we only care about hash
-    // v0.0.0-20200905004654-be1d3432aa8f => #be1d3432aa8f
-    version = version.includes("-")
-      ? `#${version.substring(version.lastIndexOf("-") + 1)}`
```
yaronschwimmer, line 23: are we absolutely sure it's safe to remove this behaviour?

lib/go-parser/go-binary.ts
```diff
@@ -43,7 +43,7 @@ export class GoBinary {
           await eventLoopSpinner.spin();
         }
 
-        const version = module.snykNormalisedVersion();
+        const version = module.version;
```
tommyknows, line 46: Nit: can we just get rid of the variable completely? We only use it twice, and I think `module.version` is clEarer about what is going on as well. (The var was introduced because we didn't want to call the function twice).

lib/go-parser/go-module.ts
tommyknows: Personally, I'm not. But we also have no record of why we have it, and it seems to not work at all.

If you have any ideas on how to figure out if it's safe, let us know ;)

tommyknows (commented):
RE the schema changes: it would be possible to ignore the versions of the schema, see the PR that implemented the schema changes:
- [Ignore it in the snapshot](https://github.com/snyk/dep-graph/pull/101/files#diff-8ed6fc483312748d4e6387e13e288c8e954b79cd2fdcb80a079d1e602f0f62f3R36561)
- [Ignore it in the test](https://github.com/snyk/dep-graph/pull/101/files#diff-12d892e46fa106fe1baa442d537e371c5087935d6a0ab3e51d23b7f75355cc49R40)
(both are required)

I'm not saying we should go that route, I don't really know if that makes sense or not. But it would be one easy way of not having to adjust that version every time :-)

lib/go-parser/go-module.ts
tommyknows: Personally, I'm not. But we also have no record of why we have it, and it seems to not work at all.

If you have any ideas on how to figure out if it's safe, let us know ;)

lib/go-parser/go-binary.ts
ahmed-agabani-snyk: I've inlined the version variable.

lib/go-parser/go-module.ts
ahmed-agabani-snyk: I've tagged you in the thread in slack

ahmed-agabani-snyk: > are we absolutely sure it's safe to remove this behaviour?

Not absolutely sure, but comparing with open source and querying the vulnerability database, it appears `v0.0.0-20210616094352-59db8d763f22` to be more correct than `#59db8d763f22`.

yaronschwimmer (approved):

PR #465, opened 2022-12-05
RUN-1928 feat: add new fact imageNames

minsiyang:
- [ ] Ready for review

#### What does this PR do?

 Add new fact imageNames for Image Model service to store reference for the image.

https://snyksec.atlassian.net/browse/RUN-1928

### DIFF
--- a/lib/analyzer/image-inspector.ts
+++ b/lib/analyzer/image-inspector.ts
@@ -4,7 +4,10 @@ import * as mkdirp from "mkdirp";
 import * as path from "path";
 
 import { Docker, DockerOptions } from "../docker";
-import {
+import { ImageName } from "../extractor/image";
+
+import type { DockerPullResult } from "@snyk/snyk-docker-pull";
+import type {
   ArchiveResult,
   DestinationDir,
   DockerInspectOutput,
@@ -118,12 +121,12 @@ async function pullFromContainerRegistry(
   imageSavePath: string,
   username: string | undefined,
   password: string | undefined,
-): Promise<void> {
+): Promise<DockerPullResult> {
   const { hostname, imageName, tag } = extractImageDetails(targetImage);
   debug(
     `Attempting to pull: registry: ${hostname}, image: ${imageName}, tag: ${tag}`,
   );
-  await docker.pull(
+  return await docker.pull(
     hostname,
     imageName,
     tag,
@@ -141,7 +144,7 @@ async function pullImage(
   username: string | undefined,
   password: string | undefined,
   platform: string | undefined,
-): Promise<void> {
+): Promise<ImageName> {
   if (await Docker.binaryExists()) {
     const pullAndSaveSuccessful = await pullWithDockerBinary(
       docker,
@@ -152,17 +155,24 @@ async function pullImage(
       platform,
     );
     if (pullAndSaveSuccessful) {
-      return;
+      return new ImageName(targetImage);
     }
   }
 
-  await pullFromContainerRegistry(
+  const { indexDigest, manifestDigest } = await pullFromContainerRegistry(
     docker,
     targetImage,
     imageSavePath,
     username,
     password,
   );
+
+  const imageName = new ImageName(targetImage, {
+    manifest: manifestDigest,
+    index: indexDigest,
+  });
+
+  return imageName;
 }
 
 /**
@@ -209,7 +219,7 @@ async function getImageArchive(
   }
 
   if (inspectResult === undefined) {
-    await pullImage(
+    const imageName = await pullImage(
       docker,
       targetImage,
       saveLocation,
@@ -220,6 +230,7 @@ async function getImageArchive(
     );
 
     return {
+      imageName,
       path: saveLocation,
       removeArchive: destination.removeCallback,
     };
@@ -230,7 +241,7 @@ async function getImageArchive(
     inspectResult &&
     !isLocalImageSameArchitecture(platform, inspectResult.Architecture)
   ) {
-    await pullImage(
+    const imageName = await pullImage(
       docker,
       targetImage,
       saveLocation,
@@ -239,14 +250,20 @@ async function getImageArchive(
       password,
       platform,
     );
+    return {
+      imageName,
+      path: saveLocation,
+      removeArchive: destination.removeCallback,
+    };
   } else {
     await docker.save(targetImage, saveLocation);
+    const imageName = new ImageName(targetImage);
+    return {
+      imageName,
+      path: saveLocation,
+      removeArchive: destination.removeCallback,
+    };
   }
-
-  return {
-    path: saveLocation,
-    removeArchive: destination.removeCallback,
-  };
 }
 
 function extractImageDetails(targetImage: string): ImageDetails {
--- a/lib/extractor/image.ts
+++ b/lib/extractor/image.ts
@@ -1,0 +1,132 @@
+export { ImageName, ImageDigest };
+class ImageName {
+  public name: string;
+  public tag?: string;
+  public digests: Record<string, ImageDigest>;
+
+  constructor(
+    targetImage: string,
+    digests: { manifest?: string; index?: string } = {},
+  ) {
+    // this regex has been copied from
+    // https://github.com/distribution/distribution/blob/fb2188868d771aa27e5781a32bf78d4c113c18a6/reference/regexp.go#L101
+    // (code has been modified to print the regex), and then adjusted to
+    // Javascript. The required modifications were replacing `[:xdigit:]` with
+    // `[a-fA-F0-9]` and escaping the slashes.
+    // Note that the digest matched in this Regex will match digests that have
+    // uppercase-letters, while the regex used in validateDigest does NOT match
+    // uppercase-letters. This simply matches the behaviour from the upstream
+    // `reference` and `go-digest `packages.
+    //
+    // we're matching pattern: <registry:port_number>(optional)/<image_name>(mandatory):<image_tag>(optional)@<tag_identifier>(optional)
+    // This Regex contains three capture groups:
+    // 1) The repository / image name
+    // 2) tag
+    // 3) digest
+    const re =
+      /^((?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])(?:(?:\.(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]))+)?(?::[0-9]+)?\/)?[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?(?:(?:\/[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?)+)?)(?::([\w][\w.-]{0,127}))?(?:@([A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][A-Fa-f0-9]{32,}))?$/gi;
+
+    const groups = re.exec(targetImage);
+    if (groups === null) {
+      if (targetImage === "") {
+        throw new Error("image name is empty");
+      }
+      if (re.exec(targetImage.toLowerCase()) !== null) {
+        throw new Error("image repository contains uppercase letter");
+      }
+      throw new Error("invalid image reference format");
+    }
+
+    const parsedGroups = {
+      name: groups[1],
+      tag: groups[2],
+      digest: groups[3],
+    };
+
+    this.name = parsedGroups.name;
+
+    const NameTotalLengthMax = 255;
+    if (this.name.length > NameTotalLengthMax) {
+      throw new Error("image repository name is more than 255 characters");
+    }
+
+    this.tag =
+      parsedGroups.tag || parsedGroups.digest ? parsedGroups.tag : "latest";
+
+    this.digests = {};
+    if (digests.index) {
+      this.digests.index = new ImageDigest(digests.index);
+    }
+    if (digests.manifest) {
+      this.digests.manifest = new ImageDigest(digests.manifest);
+    }
+
+    if (parsedGroups.digest) {
+      const digest = new ImageDigest(parsedGroups.digest);
+      if (this.digests.manifest !== digest && this.digests.index !== digest) {
+        this.digests.unknown = digest;
+      }
+    }
+  }
+
+  public getAllNames(): string[] {
+    const names: string[] = [];
+    if (this.tag) {
+      names.push(this.name + ":" + this.tag);
+    }
+    if (this.digests.manifest) {
+      names.push(this.name + "@" + this.digests.manifest.toString());
+    }
+    if (this.digests.index) {
+      names.push(this.name + "@" + this.digests.index.toString());
+    }
+    if (this.digests.unknown) {
+      names.push(this.name + "@" + this.digests.unknown.toString());
+    }
+    return names;
+  }
+}
+
+const ALGORITHM = {
+  sha256: 64,
+  sha384: 96,
+  sha512: 128,
+};
+
+class ImageDigest {
+  public alg: string;
+  public hex: string;
+
+  constructor(digest: string) {
+    // this function matches the implementation
+    // https://github.com/opencontainers/go-digest/blob/b0b31a459546bae38a6c9676e9c5f35861e58894/digest.go#L103
+    const i = digest.indexOf(":");
+    if (i <= 0 || i + 1 === digest.length) {
+      throw new Error("invalid digest format");
+    }
+
+    const alg = digest.slice(0, i);
+    this.hex = digest.slice(i + 1);
+
+    // make sure the algorithm is valid, and that we get the same
+    // amount of hex characters as we expect for the given algorithm.
+    this.alg = alg;
+    if (this.alg === undefined || !Object.keys(ALGORITHM).includes(this.alg)) {
+      throw new Error(`unsupported digest algorithm: ${alg}`);
+    }
+    if (this.hex.length !== ALGORITHM[this.alg]) {
+      throw new Error(
+        `digest algorithm ${this.alg} suggested length ${
+          ALGORITHM[this.alg]
+        }, but got digest with length ${this.hex.length}`,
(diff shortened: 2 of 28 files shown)


lib/analyzer/image-inspector.ts
```diff
@@ -152,17 +155,166 @@ async function pullImage(
...
+    targetImage,
+  );
+}
+
+function getImageNames(
+  indexImageName: string | undefined,
+  manifestImageName: string | undefined,
+  targetImage: string,
+): string[] {
+  const imageNames: string[] = [];
```
ivanstanev, line 192: Nit: to ensure uniqueness, we may want to add these to a Set<string> (which will duplicate automatically), then return the resulting set as a string[]

minsiyang: Sure, I can do that. I guess the `undefined` wouldn't get passed if I keep the if statements.

lib/analyzer/image-inspector.ts
```diff
@@ -152,17 +155,116 @@ async function pullImage(
...
+  return imageNameWithManifestDigest;
+}
+
+export function getImageParts(imageWithTag: string): {
+  imageName: string;
+  imageTag: string;
+  imageDigest: string;
+} {
+  // we're matching pattern: <registry:port_number>(optional)/<image_name>(mandatory):<image_tag>(optional)@<tag_identifier>(optional)
+  // extracted from https://github.com/docker/distribution/blob/master/reference/regexp.go
```
tommyknows, line 250: As written in Slack, I had the following comment in a local branch for exactly this regex:
```
  // this regex has been copied from
  // https://github.com/distribution/distribution/blob/fb2188868d771aa27e5781a32bf78d4c113c18a6/reference/regexp.go#L101
  // (code has been modified to print the regex), and then adjusted to
  // Javascript. The required modifications were replacing `[:xdigit:]` with
  // `[a-fA-F0-9]` and escaping the slashes.
  // Note that the digest matched in this Regex will match digests that have
  // uppercase-letters, while the regex used in validateDigest does *not* match
  // uppercase-letters. This simply matches the behaviour from the upstream
  // `reference` and `go-digest `packages.
  //
  // This Regex contains three capture groups:
  // 1) The repository / image name
  // 2) tag
  // 3) digest
```

But also, my regex was *slightly* different 😅 
```
/^((?:(?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])(?:(?:\.(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]))+)?|\[(?:[a-fA-F0-9:]+)\])(?::[0-9]+)?\/)?[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?(?:(?:\/[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?)+)?)(?::([\w][\w.-]{0,127}))?(?:@([A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][a-fA-F0-9]{32,}))?$/g;
```

(the comment mentions `validateDigest`; a function that made sure that the digest was also correct. There is this slight difference between the digest validation and the regex, for reasons I couldn't figure out)

tommyknows (commented):
I need a second to go through the rest of the changes, haven't looked at the whole image-name building functions.

What I also thought is whether it makes sense to only have a flat list of image names? We have the info in the plugin, so why not return a structure like this as the fact:
```ts
export interface ImageNamesFact {
  type: "imageNames"; // imageReferences? :) 
  data: {
    name: string,
    tag: string | undefined,
    digests: [{
      format: "sha256" | "sha384" | "sha512" , // from https://github.com/opencontainers/go-digest/blob/master/sha.go
      hex: string,
    ]},
  },
}
```

test/unit/image-inspector.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+  [
+    "sha256:b2aa76581b4388a2138869f7e7d777eaa1c8cc13352c712593466711d30f1b95",
+    "nginx@sha256:1a9df42d52d79f0fb19170ee73c61979bec11e1e33b730beaf2d0c6ec333a0f9",
+    "nginx@sha256:b2aa76581b4388a2138869f7e7d777eaa1c8cc13352c712593466711d30f1b95",
+  ],
+];
+
+describe("getImageNameWithIndexDigest", () => {
+  test.each(indexDigestCases)(
+    "given idexDigest %p, taget image %p, expect result to be %p",
```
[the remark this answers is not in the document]
minsiyang: 🙈

test/unit/image-inspector.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+    [
+      "nginx@sha256:b2aa76581b4388a2138869f7e7d777eaa1c8cc13352c712593466711d30f1b95",
+      "nginx:latest",
+    ],
+  ],
+];
+
+describe("getImageNames", () => {
+  test.each(imageNamesCases)(
+    "given manifestDigest %p, target image %p, expect result to be %p",
```
ivanstanev, line 122: Nit: this prints the wrong data (3 substitutions, but 4 arguments passed)

lib/types.ts
```diff
@@ -65,7 +65,8 @@ export type FactType =
   | "imageSizeBytes"
   | "loadedPackages"
   | "imageCreationTime"
-  | "redHatRepositories";
+  | "redHatRepositories"
+  | "imageNames";
```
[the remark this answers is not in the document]
ivanstanev: We had a super lengthy discussion about this as well 😆 
We settled on image names for a few reasons:
- OCI does not actually define a name for this...
- All other vendors in this space call the full reference an "image name" - we went with this for consistency

I also wanted to call it "image reference" but in the end we chose to be consistent - both the internal and public naming for this is "image name" to not cause confusion. 

IMO we can still call it reference here but it will be confusing eventually to have 2 different names internally for the same concept! Docker plugin would call it "reference" but anywhere in our backend it will be "image name"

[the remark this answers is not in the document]
minsiyang: We had this exact same discussion before when designing the image model and came across the same pain points. Originally it was called `imageReference`, until we found the definition of `reference` in the OCI image spec. As you can see in the [Miro board](https://miro.com/app/board/o9J_ltv1rBg=/) here. In the end, as most of the competitors (as you can see in the miro board too) including customers refer to the whole `nginx:latest` as an image name, hence we picked that 😅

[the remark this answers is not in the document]
tommyknows: 😆 fun!

Alright, I definitely agree that consistency is more important here than whatever opinion I (or Ivan ;) ) have about the seemingly correct name for this 😄 

I'm fine with `imageName`, let's keep it like that then.

test/unit/image-inspector.spec.ts
minsiyang: 🙈

lib/analyzer/image-inspector.ts
minsiyang: I am not a regex expert either, but I have tested the possible match here

Your regex https://regex101.com/r/cpokY6/1, it might be started even with the image name. It can't match the full `nginx`. I tried the debugger, and it was showing red starting there, I don't know why though 😅

The regex in Kubernetes Monitor https://regex101.com/r/E91cld/1, it matches all the case 😄

minsiyang:
> I need a second to go through the rest of the changes, haven't looked at the whole image-name building functions.
> 
> What I also thought is whether it makes sense to only have a flat list of image names? We have the info in the plugin, so why not return a structure like this as the fact:
> 
> ```ts
> export interface ImageNamesFact {
>   type: "imageNames"; // imageReferences? :) 
>   data: {
>     name: string,
>     tag: string | undefined,
>     digest: {
>       format: "sha256" | "sha384" | "sha512" , // from https://github.com/opencontainers/go-digest/blob/master/sha.go
>       hex: string,
>     },
>   },
> }
> ```

As for here, I think the format would have been more convenient as an array. In the Image Model, we hope to have a relationship between an image ID that can be related to all its image names, eg, `nginx:latest`, `nginx:1.23.0` or `nginx@sha256:blahblah` can all represent the same image ID. When user query the API they would input the same query `image_name="nginx:latest"` in the API. Hence, having the full names would be faster to look up. Am I making any sense 🤣?

ivanstanev:
Warning: long thoughts but nothing actionable for this PR

In the future, it may be possible that we want to collect all the references from the index and manifest digests for every platform. I guess in that case, we need not only the image name but also the platform associated with that name! Hence, the fact may look roughly like this:
```typescript
interface ImageNamesFact {
  type: "imageNames",
  data: Array<{
    name: string,
    platform: string,
  }>,
}
```
But I'm a bit reluctant to do this future-proofing because I'm not sure that's even something we want to do. It doesn't make sense to collect names+platform without scanning the images behind this names+platform. I don't think even the tools we are trying to integrate with do that (to scan all multi-arch images behind a name). Hence a simple list of known names seemed enough!

So your structure of the names looks pretty useful if we want to know the specific parts composing the name, e.g. the tag, the index digest, the manifest digest, etc. Maybe something we can return with our API! I can see a better need for future-proofing this way, but on the other hand we don't have a direct need for this yet 🤔

lib/analyzer/image-inspector.ts
tommyknows: Interesting, because I thought I tested that and it seemed to work 🤔 Well, good that yours is working then! :)

tommyknows:
Mh I get both of your points, but I would still say that having a concrete type is definitely better than just a list, and doesn't really need to get in the way of "convenience".

For example, the `ImageNamesFact` type could have methods / functions on it that make it easy to get the necessary information. Additionally, if we actually extract all of the image name parsing into a separate library (which afaik is what we're planning on doing), this will get even easier.

E.g. have a class `ImageName` that we can embed into the fact, as well as add some methods on top to make it's use more convenient (from creating the image name to getting a full list of all potential names)

---
Regarding the
> Hence, having the full names would be faster to look up.
You'll need a central store for all image names anyway, right? So does it matter whether you do `allImages.push(newimages...)` or `allImages.push(newImages.names()`? Or am I missing something? 🤔

lib/analyzer/image-inspector.ts
```diff
@@ -152,17 +155,116 @@ async function pullImage(
...
+  if (manifestImageName) {
+    imageNames.add(manifestImageName);
+  }
+  if (indexImageName !== targetImage && manifestImageName !== targetImage) {
+    imageNames.add(targetImage);
+  }
+  return Array.from(imageNames);
+}
+
+export function getImageNameWithIndexDigest(
```
tommyknows, line 206: If you `s/index/manifest/g`, you get the exact same function as `getImageNameWithManifestDigest` 😉 
Or am I missing something? 😅 

I don't think we need two different functions for that 😄 Simply call it `getImageNameWithDigest`, or even just `getImageName`?

tommyknows: (If this is all extracted and put into a class, it doesn't really matter / would need a different API anyway.)

lib/extractor/image.ts
```diff
@@ -1,25 +1,145 @@
...
-  // we're matching pattern: <registry:port_number>(optional)/<image_name>(mandatory):<image_tag>(optional)@<tag_identifier>(optional)
-  // extracted from https://github.com/docker/distribution/blob/master/reference/regexp.go
-  const regex =
-    /^((?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])(?:(?:\.(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]))+)?(?::[0-9]+)?\/)?[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?(?:(?:\/[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?)+)?)(?::([\w][\w.-]{0,127}))?(?:@([A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][A-Fa-f0-9]{32,}))?$/gi;
-  const groups = regex.exec(imageWithTag);
-
-  if (!groups) {
-    throw new Error("Image with tag is malformed, cannot extract valid parts!");
+export { ImageName, ImageDigest, DigestType };
+class ImageName {
```
ivanstanev, line 2: Nit: may need to be public, if `snyk-monitor` would have to provide an `ImageName` passed through the plugin options

lib/extractor/image.ts
```diff
@@ -0,0 +1,145 @@
...
+    const NameTotalLengthMax = 255;
+    if (groups[GROUP.NAME].length > NameTotalLengthMax) {
+      throw Error("image repository name is more than 255 characters");
+    }
+    this.name = groups[GROUP.NAME];
+
+    this.tag =
+      !groups[GROUP.TAG] && !groups[GROUP.DIGEST]
+        ? "latest"
+        : groups[GROUP.TAG];
```
tommyknows, line 55: I think this is easier if you keep the happy path the "primary" one, e.g.
```js
this.tag = groups[GROUP.TAG] || groups[GROUP.DIGEST] ? groups[GROUP.TAG] : "latest"
```

However, we could also do:
```ts
this.tag = groups[GROUP.TAG]
if (!this.tag && !groups[GROUP.DIGEST]) {
  this.tag = "latest"
}
```
This is how'd I do it in Go, not sure if it's also common in JS though 😅

lib/extractor/image.ts
```diff
@@ -0,0 +1,145 @@
...
+        throw Error("image repository contains uppercase letter");
+      }
+      throw Error("invalid image reference format");
+    }
+
+    const GROUP = {
+      NAME: 1,
+      TAG: 2,
+      DIGEST: 3,
+    };
```
tommyknows, line 44: How about doing this instead:
```js
const parsed = {
  name: groups[1],
  tag: groups[2],
  digest: groups[3],
}
```

That way we don't need to constantly index the `group` and can instead directly access the elements, e.g.
```
if (parsed.name.length > NameTotalLengthMax) { ... }
this.name = parsed.name
this.tag = parsed.tag
```

We could also go further and directly assign the name & tag variable to the `this` fields:
```js
    this.name = groups[1];
    this.tag = groups[2];
    const parsedDigest = groups[3];

    const NameMaxLength = 255;
    if (this.name.length > NameMaxLength) {
      throw Error("image repository name is more than 255 characters");
    }

    if (!this.tag && !parsedDigest) {
      this.tag = "latest";
    }
```
(the only thing that we can't directly assign is the digest)

tommyknows (commented):
Nice work!
Few small things, but this is looking pretty good already :)

lib/extractor/image.ts
```diff
@@ -0,0 +1,145 @@
...
+      names.add(this.name + "@" + this.digests[DigestType.Index].toString());
+    }
+    if (this.digests[DigestType.Unknown]) {
+      names.add(this.name + "@" + this.digests[DigestType.Unknown].toString());
+    }
+    return Array.from(names);
+  }
+}
+
+enum DigestType {
```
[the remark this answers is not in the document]
minsiyang: I am afraid you are right about the `unknown` field though. As you can see in the test case here https://github.com/snyk/snyk-docker-plugin/pull/465/files#diff-22abcd98264342dbdf9b059a66cc7c55a1fb14c3bf9b65080b6870667699bfdbR5-R13, when you don't have `manifest` or `index`, if targetImage consist with a digest, only the `unknown` field can deal with it.

lib/extractor/image.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+    if (this.digests[DigestType.Manifest]) {
+      names.add(this.name + "@" + this.digests[DigestType.Manifest].toString());
+    }
+    if (this.digests[DigestType.Index]) {
+      names.add(this.name + "@" + this.digests[DigestType.Index].toString());
+    }
+    if (this.digests[DigestType.Unknown]) {
+      names.add(this.name + "@" + this.digests[DigestType.Unknown].toString());
+    }
```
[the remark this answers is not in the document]
minsiyang: I think for JavaScript, I don't have this option, as you can't iterate an object, but its key or value, which they can't call the `toString()` method. 
https://www.freecodecamp.org/news/how-to-iterate-over-objects-in-javascript/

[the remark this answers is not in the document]
minsiyang: I might totally be lying about can't iterate through an object, but I think the current format is easier to read, especially when the digest only has one entry, it might be overkill to use a loop? 😄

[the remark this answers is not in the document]
tommyknows: > I think for JavaScript, I don't have this option, as you can't iterate an object, but its key or value, which they can't call the toString() method.

You are lying 😛 Afaik this should work:
```js
for (const digest of Object.values(this.digests)) {
  names.push(this.name+"@"+digest.toString())
}
```

I don't think it matters from a readability perspective to do `this.digests.unknown` vs `this.digests[DigestType.Unknown]`. 

Regarding the "to loop or not to loop": Also not sure I agree, but I also don't really mind 😄 

Keep it as is :)

lib/extractor/image.ts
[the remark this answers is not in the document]
tommyknows: Yes, I'm aware :) 
I'm just saying to remove that enum, not the states themselves in the object / map / whatever.

lib/extractor/image.ts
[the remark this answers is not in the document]
minsiyang: I know, that is what I said `I might totally be lying about can't iterate through an object` 😆. 

you can even use
```javascript
Object.values(this.digests).forEach(digest => names.push(this.name+"@"+digest.toString()));
```
However, I don't know whether writing it out specifically would help retain the order, as you can try this [test case](https://github.com/snyk/snyk-docker-plugin/pull/465/files#diff-22abcd98264342dbdf9b059a66cc7c55a1fb14c3bf9b65080b6870667699bfdbR27-R42), the order in the array was reversed if we use a loop. I don't know when iterating the object, whether the order would be preserved, or maybe we don't care about the order, but just need the array `includes()` these names?

I still prefer writing them out explicitly 🥺

PR #470, opened 2022-12-28
fix: Restore system proxy configuration if available

PeterSchafer:
- [x] Ready for review

#### What does this PR do?
The Extensible CLI uses an internal proxy/gateway to forward traffic to snyk.io. Some external tools like docker would by default also pass their traffic through the internal proxy. In case of docker this can cause different issues, which is why this PR restores the system proxy configuration before spawning the docker client.

lib/sub-process.ts

HMMR-627

### DIFF
--- a/lib/sub-process.ts
+++ b/lib/sub-process.ts
@@ -12,11 +12,23 @@ function execute(
   args: string[],
   options?,
 ): Promise<CmdOutput> {
-  const spawnOptions: any = { shell: true };
+  const spawnOptions: any = { shell: true, env: process.env };
   if (options && options.cwd) {
     spawnOptions.cwd = options.cwd;
   }
   args = quoteAll(args, spawnOptions);
+
+  // Before spawning an external process, we look if we need to restore the system proxy configuration,
+  // which overides the cli internal proxy configuration.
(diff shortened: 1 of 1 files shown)


tommyknows (approved):
Changes LGTM, but can you tell me again how this fixes the issue? :D 

Like what's the combination of variables that are currently set, and how they should be set instead.

PeterSchafer:
> Changes LGTM, but can you tell me again how this fixes the issue? :D
> 
> Like what's the combination of variables that are currently set, and how they should be set instead.

The CLI sets the proxy configuration similar to the example below:

> HTTP_PROXY=http://127.0.0.1:Port
> HTTPS_PROXY=http://127.0.0.1:Port
> NO_PROXY=127.0.0.1,localhost
> SNYK_SYSTEM_HTTP_PROXY=<HTTP_PROXY from the env the CLI runs in>
> SNYK_SYSTEM_HTTPS_PROXY=<HTTPS_PROXY from the env the CLI runs in>
> SNYK_SYSTEM_NO_PROXY=<NO_PROXY from the env the CLI runs in>

The change now takes the values from SNYK_SYSTEM_* and restores them by overriding the others.

PR #471, opened 2022-12-29
RUN-2459 fix: pass image with digest via options for kubernetes monitor

minsiyang:
- [x] Ready for review

#### What does this PR do?

Pass the image name with digest via options for Kubernetes Monitor, so we can add that into the imageName facts.

https://snyksec.atlassian.net/browse/RUN-2459

### DIFF
--- a/lib/extractor/image.ts
+++ b/lib/extractor/image.ts
@@ -1,4 +1,6 @@
-export { ImageName, ImageDigest };
+import { PluginOptions } from "../types";
+
+export { ImageName, ImageDigest, getImageNames };
 class ImageName {
   public name: string;
   public tag?: string;
@@ -129,4 +131,29 @@ class ImageDigest {
   public toString(): string {
     return this.alg + ":" + this.hex;
   }
+}
+
+function getImageNames(
+  options?: Partial<PluginOptions>,
+  imageName?: ImageName,
+): string[] {
+  if (imageName) {
+    return imageName.getAllNames();
+  }
+
+  const names: string[] = [];
+  if (options?.imageNameAndTag) {
+    const imageName = new ImageName(options.imageNameAndTag);
+    names.push(...imageName.getAllNames());
+  }
+
+  if (
+    options?.imageNameAndDigest &&
+    options?.imageNameAndDigest !== options?.imageNameAndTag
+  ) {
+    const imageName = new ImageName(options.imageNameAndDigest);
+    names.push(...imageName.getAllNames());
+  }
+
+  return names;
 }
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -151,6 +151,14 @@ export interface PluginOptions {
   imageNameAndTag: string;
 
   /**
+   * WARNING! This is NOT used by the Snyk CLI!
+   *
+   * It is used by K8s-Monitor to preserve the imageNameAndDigest if we can extract this
+   * information from the workload metadata when scanning archives.
+   */
+  imageNameAndDigest: string;
+
+  /**
    * Provide patterns on which to match for detecting package manager manifest files.
    * Used for the APP+OS deps feature, not by the CLI.
    */
--- a/lib/response-builder.ts
+++ b/lib/response-builder.ts
@@ -6,7 +6,6 @@ import * as facts from "./facts";
 
 import { instructionDigest } from "./dockerfile";
 import { DockerFileAnalysis, DockerFilePackages } from "./dockerfile/types";
-import { ImageName } from "./extractor/image";
 import * as types from "./types";
 
 export { buildResponse };
@@ -18,7 +17,7 @@ async function buildResponse(
   },
   dockerfileAnalysis: DockerFileAnalysis | undefined,
   excludeBaseImageVulns: boolean,
-  imageName?: ImageName,
+  names?: string[],
 ): Promise<types.PluginResponse> {
   const deps = depsAnalysis.depTree.dependencies;
   const dockerfilePkgs = collectDockerfilePkgs(dockerfileAnalysis, deps);
@@ -182,8 +181,7 @@ async function buildResponse(
     data: depGraph,
   };
 
-  if (imageName) {
-    const names = imageName.getAllNames();
+  if (names) {
     if (names.length > 0) {
       const imageNameInfo = { names };
       const imageNamesFact: facts.ImageNamesFact = {
--- a/lib/static.ts
+++ b/lib/static.ts
@@ -2,7 +2,7 @@ import * as analyzer from "./analyzer";
 import { StaticAnalysis } from "./analyzer/types";
 import { buildTree } from "./dependency-tree";
 import { DockerFileAnalysis } from "./dockerfile/types";
-import { ImageName } from "./extractor/image";
+import { getImageNames, ImageName } from "./extractor/image";
 import { isTrue } from "./option-utils";
 import { parseAnalysisResults } from "./parser";
 import { buildResponse } from "./response-builder";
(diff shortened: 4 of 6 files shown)


lib/types.ts
```diff
@@ -150,6 +150,14 @@ export interface PluginOptions {
    */
   imageNameAndTag: string;
 
+  /**
+   * WARNING! This is NOT used by the Snyk CLI!
+   *
+   * It is used by K8s-Monitor preserve the imageNameAndDigest when we CAN acquire this
+   * information from workload metadata when scanning archives.
```
tommyknows, line 157:
```suggestion
   * It is used by K8s-Monitor to preserve the imageNameAndDigest if we can extract this
   * information from the workload metadata when scanning archives.
```

lib/extractor/image.ts
```diff
@@ -130,3 +132,23 @@ class ImageDigest {
...
+    names = imageName.getAllNames();
+
+    if (options?.imageNameAndDigest) {
+      names.push(options.imageNameAndDigest);
+    }
+  }
+
+  if (imageName) {
+    names = imageName.getAllNames();
+  }
```
tommyknows, line 152: I'm a bit confused about this. In which cases is `imageNameAndTag` set, and in which ones `imageNameAndDigest`? 

I would have expected that it's either *both* (as we're always able to extract the digest from an image), or only `imageNameAndDigest` (if people are referring to images by digest).

Also, I assume both are prefixed with the image name? Would it make sense to separate these fields out, e.g. to have `imageName`, `imageTag` and `imageDigest` instead?

minsiyang: It is a weird case for Kubernetes Workload metadata. The [imageId](https://github.com/snyk/kubernetes-monitor/blob/694f89958c0e056e042fb00b22ad9855ed604542/src/scanner/index.ts#L100) could be empty on some platforms, if I remember correctly. In that case, `imageNameAndDigest` could be empty sometimes. 

So it can be the case that only `imageNameAndTag` is provided, or you can get both. 

When you do get `imageNameAndDigest`, it is actually either `name@manifestDigest` or `name@indexDigest`, I don't really know how to distinguish it. 
https://github.com/snyk/kubernetes-monitor/blob/694f89958c0e056e042fb00b22ad9855ed604542/src/scanner/index.ts#L105

tommyknows: > So it can be the case that only imageNameAndTag is provided, or you can get both tag and digest.

But it could also be that `imageNameAndTag` is not provided, right? 

So it's either `(imageName xor (imageNameAndTag or imageDigest))`?

> When you do get imageNameAndDigest, it is actually either name@manifestDigest or name@indexDigest, I don't really know how to distinguish it.
That's fine

minsiyang: > But it could also be that imageNameAndTag is not provided, right?

Yes, so the way I see this is as follows:

**Archive flow**
* when DRA calls [scan](https://github.com/snyk/docker-registry-agent/blob/main/src/lib/registries/base-registry.ts#L194-L199), it is a `docker-archive` with only `imageNameAndTag`.
* when k8s calls [scan](https://github.com/snyk/kubernetes-monitor/blob/staging/src/scanner/images/index.ts#L121-L126), it is also a `docker-archive` with `imageNameAndTag` or sometimes with `imageNameAndDigest` as well.
* when it is an actual archive, it won't have `imageNameAndTag`, but just the `path`, eg, https://github.com/snyk/snyk-docker-plugin/blob/5b3d5ef3b4db17b4585f9fb85040257da8e59504/lib/scan.ts#L112-L115

Then, this flow would call [localArchiveAnalysis](https://github.com/snyk/snyk-docker-plugin/blob/5b3d5ef3b4db17b4585f9fb85040257da8e59504/lib/scan.ts#L91), and follow by calling [analyzeStatically](https://github.com/snyk/snyk-docker-plugin/blob/5b3d5ef3b4db17b4585f9fb85040257da8e59504/lib/scan.ts#L117) without `imageName`.

**Docker Pull/Binary Pull**
* it calls [imageIdentifierAnalysis](https://github.com/snyk/snyk-docker-plugin/blob/5b3d5ef3b4db17b4585f9fb85040257da8e59504/lib/scan.ts#L127) , and follow by calling [analyzeStatically](https://github.com/snyk/snyk-docker-plugin/blob/5b3d5ef3b4db17b4585f9fb85040257da8e59504/lib/scan.ts#L150-L158) with `imageName`, but here we won't encounter the `imageNameAndTag` at all.

tommyknows: Okay so if users refer to images with digests (e.g. their `deployment.yaml` specifies `nginx@sha256:.p..`), we will get the `imageNameWithDigest`, but not the `imageNameWithTag` at least in the case of `k8s`. Correct?

If that's true, the current flow doesn't really reflect that I'd argue.
Shouldn't it be something like:
```ts
  if (imageName) {
    // this is the "unique" source of truth, if this is set, the others will never be.
    // At least that mimics the current version.
    return imageName.getAllNames();
  }
  
  let names: string[] = [];
  if (options?.imageNameAndTag) {
    names.push(...new ImageName(options.imageNameAndTag).getAllNames());
  }

  if (options?.imageNameAndDigest) {
    // why do we not call "new ImageName" here too? It would at least make sure that
    // it's well-formatted & makes sure we don't miss anything.
    names.push(options.imageNameAndDigest);
  }
  return names
```

Given that they all could be of type `imageName`, another option would be to:
```ts
var names: ImageName[] = []
if (imageName) {
  names.push(imageName)
}

if (options?.imageNameAndTag) {
  names.push(new ImageName(options.imageNameAndTag))
}

if (options?imageNameAndDigest) {
  names.push(new ImageName(options.imageNameAndDigest);
}
// not sure if that's the right names of methods :) 
return names.map((name) => name.getAllNames()).flatten()
```

But at least the last line I don't find as nice, so... judge for yourself :D

minsiyang: I forgot that part 🙈
OK, so we always have [imageNameAndTag](https://github.com/snyk/kubernetes-monitor/blob/staging/src/scanner/index.ts#L104-L108), it can be in the form of either `name:tag` or `name@digest`.

We sometimes can have `imageNameAndDigest`, if the workload metadata image Id field is present. 
If we have `imageNameAndTag` in the format of `name@digest`, then we would have both `imageNameAndTag` and `imageNameAndDigest`, and they would be the same value. 

**Kubernetes Monitor log for k-m image**
```json
{"name":"kubernetes-monitor","hostname":"snyk-monitor-5d45df59d9-zb424",
  "pid":10,"level":30,"workloadName":"snyk-monitor",
  "imageName":"snyk/kubernetes-monitor:1.100.2","id":"docker.io/snyk/kubernetes-monitor@sha256:85a04bd2ab8547dda75dc557d0898c0e63c0ddc6d1ed118513a366ceec2920a1",
  "msg":"image metadata","time":"2022-12-29T14:54:39.401Z","v":0}
```
k-m workload running in cluster
Screenshot 2022-12-30 at 09 50 48

**Node workload log for node image with digest**
```json
{"name":"kubernetes-monitor","hostname":"snyk-monitor-5d45df59d9-zb424",
  "pid":10,"level":30,"workloadName":"node-deployment",
  "imageName":"node@sha256:345bca77d12ad8f6d2c75c5b8ca0e6de34a84cd7d185af6a3a6c457e7ca64f77",
  "id":"docker.io/library/node@sha256:345bca77d12ad8f6d2c75c5b8ca0e6de34a84cd7d185af6a3a6c457e7ca64f77",
  "msg":"image metadata","time":"2022-12-30T09:57:01.069Z","v":0}
```
node workload running in the cluster
Screenshot 2022-12-30 at 09 58 00

tommyknows (approved):
Thanks a lot, LGTM!

PR #478, opened 2023-03-03
Feat/containerd support

pecodez:
- [x] Ready for review

#### What does this PR do?
Adds support for images pulled using the Docker binary when configured to `containerd` under the hood. `containerd` pulls images and compiles them using an OCI layout but uses the Docker media types as opposed to OCI media types. This broke the scanner as the OCI extractor helpers that identify the manifest indexes use the media type to detect the type of manifest index.

Scanned images that do NOT use the `image identity` now falls back to use the OCI extractor when the Docker extractor fails with an `InvalidArchiveError`.

Setting the beta feature for `containerd` in Docker Desktop and scanninga public image will now succeed, previously it would fail with "Invalid archive" error.

https://snyk.slack.com/archives/C04HJ3ZD0UC/p1677171041359149

https://snyksec.atlassian.net/browse/LUM-147

#### Screenshots
Screenshot 2023-03-03 at 15 41 24


lib/extractor/docker-archive/index.ts
```diff
@@ -26,53 +22,3 @@ export function getImageIdFromManifest(
...
-      diffInHours(new Date(layer.created), lastInstructionTime) <=
-      maxDiffInHours
-    );
-  });
-  // should only happen if there are no layers created by user instructions
-  if (userInstructionLayers.length === history.length) {
-    return [];
-  }
-  return userInstructionLayers;
-}
```
pecodez, line 78: These functions are common to both docker and oci images so moving out of the docker extractor

lib/extractor/index.ts
```diff
@@ -1,13 +1,58 @@
...
+    this.extractActions = actions;
+    this.extractor = extractor;
+  }
+
+  public getExtractor(): Extractor {
+    return this.extractor;
+  }
+
+  public async getLayersAndManifest(): Promise<ExtractedLayersAndManifest> {
+    return await this.extractor?.extractArchive(
```
RotemS, line 37: In which scenario the extractor is undefined?

lib/extractor/index.ts
```diff
@@ -20,55 +65,113 @@ export async function extractImageContent(
...
+    rootFsLayers: getRootFsLayersFromConfig(archiveContent.imageConfig),
+    autoDetectedUserInstructions: getDetectedLayersInfoFromConfig(
+      archiveContent.imageConfig,
+    ),
+    platform: getPlatformFromConfig(archiveContent.imageConfig),
+    imageLabels: archiveContent.imageConfig.config.Labels,
+  };
+}
+
+export function getRootFsLayersFromConfig(imageConfig: ImageConfig): string[] {
```
RotemS, line 127: All of this is a refactor, right?

lib/extractor/index.ts
pecodez: None, it was "leftover" from a previous version where it could have been, but I changed it as it seemed silly 😄

lib/extractor/index.ts
pecodez: Yes, made the logic and fallback easier

lib/extractor/oci-archive/layer.ts
```diff
@@ -176,11 +174,17 @@ function getAllManifestsIndexItems(
 ): OciManifestInfo[] {
   const allManifestsInfo: OciManifestInfo[] = [];
   for (const manifest of imageIndex.manifests) {
-    if (manifest.mediaType === "application/vnd.oci.image.manifest.v1+json") {
+    if (
+      manifest.mediaType === "application/vnd.oci.image.manifest.v1+json" ||
```
RotemS, line 178: Non-blocking: It may be worth defining constants for these media types for readability.

lib/extractor/index.ts
```diff
@@ -1,13 +1,58 @@
...
   ExtractedLayers,
+  ExtractedLayersAndManifest,
   ExtractionResult,
+  Extractor,
   FileContent,
+  ImageConfig,
+  OciArchiveManifest,
 } from "./types";
 
+export class InvalidArchiveError extends Error {}
```
SteveShani, line 20: This is probably not important, but the convention is to add `this.name = "InvalidArchiveError"` in the constructor. I don't actually know why this is needed.

SteveShani (approved):

lib/extractor/oci-archive/layer.ts
pecodez: Added constants

SteveShani (approved):

PR #480, opened 2023-03-14
Go improvements

tommyknows:
**feat: support Go 1.20 binaries**

This commit adds support for Go binaries built with Go 1.20. There
hasn't been any changes to the pclntab or symtab that affect us, see the
change that introduced the new format / magic string
[here](https://cs.opensource.google/go/go/+/0f8dffd0aa71ed996d32e77701ac5ec0bc7cde01).

**test: fix recurring test not actually testing latest**

We added a recurring test to ensure that new Go versions worked or at
least we'd know that they didn't work. However, we never noticed that
the binaries that we build for this recurring test, prefixed with
`latest_<attributes>` are skipped by the test...

This adds it back.

**feat: support replace directives in go.mod**


pecodez (approved):
LGTM

PR #482, opened 2023-04-04
feat(oci): support oci images without local docker installation

jonnyowenpowell:
- [x] Ready for review

#### What does this PR do?

This PR adds support for analyzing OCI images without a local Docker installation. This PR is little more than a version bump to the libraries `snyk-docker-pull` and `docker-registry-v2-client`; the new behaviour is triggered by passing additional content types to avoid being breaking.

Ensure non-OCI and OCI images are both correctly analyzed.

### DIFF
--- a/lib/docker.ts
+++ b/lib/docker.ts
@@ -1,3 +1,4 @@
+import { contentTypes } from "@snyk/docker-registry-v2-client";
 import {
   DockerPull,
   DockerPullOptions,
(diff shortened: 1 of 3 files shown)


lib/docker.ts
```diff
@@ -47,6 +48,14 @@ class Docker {
       password,
       loadImage: false,
       imageSavePath,
+      reqOptions: {
```
shlomiSnyk, line 51: So if I understand correctly `snyk-docker-pull` will use the `reqOptions` to understand if the image is OCI and pull accordingly?

shlomiSnyk (approved):

lib/docker.ts
jonnyowenpowell: `snyk-docker-pull` will pass these accept headers all the way through to the actual request to the OCI distribution spec compliant registry, which will therefore sometimes respond with an OCI image. Additional changes were made in `snyk-docker-pull` to ensure it does correctly handle these images.

PR #483, opened 2023-04-16
fix: fallback to other archive extractor in case of failure

RotemS:
#### What does this PR do?

Attempt to extract oci archive even if the use specified that the archive is a docker archive.
The user of the plugin doesn't necessarily know the archive type. 
Changed to always attempt to use the OCI extractor if the Docker extractor failed.

### DIFF
--- a/lib/extractor/index.ts
+++ b/lib/extractor/index.ts
@@ -93,26 +93,25 @@ export async function extractImageContent(
   let extractor: ArchiveExtractor;
   let archiveContent: ExtractedLayersAndManifest;
 
-  if (extractors.has(imageType)) {
-    extractor = extractors.get(imageType) as ArchiveExtractor;
+  if (!extractors.has(imageType)) {
+    // default to Docker extractor if image type is unknown
+    imageType = ImageType.DockerArchive;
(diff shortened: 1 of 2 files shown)


pecodez (commented):
I am a little conflicted with this one...

This PR assumes we (or the user) does not know what the archive type is and so trys `docker`, then falls back to `OCI`. This makes sense, however, if the user does know the archive type and specifies `oci-archive=` then we still try `docker` first. I'm thinking it would make more sense to try the type defined by `imageType` and fall back to the other on failure.

WDYT?

RotemS:
> I am a little conflicted with this one...
> 
> This PR assumes we (or the user) does not know what the archive type is and so trys `docker`, then falls back to `OCI`. This makes sense, however, if the user does know the archive type and specifies `oci-archive=` then we still try `docker` first. I'm thinking it would make more sense to try the type defined by `imageType` and fall back to the other on failure.
> 
> WDYT?

Convinced. Changed the logic :)

pecodez (approved):

pecodez (approved):
LGTM

PR #485, opened 2023-04-21
Repro/windows go issues

tommyknows:
**chore: reproduce windows scanning issues**

Customer has reported a bug with scanning some Go binaries on windows.
Interestingly, not all binaries seem to fail, only some. For the
failing binaries, no dependencies are being found.

This commit adds a test that runs on Windows to make sure that the
scanning works. The snapshot has been generated by running the test on
MacOS, which illustrates the point that they _should_ produce the exact
same snapshot as the Windows one.

The only thing that needed changing was the filepaths, e.g. from
`"targetFile": "/livenessprobe"` to `"targetFile": "\\livenessprobe"`.

**fix: use path.posix for Go file / module lookup**

This commit switches the usage of the `path` module to use `path.posix`
instead, fixing an issue where when the CLI was running on Windows, the
Go dependency resolution returned no results.

The underlying issue is that the Go binary metadata always contains
normal slashes in it's path, while our usage of `path.join` and
`path.parse` meant that we were constructing paths with backward-slashes
instead.

### DIFF
--- a/lib/go-parser/go-binary.ts
+++ b/lib/go-parser/go-binary.ts
@@ -104,7 +104,7 @@ export class GoBinary {
           // (github.com/my/pkg@v0.0.1/a.go), the path.parse expression returns
           // just a slash. This would result in a package name with a trailing
           // slash, which is incorrect.
-          let dirName = path.parse(parts[1]).dir;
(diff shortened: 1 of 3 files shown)


pecodez (approved):
LGTM

PR #497, opened 2023-06-02
feat: include upstream/source name and version in purl

ivanstanev:
- [x] Ready for review

#### What does this PR do?

The source version of a package may not be the same as the package version, which is why we should track them whenever they differ. This change ensures we populate an "upstream" qualifier (similarly to how Syft does this) with the source name and version. "upstream" is a non-standard qualifier, but matches Syft for consistency.

Tracking the source version is important because Snyk's vuln intel on Debian is based on the source version not the binary version. Some packages like postgres-client have differing binary and source versions, which results in reporting false positives once the wrong vulnerable version range is matched.

### DIFF
--- a/lib/analyzer/package-managers/apt.ts
+++ b/lib/analyzer/package-managers/apt.ts
@@ -1,3 +1,4 @@
+import { PackageURL } from "packageurl-js";
 import {
   AnalysisType,
   AnalyzedPackage,
@@ -43,18 +44,45 @@ export function analyzeDistroless(
 
 function parseDpkgFile(text: string): AnalyzedPackageWithVersion[] {
   const pkgs: AnalyzedPackageWithVersion[] = [];
-  let curPkg: any = null;
+  let curPkg: AnalyzedPackageWithVersion | null = null;
   for (const line of text.split("\n")) {
-    curPkg = parseDpkgLine(line, curPkg, pkgs);
+    curPkg = parseDpkgLine(line, curPkg!, pkgs);
+    if (curPkg) {
+      curPkg.Purl = purl(curPkg);
+    }
   }
   return pkgs;
+}
+
+export function purl(curPkg: AnalyzedPackageWithVersion): string | undefined {
+  if (!curPkg.Name || !curPkg.Version) {
+    return undefined;
+  }
+
+  const qualifiers: { [key: string]: string } = {};
+  if (curPkg.Source && curPkg.SourceVersion) {
+    qualifiers.upstream = `${curPkg.Source}@${curPkg.SourceVersion}`;
+  } else if (curPkg.Source) {
+    qualifiers.upstream = curPkg.Source;
+  }
+
+  return new PackageURL(
+    "deb",
+    "",
+    curPkg.Name,
+    curPkg.Version,
+    // make sure that we pass in undefined if there are no qualifiers, because
+    // the packageurl-js library doesn't handle that properly...
+    Object.keys(qualifiers).length !== 0 ? qualifiers : undefined,
+    undefined,
+  ).toString();
 }
 
 function parseDpkgLine(
   text: string,
   curPkg: AnalyzedPackageWithVersion,
   pkgs: AnalyzedPackageWithVersion[],
-): AnalyzedPackage {
+): AnalyzedPackageWithVersion {
   const [key, value] = text.split(": ");
   switch (key) {
     case "Package":
@@ -62,6 +90,7 @@ function parseDpkgLine(
         Name: value,
         Version: "",
         Source: undefined,
+        SourceVersion: undefined,
         Provides: [],
         Deps: {},
         AutoInstalled: undefined,
@@ -72,7 +101,22 @@ function parseDpkgLine(
       curPkg.Version = value;
       break;
     case "Source":
-      curPkg.Source = value.trim().split(" ")[0];
+      /**
+       * The value may look something like this:
+       * libgcc6 (1.3.0-b1)
+       * <name> (<version>)
+       *
+       * For example, Syft matches these values with a regex:
+       * https://github.com/anchore/syft/blob/1764e1c3f6bd66781f8350d957a1f95e4d9ad3de/syft/pkg/cataloger/deb/parse_dpkg_db.go#L169-L173
+       */
+      const parts = value.split(" ");
+      curPkg.Source = parts[0];
+      if (parts.length > 1) {
+        curPkg.SourceVersion = parts[1]
+          .trim()
+          .replace("(", "")
+          .replace(")", "");
+      }
       break;
     case "Provides":
       for (let name of value.split(",")) {
--- a/test/system/dockerfile-analysis/__snapshots__/dockerfile.spec.ts.snap
+++ b/test/system/dockerfile-analysis/__snapshots__/dockerfile.spec.ts.snap
@@ -4922,6 +4922,7 @@ Object {
                 "id": "acl@2.2.52-2",
                 "info": Object {
                   "name": "acl",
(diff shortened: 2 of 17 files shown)


lib/analyzer/package-managers/apt.ts
```diff
@@ -43,25 +44,53 @@ export function analyzeDistroless(
 
 function parseDpkgFile(text: string): AnalyzedPackageWithVersion[] {
   const pkgs: AnalyzedPackageWithVersion[] = [];
-  let curPkg: any = null;
+  let curPkg: AnalyzedPackageWithVersion | null = null;
   for (const line of text.split("\n")) {
-    curPkg = parseDpkgLine(line, curPkg, pkgs);
+    curPkg = parseDpkgLine(line, curPkg!, pkgs);
+    assignPurl(curPkg);
```
tommyknows, line 50: I think it would be more explicit to instead "compute" a purl and return it, e.g.
```
curPkg.Purl = purl(curPkg)
```

lib/analyzer/package-managers/apt.ts
```diff
@@ -43,25 +44,53 @@ export function analyzeDistroless(
...
+    return;
+  }
+
+  const pkgName = curPkg.Source
+    ? `${curPkg.Source}/${curPkg.Name}`
+    : curPkg.Name;
+  const qualifiers =
+    curPkg.Source && curPkg.SourceVersion
+      ? {
+          upstream: encodeURIComponent(
```
tommyknows, line 66: A comment on why we need to encode this would be helpful, I don't think it's obvious :)

lib/analyzer/package-managers/apt.ts
```diff
@@ -72,7 +101,14 @@ function parseDpkgLine(
...
     case "Source":
-      curPkg.Source = value.trim().split(" ")[0];
+      const parts = value.trim().split(" ");
+      curPkg.Source = parts[0];
+      if (parts.length > 1) {
+        curPkg.SourceVersion = parts[1]
+          .trim()
+          .replace("(", "")
+          .replace(")", "");
+      }
```
tommyknows, line 111: A comment explaining what the value might look like and what we need to extract would be helpful here.

I understand what it's doing, but I don't understand what the outcome of it is :) (I can guess, but an example would make it easier)

lib/analyzer/package-managers/apt.ts
```diff
@@ -43,25 +44,53 @@ export function analyzeDistroless(
...
+      ? {
+          upstream: encodeURIComponent(
+            `${curPkg.Source}@${curPkg.SourceVersion}`,
+          ),
+        }
+      : undefined;
+  curPkg.Purl = new PackageURL(
+    "deb",
+    "",
+    pkgName,
```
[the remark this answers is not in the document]
ivanstanev: Wouldn't that break existing Snyk consumers of this purl? I think the purl exporter includes the source, unless this was changed? CC @ChristinaDara if you know!

Though I agree it should be part of the qualifiers, I think doing so requires adapting all the code that reads purls to use this new location... (again unsure, so tagging @ChristinaDara if you know!)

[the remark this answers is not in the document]
tommyknows: I don't think it's necessarily relevant here, as we're just generating the container and doing the backend.

It ties into Christina's work only if you say that our backend is not going to be able to parse the `source/package` style names anymore.

I'd simply prefer a `source=abc` qualifier, if exists, but fall back to checking the package name incl. the slash (we need to do this anyway to stay backwards-compatible!)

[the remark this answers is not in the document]
ChristinaDara: @ivanstanev 
In the `os-purl-parser-container library` we are using the distro from the qualifiers, name, version and namespace to get the `export interface VulnDbUrlComponents {
  packageName: string;
  packageManager: string;
  packageVersion?: string;
}` The manager is a combination of the distro and namespace.
I hope this is the right answer

lib/analyzer/package-managers/apt.ts
```diff
@@ -43,25 +44,55 @@ export function analyzeDistroless(
...
+  } else if (curPkg.Source) {
+    qualifiers = {
+      upstream: encodeURIComponent(curPkg.Source),
+    };
+  }
+
+  return new PackageURL(
+    "deb",
+    "",
+    curPkg.Source ? `${curPkg.Source}/${curPkg.Name}` : curPkg.Name,
```
tommyknows, line 76: Again I think this should *only* be the name (and the validation ensures that as well right now).

But maybe I'm missing something, let me know if you want to pair on this.

test/system/dockerfile-analysis/__snapshots__/dockerfile.spec.ts.snap
```diff
@@ -4922,1896 +4922,2167 @@ Object {
...
                   "name": "apt",
+                  "purl": "pkg:deb/apt@1.0.9.8.4",
                   "version": "1.0.9.8.4",
                 },
               },
               Object {
                 "id": "libsigsegv/libsigsegv2@2.10-4+b1",
                 "info": Object {
                   "name": "libsigsegv/libsigsegv2",
+                  "purl": "pkg:deb/libsigsegv2@2.10-4%2Bb1?upstream=libsigsegv%402.10-4",
```
ivanstanev, line 4965: An example of a properly encoded purl 🙏

tommyknows (approved):
🎉 

Note: we only attach the `upstream` qualifier if the `Source` is set - we omit it if the package doesn't have a source set. 

I know we decided that we're always adding the source even if it may be equal to the package name, but I agree that in the case of the source not being set at all, this hasn't been decided 😄 

I'm approving this PR, as I don't have any objections to omitting it if it's not set in the apt-DB. 

I'll leave it to you to decide :)

test/system/operating-systems/__snapshots__/distroless.spec.ts.snap
```diff
@@ -102,41 +102,47 @@ Object {
...
                   "name": "netbase",
+                  "purl": "pkg:deb/netbase@5.6",
                   "version": "5.6",
                 },
               },
               Object {
                 "id": "openssl/libssl1.1@1.1.1d-0+deb10u3",
                 "info": Object {
                   "name": "openssl/libssl1.1",
+                  "purl": "pkg:deb/libssl1.1@1.1.1d-0%2Bdeb10u3?upstream=openssl",
```
tommyknows, line 129: I just added a test to our [container-sbom-export](https://github.com/snyk/container-sbom-export/blob/main/pkg/purl) library and that doesn't seem to escape the special chars:
```
pkg:deb/debian/libssl1.1@1.1.1d-0+deb10u3?upstream=openssl&distro=buster
```
I'm wondering if it should really be escaped? 🤔

tommyknows: I guess it's the purl libraries that behave differently, as the TS seems to escape it and the Go one doesn't? 🥴

tommyknows: The Purl spec says that "purl is just a URL", so I'd argue URL rules apply.
That means however that a plus-sign is a valid character and does *not* need to be escaped unless it's in the query component of the URL.
See [this S/O answer](https://stackoverflow.com/a/1006074) for more details.

tommyknows: purl's Go library's `Parse` function calls a `url.PathUnescape` function, which mentions in it's docs that:
> PathUnescape is identical to QueryUnescape except that it does not unescape '+' to ' ' (space).

😄 

So both examples (escaped and not) work! 
I'm wondering whether the JS library behaves the same though....

PR #499, opened 2023-06-07
chore: add openapi common components

ahmed-agabani-snyk:
#### What does this PR do?

Adds OpenAPI common components

### DIFF
--- a/components/common.yaml
+++ b/components/common.yaml
@@ -1,0 +1,66 @@
+schemas:
+  ContainerTarget:
+    type: object
+    description: The target defines "where" you found this scan result.
+    properties:
+      image:
+        type: string
+  Fact:
+    type: object
+    description: The thing you found.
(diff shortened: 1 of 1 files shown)


minsiyang (approved):

tommyknows (approved):
I'm _generally_ okay with this approach, but as long as we don't generate the types from this, it's in the same danger of going out of sync with the code (granted, at least _if you know about it_, there's only one repo you'll need to edit now!).

Looking at the definitions here, I'm actually not sure if we _can_ generate (nice) types out of this; the TS facts are all typed, meaning that the `data` field has a more concrete type than just `object`.

Maybe we should instead have a map / object with optional fields, e.g.
```
{
    depGraph: <depGraphData>
    keyBinariesHashes: <data>
}
```
but that's another, larger change 😄

PR #500, opened 2023-06-07
chore: add openapi spec fact types

ahmed-agabani-snyk:
#### What does this PR do?

- chore: alphabetically sort fact types
- chore: document openapi missing fact types

### DIFF
--- a/components/common.yaml
+++ b/components/common.yaml
@@ -17,8 +17,10 @@ schemas:
     type: string
     enum:
       - autoDetectedUserInstructions
+      - binaries
       - depGraph
       - dockerfileAnalysis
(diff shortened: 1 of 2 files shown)


components/common.yaml
```diff
@@ -17,8 +17,10 @@ schemas:
     type: string
     enum:
       - autoDetectedUserInstructions
+      - binaries
```
minsiyang, line 20: Should these missing types also exit in `lib/types.ts` as well?

ahmed-agabani-snyk: Adding them to typescript would be misleading because snyk-docker-plugin doesn't actively use them.

PR #508, opened 2023-06-28
feat: passing manifest and index digest to archive scanning

minsiyang:
- [x] Ready for review

#### What does this PR do?

Passing digests scan option to docker and oci archive scanning to enrich imageName fact type.

* lib/scan.ts
* lib/types.ts

scan an OCI/docker archive by providing both `imageNameAndTag` and `digests` scan options.

This implementation would enable Docker Registry Agent to pass both manifest and index digest information to enrich imageName fact type in order for Container Image Service to consume this data.

### DIFF
--- a/lib/scan.ts
+++ b/lib/scan.ts
@@ -4,6 +4,7 @@ import * as path from "path";
 import { getImageArchive } from "./analyzer/image-inspector";
 import { readDockerfileAndAnalyse } from "./dockerfile";
 import { DockerFileAnalysis } from "./dockerfile/types";
+import { ImageName } from "./extractor/image";
 import { fullImageSavePath } from "./image-save-path";
 import { getArchivePath, getImageType } from "./image-type";
 import { isNumber, isTrue } from "./option-utils";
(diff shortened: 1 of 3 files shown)


ivanstanev (approved):

lib/scan.ts
```diff
@@ -114,13 +115,25 @@ async function localArchiveAnalysis(
     // The target image becomes the base of the path, e.g. "archive.tar" for "/var/tmp/archive.tar"
     path.basename(archivePath);
 
+  let imageName: ImageName | undefined;
+  if (
+    (options.digests?.manifest || options.digests?.index) &&
```
tommyknows, line 120: Why are these checks necessary - afaict both digests are optional in the `new ImageName` call, so I'm not sure why we need to ensure one is set :)

lib/types.ts
```diff
@@ -157,6 +157,14 @@ export interface PluginOptions {
    */
   imageNameAndDigest: string;
 
+  /**
+   * WARNING! This is NOT used by the Snyk CLI!
+   *
+   * It is used by Docker Registry Agent to preserve the image digests if we can extract this
+   * information from pulling the image with Snyk Docker Pull.
+   */
+  digests: { manifest?: string; index?: string };
```
tommyknows, line 166: I thought about asking to have some "complex" type to do something like this (pseudocode):
```
digests?: AtLeastOneOf{ manifest: string, index: string}
```
But the solution's actually quite complex, so...let's keep it the way it is 😆

tommyknows (approved):
Only a small, non-blocking question :) Otherwise LGTM!

lib/scan.ts
tommyknows: Yeah that makes total sense. Thanks! :)

PR #511, opened 2023-07-12
fix docker pull issues

tommyknows:
**chore: update shescape**

When debugging issues within our Windows CI, I've noticed that
`shescape` has a couple of bug fixes related to Windows. Although it
turned out that `shescape` is not the issue here, I'm keeping that
update in just for good measure.

**fix: error handling when pulling images with pull library**

So far, we've only handled errors when pulling with the Docker CLI. With
the Docker CLI, we've "translated" errors to make them more helpful.

The Windows tests have now started to fail because of a broken Docker
installation, as we then fell back to the `snyk-docker-pull` library,
which didn't have that error translation.

This commit fixes that, and applies that error handling to both the CLI
errors as well as errors from snyk-docker-pull.

**fix: error messages in debugs**

Many may `[object Object]`s in our debug messages which are not helpful.

**chore: windows docker tests**

As we noticed that the tests failed because of a broken Docker
installation, we've decided to duplicate the job and run one job with a
working, and one with a broken Docker CLI.

This should give us some guarantee that these environments and the
respective pull libraries behave the same.

**chore: ensure Docker in CI to work**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -157,6 +157,7 @@ jobs:
           name: Use snyk-main npmjs user
           command: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" >> .npmrc
       - run: npm ci
+      - run: docker version
       - run:
           command: npm run test-jest-windows
           no_output_timeout: 20m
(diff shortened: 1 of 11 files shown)


minsiyang (approved):
🚀

pecodez (approved):
LGTM

PR #518, opened 2023-07-24
test: fix test flake by using rmdirSync

tommyknows:
The test asserts that cleanup has worked and the directory has been removed. However, because `rmdir` is asynchronous (internally, it's not a Promise), sometimes the directory was still there when our assertions ran.

Using `rmdirSync` should fix these issues and make us wait for the directory to be deleted.
The impact to runtime should be negligible, due to the fact that we delete the image-archive *before* the `rmdir` call with an `unlinkSync` already.


pecodez (approved):
LGTM

PR #523, opened 2023-08-09
fix: removes meta-common-packages generation for Linux package managers

pecodez:
- [x] Ready for review

#### What does this PR do?
Frequently found dependencies can cause many upgrade paths to be generated which creates a bad UX when issue grouping is applied in the UI. Grouping these packages under a "fake" package, `meta-common-packages` solved this but broke the upgradable logic for Linux package managers, which need direct dependencies to be upgradable.

This commit attaches frequently found dependencies directly to the tree root for Linux package managers so the upgrade path shows the package as a direct dependency and becomes upgradable.

> Updated test snapshots to take account of `meta-common-packages` package no longer being present in scan results.

[SUP-1367](https://snyksec.atlassian.net/browse/SUP-1637)

#### Screenshots
Screenshot 2023-07-21 at 17 52 03

[SUP-1367]: https://snyksec.atlassian.net/browse/SUP-1367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/lib/dependency-tree/index.spec.ts
+++ b/test/lib/dependency-tree/index.spec.ts
@@ -1,0 +1,78 @@
+import { AnalyzedPackageWithVersion } from "../../../lib/analyzer/types";
+import { buildTree } from "../../../lib/dependency-tree";
+
+const targetImage = "snyk/deptree-test@1.0.0";
+const targetOS = {
+  name: "snykos",
+  version: "1",
+  prettyName: "SnykOS GNU/Linux 1",
+};
+const freqDep = {
+  Name: `libsut`,
+  Version: "1.0",
+  Provides: [],
+  Deps: {},
+  AutoInstalled: true,
+  Purl: `pkg:snyk/libsut@1.0`,
+};
+
+function buildFreqDepsList(count: number) {
+  const depsInfoList: AnalyzedPackageWithVersion[] = [];
+
+  for (let i = 0, l = count; i < l; i = i + 1) {
+    depsInfoList.push({
+      Name: `libtest-${i + 1}`,
+      Version: "1.0",
+      Provides: [],
+      Deps: { libsut: true },
+      AutoInstalled: true,
+      Purl: `pkg:snyk/libtest-${i + 1}@1.0`,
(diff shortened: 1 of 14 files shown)


test/lib/dependency-tree/index.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+
+describe("dependency-tree", () => {
+  describe("buildTree", () => {
+    describe("Linux Package Managers", () => {
+      it("should attach frequent deps to parent when threshold is exceeded", () => {
+        const fixture = buildFreqDepsList(100);
+        const tree = buildTree(targetImage, "deb", fixture, targetOS);
+        const res = obj2array(tree.dependencies);
+        expect(
+          res.filter((dep) => !!dep.dependencies[freqDep.Name]),
```
SteveShani, line 57: First time I've seen `!!` being used. I think it's okay, just letting you know it's not super clear.
Instead, I would have used `dep.dependencies.includes(freqDep.Name)`.

SteveShani (approved):

test/lib/dependency-tree/index.spec.ts
pecodez: Thanks @SteveShani I think this was left over from my initial tests that were coercing a `"true"` value to `true` where using `!!` seemed appropriate. This, however, is indeed confusing!

pecodez: Yes, I would use `hasOwnProperty` to check the scope of a given property to ensure it has not been inherited or within a `for in` loop to ensure I filter out built-ins (or avoid `for in` loops altogether 😄), but feel if I am simply checking for a known property in an object then this type of check is fine and clear on intent WDYT?

SteveShani: Yeah definitely, whatever you choose to do is fine with me.

SteveShani (approved):

PR #525, opened 2023-08-24
fix: nodejs binary wrongfully identified as dpkg

RotemS:
- [x] Ready for review

#### What does this PR do?

In distroless node images, a nodejs binary copied into the image is also included in /var/lib/dpkg/status.d/
which results in adding this binary to the dependnecy graph vulnerability data for the binary and dpkg is different to avoid false positives, this removes it from the graph.

 https://github.com/GoogleContainerTools/distroless/blob/main/private/remote/node_archive.bzl#L29

The added test describes the expected output, where `nodejs` should not be included in the graph, but only in the `keyBinaryHashes` fact.


pecodez (approved):
LGTM

PR #526, opened 2023-08-27
fix: nodejs binary wrongfully identified as dpkg

RotemS:
- [ ] Ready for review

#### What does this PR do?

In distroless node images, a nodejs binary copied into the image is also included in /var/lib/dpkg/status.d/
which results in adding this binary to the dependnecy graph vulnerability data for the binary and dpkg is different to avoid false positives, this removes it from the graph.

 https://github.com/GoogleContainerTools/distroless/blob/main/private/remote/node_archive.bzl#L29

The added test describes the expected output, where `nodejs` should not be included in the graph, but only in the `keyBinaryHashes` fact.

### DIFF
--- a/lib/inputs/distroless/static.ts
+++ b/lib/inputs/distroless/static.ts
@@ -16,6 +16,12 @@ export function getAptFiles(extractedLayers: ExtractedLayers): string[] {
     if (!("dpkg" in extractedLayers[fileName])) {
       continue;
     }
+    // when the nodejs distroless image is build, the metadata is added to status.d
+    // this causes us to wronfgully identify nodejs as deb package
+    // https://github.com/GoogleContainerTools/distroless/blob/main/private/remote/node_archive.bzl#L29
+    if (fileName === "/var/lib/dpkg/status.d/nodejs") {
+      continue;
+    }
     files.push(extractedLayers[fileName].dpkg.toString("utf8"));
   }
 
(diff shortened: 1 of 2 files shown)


lib/inputs/distroless/static.ts
```diff
@@ -16,6 +16,12 @@ export function getAptFiles(extractedLayers: ExtractedLayers): string[] {
     if (!("dpkg" in extractedLayers[fileName])) {
       continue;
     }
+    // when the nodejs distroless image is build, the metadata is added to status.d
```
SteveShani, line 19: If you care about this, built*

lib/inputs/distroless/static.ts
```diff
@@ -16,6 +16,12 @@ export function getAptFiles(extractedLayers: ExtractedLayers): string[] {
...
+    // this causes us to wronfgully identify nodejs as deb package
+    // https://github.com/GoogleContainerTools/distroless/blob/main/private/remote/node_archive.bzl#L29
+    if (fileName === "/var/lib/dpkg/status.d/nodejs") {
```
SteveShani, line 22: Are you sure this is the correct approach? It works, but would this be a potential for false negatives in other images?

RotemS: Sorry, meant to close this PR - no, I don't think it's safe :) Was checking different approaches but everything I came across so far has the risk you mentioned, to replace a false positive with a false negative, which is worse. 
Closed it now

PR #527, opened 2023-08-31
fix: correctly handle tag and digest

rmutuleanu:
#### What does this PR do?

Fixes the case when the docker daemon is unavailable, and both the image digest and tag are present.

The reviewer can start with the newly added unit test to better understand the issue.

Snyk container scanning fails when the docker daemon is unavailable, and both the image tag and digest are passed. This is because of a bug in the code that extracts the image repository from the provided target image argument.

### DIFF
--- a/test/lib/analyzer/image-inspector.spec.ts
+++ b/test/lib/analyzer/image-inspector.spec.ts
@@ -40,11 +40,6 @@ describe("extractImageDetails", () => {
       imageName: "library/nginx",
       tag: "1.18",
     }}
-    ${"nginx:1.18"} | ${{
-      hostname: "registry-1.docker.io",
-      imageName: "library/nginx",
-      tag: "1.18",
-    }}
     ${"calico/cni:release-v3.14"} | ${{
       hostname: "registry-1.docker.io",
       imageName: "calico/cni",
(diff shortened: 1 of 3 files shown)


test/lib/analyzer/image-inspector.spec.ts
```diff
@@ -75,6 +75,11 @@ describe("extractImageDetails", () => {
       imageName: "foo/bar",
       tag: "sha256:8756a25c4c5e902c4fe20322cc69d510a0517b51eab630c614efbd612ed568bf",
     }}
+    ${"localhost:1234/foo/bar:alpine3.15@sha256:8756a25c4c5e902c4fe20322cc69d510a0517b51eab630c614efbd612ed568bf"} | ${{
+      hostname: "localhost:1234",
+      imageName: "foo/bar",
+      tag: "sha256:8756a25c4c5e902c4fe20322cc69d510a0517b51eab630c614efbd612ed568bf",
```
gergo-papp, line 81: comment for a future discussion:

technically this is not a `tag` - it's a `digest`. should we model this differently in the future?

gergo-papp (approved):

test/lib/analyzer/image-inspector.spec.ts
rmutuleanu: Yes, that is right. I think the terminology is not consistent everywhere.

RotemS (approved):
LGTM

PR #548, opened 2023-10-26
fix: resets last base image in multistage build when not resolveable

pecodez:
- [x] Ready for review

#### What does this PR do?
When parsing a multistage Dockerfile the last resolved base image was not getting "reset" when a subsequent stage is not resolvable; this results in an incorrect base image being returned.

The added test cases show examples of Dockerfiles that were previously being incorrectly parsed. We previously tested that non resolvable args were handled but we had no tests to check multi-stage builds where some stages were resolvable and some not. There was also a bug with aliases and not checking that an alias of alias was resolvable.

We now look for unresolved variables in the stage, then look for and attempt to resolve any aliases, and "reset" the last known stage if any of the above checks result in an unresolvable base image name for the stage. We only save the base image name after all checks are complete.

```
ARG runtime
FROM ubuntu
FROM ${runtime}
```

The above Dockerfile previously reported the base image as `ubuntu`, as that was the last resolvable stage, it now fails with the error code `BASE_IMAGE_NON_RESOLVABLE`. The logic has been updated to "reset" the base image on any stage that is no-resolvable. The base image will only be returned if the final stage is fully resolvable.

Reported to us via #ask-lumos: https://snyk.slack.com/archives/C04HJ3ZD0UC/p1696494324311409

See Ticket [https://snyksec.atlassian.net/browse/LUM-698](https://snyksec.atlassian.net/browse/LUM-698)

### DIFF
--- a/test/lib/dockerfile/index.spec.ts
+++ b/test/lib/dockerfile/index.spec.ts
@@ -54,6 +54,46 @@ describe("base image parsing", () => {
 
   it.each`
     dockerfile
+    ${"ARG A" + EOL + "FROM image" + EOL + "FROM ${A}"}
+    ${"ARG A" + EOL + "FROM image" + EOL + "FROM ${A}:1.0.0"}
+    ${"ARG A" + EOL + "FROM image" + EOL + "FROM ${A}@sha256:abcd1234"}
+    ${"ARG A" + EOL + "FROM image" + EOL + "FROM image:${A}"}
+    ${"ARG A" + EOL + "FROM image" + EOL + "FROM image@${A}"}
+  `(
+    "does not detect unresolvable final stage: $dockerfile",
+    ({ dockerfile }) => {
+      const result = getDockerfileBaseImageName(
+        DockerfileParser.parse(dockerfile),
+      );
+
+      expect(result.baseImage).toBeUndefined();
+      expect(result).toEqual({
+        error: {
+          code: DockerFileAnalysisErrorCode.BASE_IMAGE_NON_RESOLVABLE,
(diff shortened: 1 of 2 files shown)


test/lib/dockerfile/index.spec.ts
```diff
@@ -52,6 +52,45 @@ describe("base image parsing", () => {
...
+      expect(result.baseImage).toBeUndefined();
+      expect(result).toEqual({
+        error: {
+          code: DockerFileAnalysisErrorCode.BASE_IMAGE_NON_RESOLVABLE,
+        },
+      });
+    },
+  );
+
+  it.each`
```
gergo-papp, line 78: do we also have a test case for an unresolvable stage, where the final image depends on a resolvable image ( hope that makes sense)?

```dockerfile
ARG A

FROM image:tag as base

FROM ${A} # not resolved

FROM base # resolved 
```

gergo-papp (approved):
To be honest I don't get most of the [instruction-parser.ts](https://github.com/snyk/snyk-docker-plugin/pull/548/files?diff=split&w=1#diff-5b09ee74ab76764895a17dcdb48312c5db075ea2ec06cc8c12a787e6c62e85cb) part, but the tests make it pretty clear what's going on.

Only added one minor comment about an additional test

gergo-papp (approved):

test/lib/dockerfile/index.spec.ts
pecodez: Makes perfect sense. Well spotted 😄

PR #558, opened 2023-12-13
fix: keep packageurl-js at an older version

SteveShani:
- [x] Ready for review

#### What does this PR do?
Upgrading packageurl-js will break the way dep-graphs are currently generated. This PR keeps it at an older version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "snyk-docker-plugin",
-  "lockfileVersion": 1,
   "requires": true,
+  "lockfileVersion": 1,
   "dependencies": {
     "@ampproject/remapping": {
       "version": "2.2.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -42,7 +42,7 @@
     "event-loop-spinner": "^2.0.0",
     "gunzip-maybe": "^1.4.2",
     "mkdirp": "^1.0.4",
-    "packageurl-js": "^1.1.1",
+    "packageurl-js": "^1 < 1.2.1",
```
gergo-papp, line 45: why not pin it to the last good version?
```suggestion
    "packageurl-js": "1.2.0",
```

package.json
```diff
@@ -42,7 +42,7 @@
     "event-loop-spinner": "^2.0.0",
     "gunzip-maybe": "^1.4.2",
     "mkdirp": "^1.0.4",
-    "packageurl-js": "^1.1.1",
+    "packageurl-js": "1.2.0",
```
gergo-papp, line 45: you also need to update `package-lock.json`

can be done by running `npm install` (and then run `npm ci` just to make sure the lock file is in sync with the package file)

SteveShani: I did. It didn't change

SteveShani: Since this is a library, I don't think package-lock.json is relevant.

SteveShani: either way, it didn't affect packageurl-js (and it has no dependencies)

gergo-papp: do we want to allow clients of this lib to use an older version? (I don't think we need to, but if we do, then your original fix was probably better)

gergo-papp: we still have a lock file, so you could update it - that will make sure that we run local tests with the desired version

SteveShani: I don't think we care all that much about which version they use.

I'll push the lockfile anwyay

package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "snyk-docker-plugin",
```
gergo-papp, line 2: this doesn't look right, the `packageurl-js` lib did not change

SteveShani: I believe that's because the version was already 1.2.0

gergo-papp (approved):

PR #563, opened 2024-01-18
fix: ignore composer lock files that are invalid

SteveShani:
- [x] Ready for review

#### What does this PR do?

Sometimes (like in drupal 10), empty "composer.lock" files are created. This change skips over files that are invalid, such as when the "packages" property is missing.
Run a test against `drupal:fpm-alpine` and it should succeed.


test/lib/analyzer/php.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+const NUMBER_OF_INVALID_FILES_IN_DRUPAL_10 = 2;
+describe("Can create dependency tree when some files are invalid", () => {
+  it("Should succeed and return scan results", async () => {
+    const scanResults = await phpFilesToScannedProjects(
+      drupal10FilePathToContents,
+    );
+    expect(scanResults[0].facts[0].data._depPkgsList.length).toEqual(
+      Object.keys(drupal10FilePathToContents).length +
+        NUMBER_OF_INVALID_FILES_IN_DRUPAL_10,
+    );
```
gergo-papp, line 13: what exactly are we testing here? I don't have context here, but just from reading the test I have no idea why this is the expected outcome

test/fixtures/php/drupal10FilePathToContent.json
```diff
@@ -0,0 +1,100 @@
+{
```
gergo-papp, line 1: nit: could we come up with a simpler example, so we can see what the problem in this file is?

lib/analyzer/applications/php.ts
```diff
@@ -22,13 +23,25 @@ export async function phpFilesToScannedProjects(
...
+        filePathToContent[pathPair.manifest],
+        pathPair.manifest,
+        {},
+        shouldIncludeDevDependencies,
+      );
+    } catch (e) {
+      // This will skip parsing all files that error due to being malformed.
+      // If we do not do this, the entire scan will fail.
+      // Ideally, we'd like to log this, but logging does not exist in this library.
+      if (e.name === "InvalidUserInputError") {
```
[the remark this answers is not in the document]
SteveShani: I would but it doesn't look like they're not exporting this error

tsconfig.json
```diff
@@ -20,7 +20,7 @@
     "importHelpers": true,                    /* Import emit helpers from 'tslib'. */
     // "downlevelIteration": true,            /* Provide full support for iterables in 'for-of', spread, and destructuring when targeting 'ES5' or 'ES3'. */
     // "isolatedModules": true,               /* Transpile each file as a separate module (similar to 'ts.transpileModule'). */
-
+    "resolveJsonModule": true,
```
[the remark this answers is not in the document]
SteveShani: It looks like this library only has 1 tsconfig

test/lib/analyzer/php.spec.ts
SteveShani: That's fair. At first, I just had a check to see if the function throws an error, but I thought that wasn't good enough.

I decided I wanted to ensure that only the specific problematic files were skipped

SteveShani: So I'm comparing the number of files sent to the function with the number of dependencies received. They should be the same, except for the 2 problematic files.

tsconfig.json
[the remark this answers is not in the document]
gergo-papp: I guess that's not a big issue... this is compilte-time stuff only, so feel free to ignore it

test/fixtures/php/drupal10FilePathToContent.json
SteveShani: I tried to cut it down, but it didn't parse correctly. I don't think it's worth the effort

test/lib/analyzer/php.spec.ts
SteveShani: Well, the result already has 2 removed. I could change it to `result + 2 = original` as opposed to how I do it now which is `result = original - 2`

gergo-papp: that makes sense, but I think you wrote the test as `result = original + 2` (`+` instead of `-`)?

SteveShani: You are absolutely right. The test was not correct from the get go. I fixed it and now it's much simpler.

lib/analyzer/applications/php.ts
```diff
@@ -22,13 +26,25 @@ export async function phpFilesToScannedProjects(
...
+        filePathToContent[pathPair.manifest],
+        pathPair.manifest,
+        {},
+        shouldIncludeDevDependencies,
+      );
+    } catch (e) {
+      // This will skip parsing all files that error due to being malformed.
+      // If we do not do this, the entire scan will fail.
+      // Ideally, we'd like to log this, but logging does not exist in this library.
+      if (e instanceof InvalidUserInputError) {
```
rmutuleanu, line 42: Could there be other `InvalidUserInputError` that we are skipping but don't want to.

rmutuleanu (approved):

lib/analyzer/applications/php.ts
SteveShani: I spoke to Robert about this privately, and added the conclusion in the commit message.

PR #565, opened 2024-02-01
feat: add oci distribution metadata fact [OI-37]

jonnyowenpowell:
- [x] Ready for review

#### What does this PR do?

This PR adds the `ociDistributionMetadata` fact to scan results _provided_ that:
- `manifestDigest` is provided - without it, we cannot determine whether a digest in the image name is an index or a manifest
- An image identifier is present in the `imageNameAndTag` option

- Jira ticket: https://snyksec.atlassian.net/browse/OI-37

### DIFF
--- a/lib/extractor/oci-distribution-metadata.ts
+++ b/lib/extractor/oci-distribution-metadata.ts
@@ -1,0 +1,125 @@
+import { parseAll } from "@swimlane/docker-reference";
+
+export interface OCIDistributionMetadata {
+  // Must be a valid host, including port if one was used to pull the image.
+  // Max size: 255 bytes.
+  registryHost: string;
+  // Must be a valid OCI repository namespace.
+  // Max size: 2048 bytes.
+  repository: string;
+  // Must match sha256:[a-f0-9]{64} (https://github.com/opencontainers/image-spec/blob/d60099175f88c47cd379c4738d158884749ed235/descriptor.md?plain=1#L143).
+  // Max size: 64 bytes.
+  manifestDigest: string;
+  // (Optional) Must match sha256:[a-f0-9]{64} (https://github.com/opencontainers/image-spec/blob/d60099175f88c47cd379c4738d158884749ed235/descriptor.md?plain=1#L143).
+  // Max size: 64 bytes.
+  indexDigest?: string;
+  // (Optional) Must match [a-zA-Z0-9_][a-zA-Z0-9._-]{0,127} (https://github.com/opencontainers/distribution-spec/blob/3940529fe6c0a068290b27fb3cd797cf0528bed6/spec.md?plain=1#L160).
+  // Max size: 127 bytes.
+  imageTag?: string;
+  // Must be a valid OCI Image architecure/os string (https://github.com/opencontainers/image-spec/blob/93f6e65855a1e046b42afbad0ad129a3d44dc971/config.md?plain=1#L103).
+  // Max size: 64 bytes.
+  platform: string;
+}
+
+interface OCIDistributionMetadataConstructorInput {
+  imageName: string;
+  manifestDigest: string;
+  indexDigest?: string;
+  platform: string;
+}
+
+export function constructOCIDisributionMetadata({
+  imageName,
+  manifestDigest,
+  indexDigest,
+  platform,
+}: OCIDistributionMetadataConstructorInput):
+  | OCIDistributionMetadata
+  | undefined {
+  try {
+    const ref = parseAll(imageName);
+    if (!ref.domain || !ref.repository) {
+      return;
+    }
+
+    const metadata: OCIDistributionMetadata = {
+      registryHost: ref.domain,
+      repository: ref.repository,
+      manifestDigest,
+      indexDigest,
+      imageTag: ref.tag,
+      platform,
+    };
+
+    if (!ociDistributionMetadataIsValid(metadata)) {
+      return;
+    }
+
+    return metadata;
+  } catch {
+    return;
+  }
+}
+
+function ociDistributionMetadataIsValid(
+  data: OCIDistributionMetadata,
+): boolean {
+  // 255 byte limit is enforced by RFC 1035.
+  if (Buffer.byteLength(data.registryHost) > 255) {
+    return false;
+  }
+
+  // 2048 byte limit is enforced by Snyk for platform stability.
+  // Longer strings may be valid, but nothing close to this limit has been observed by Snyk at time of writing.
+  if (
+    Buffer.byteLength(data.repository) > 2048 ||
+    !repositoryNameIsValid(data.repository)
+  ) {
+    return false;
+  }
+
+  if (!digestIsValid(data.manifestDigest)) {
+    return false;
+  }
+
+  if (data.indexDigest && !digestIsValid(data.indexDigest)) {
+    return false;
+  }
+
+  if (data.imageTag && !tagIsValid(data.imageTag)) {
+    return false;
+  }
+
+  // 64 byte limit is enforced by Snyk for platform stability.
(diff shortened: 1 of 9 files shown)


lib/extractor/oci-distribution-metadata.ts
```diff
@@ -0,0 +1,122 @@
...
+  imageName: string,
+  manifestDigest: string,
+  indexDigest?: string,
+  platform: string,
+};
+
+export function constructOCIDisributionMetadata({ imageName, manifestDigest, indexDigest, platform }: OCIDistributionMetadataConstructorInput): OCIDistributionMetadata | undefined {  
+  let registryHost: string, repositoryAndReference: string;
+  let hostOmitted = false;
+  const imageNameSplit = imageName.split('/', 2);
```
kat1906, line 34: is it possible to have more than 1 `/` in `imageName`?

lib/extractor/oci-distribution-metadata.ts
```diff
@@ -0,0 +1,122 @@
...
+    [repository] = repositoryAndReference.split('@', 2);
+  } else if (repositoryAndReference.includes(':')) {
+    [repository, imageTag] = repositoryAndReference.split(':', 2);
+  } else {
+    imageTag = 'latest';
+    [repository] = repositoryAndReference;
+  }  
+
+  if (hostOmitted && !repository.includes('/')) {
+    repository = 'library/' + repository;
```
kat1906, line 54: what does `library` mean here?

lib/extractor/oci-distribution-metadata.ts
jonnyowenpowell: Yes definitely; although when I come to write tests I'd have noticed that `split(c, 2)` doesn't do what I thought it did - the intention is to split on the first `/`.

lib/extractor/oci-distribution-metadata.ts
jonnyowenpowell: When you docker pull short names like `ubuntu`, you're really asking for `docker.io/library/ubuntu`.

lib/extractor/oci-distribution-metadata.ts
kat1906: if I am reading the regex in `repositoryNameIsValid` correctly I believe this is possible? if so, we are filling the array with the first 2 items then discarding the rest of the string

jonnyowenpowell: It seems we agree 😁

lib/image-type.ts
```diff
@@ -15,6 +15,15 @@ export function getImageType(targetImage: string): ImageType {
   }
 }
 
+export function getImageIdentifier(target: string): string | undefined {
+  const targetSplit = target.split(":");
+  if (targetSplit[0] === "docker-archive" || targetSplit[0] === "oci-archive") {
```
kat1906, line 20: might be good to have some comments, for example here, to clarify why we are looking for these strings and excluding them (I think in this case it is because of requests coming from the CLI? so need knowledge of this if there is no comment), wdyt?

jonnyowenpowell: Yeah agreed comments would be good - I'll add some before I un-draft.

kat1906:
I just noticed we use this: https://github.com/snyk/snyk-docker-plugin/blob/main/components/common.yaml for the definition of a scan result in (at least) kubernetes upstream so will have to add the new fact type there too 🙏

lib/static.ts
```diff
@@ -53,11 +57,21 @@ export async function analyzeStatically(
   const excludeBaseImageVulns = isTrue(options["exclude-base-image-vulns"]);
 
   const names = getImageNames(options, imageName);
+  let ociDistributionMetadata: OCIDistributionMetadata | undefined;
+  if (options.imageNameAndTag && options.digests?.manifest) {
+    ociDistributionMetadata = constructOCIDisributionMetadata({
+      imageName: options.imageNameAndTag,
+      manifestDigest: options.digests.manifest,
+      indexDigest: options.digests.index,
+      platform: options.platform ?? "linux/amd64",
```
kat1906, line 66: is this because the default platform pulled is `linux/amd64`?

test/lib/extractor/oci-distribution-metadata.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+        indexDigest: undefined,
+        manifestDigest:
+          "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
+        platform: "linux/arm64",
+        registryHost: "gcr.io",
+        repository: "example/repo",
+      },
+    ],
+    [
+      "given an image name with a digest it is not included in the result",
```
kat1906, line 77: nit:
```suggestion
      "given an image name with a digest imageTag is not included in the result",
```

test/lib/extractor/oci-distribution-metadata.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+        indexDigest: undefined,
+        manifestDigest:
+          "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
+        platform: "linux/amd64",
+        registryHost: "docker.io",
+        repository: "example/repo",
+      },
+    ],
+    [
+      "given an image name without a namespace include the default in the result",
```
kat1906, line 114: nit:
```suggestion
      "given an image name without a namespace include the default repository in the result",
```

test/lib/extractor/oci-distribution-metadata.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+          "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
+        platform: "linux/amd64",
+      },
+      undefined,
+    ],
+    [
+      "given an image name with a namespace that is too long should return undefined",
+      {
+        // 2048 is the maxmimum length, there are 2049 'a's here:
+        imageName:
```
kat1906, line 147: nit: to make the test read nicely you could use `'gcr.io/' + 'a'.repeat(2049)`

test/lib/extractor/oci-distribution-metadata.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+        imageName: "gcr.io/example/repo:test",
+        indexDigest: "sha256:abc",
+        manifestDigest:
+          "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
+        platform: "linux/amd64",
+      },
+      undefined,
+    ],
+    [
+      "given an invalid platform should return undefined",
```
kat1906, line 218: should we add some logging to detail why undefined is returned seeing as there are a number of cases for this result?

lib/static.ts
jonnyowenpowell: Yes, we need to be careful when we consume this library though to set platform appropriately - we don't want to pull something exotic and then have this default to `"linux/amd64"`. I can't really avoid a default here though without a breaking change to the API which doesn't feel necessary.

pecodez: iirc the default is set to the arch of the machine the scan runs on, which is why Apple chip users need to set `--platform=linux/amd64` when running a test/monitor. This is done in the CLI and passed as an option to `scan()` I think. Be careful if this overrides that behaviour as I would class that as a breaking change.

kat1906 (approved):

pecodez (approved):
LGTM

test/lib/extractor/oci-distribution-metadata.spec.ts
jonnyowenpowell: Declining to add logging here because this is a library and we don't have a good mechanism to support BYOL.

PR #568, opened 2024-02-14
fix: add support for deleted folders

SteveShani:
When a file or folder is deleted, an entry that starts with .wh is created. Up until now, Snyk detected files that were deleted, but not folders.

This change adds support for deleted folders by checking whether a file path starts with the path of the deleted folder. Meaning if we know that "/app/dependencies" was deleted, then a file path like "/app/dependencies/jar/library.jar" will be marked as deleted as well.

- [x] Ready for review

To test this change, create an image from a dockerfile that:
* creates a folder
* downloads / copies a file that is recognized by Snyk (like `.jar` or `package-lock.json`) to that folder
* deletes the folder

### DIFF
--- a/lib/extractor/index.ts
+++ b/lib/extractor/index.ts
@@ -1,3 +1,4 @@
+import path = require("path");
 import {
   getLayersFromPackages,
   getPackagesFromRunInstructions,
@@ -198,7 +199,16 @@ function layersWithLatestFileModifications(
       // and removing it from the set since it can be found in consecutive layers
       if (removedFilesToIgnore.has(filename)) {
         removedFilesToIgnore.delete(filename);
-      } else if (
+        continue;
+      }
+      if (
+        Array.from(removedFilesToIgnore).some((removedFile) =>
+          isFileInFolder(removedFile, filename),
+        )
+      ) {
+        continue;
+      }
+      if (
         // file was not found + avoid adding deleted files with .wh.
         !Reflect.has(extractedLayers, filename) &&
         !isDeletedFile(filename)
@@ -254,4 +264,11 @@ function getContent(
   return firstFileNameMatch !== undefined
     ? extractedLayers[firstFileNameMatch][extractAction.actionName]
     : undefined;
+}
+
+function isFileInFolder(folder: string, file: string): boolean {
+  const folderPath = path.normalize(folder);
+  const filePath = path.normalize(file);
+
+  return filePath.startsWith(path.join(folderPath, path.sep));
 }
--- a/test/system/deleted-files.spec.ts
+++ b/test/system/deleted-files.spec.ts
@@ -65,3 +65,78 @@ describe("deleted files", () => {
     ).toEqual(1);
   });
 });
+
+describe("Deleted Folders", () => {
+  const getFixture = (fixturePath) =>
+    path.join(__dirname, "../fixtures/docker-archives", fixturePath);
+
+  /*
+    The image in the example below was built with the following Dockerfile:
+
+    FROM alpine:latest
+    WORKDIR /app
+    RUN mkdir jar_folder && wget https://repo1.maven.org/maven2/commons-collections/commons-collections/3.2.1/commons-collections-3.2.1.jar && mv commons-collections-3.2.1.jar jar_folder
+    RUN rm -rf /app/jar_folder
+    RUN echo "hello"
+  */
+  it("commons-collections should be excluded from dependency tree", async () => {
+    const fixturePath = getFixture("docker-save/deleted-folder.tar");
+    const imagePath = `docker-archive:${fixturePath}`;
+
+    const pluginResult = await plugin.scan({
+      path: imagePath,
+    });
+
+    // There should only be the distro scan result. The maven scan result should not exist.
+    expect(pluginResult.scanResults.length).toEqual(1);
+  });
+
+  /*
+    The image in the example below was built with the following Dockerfile:
+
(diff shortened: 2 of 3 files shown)


lib/extractor/index.ts
```diff
@@ -196,7 +196,12 @@ function layersWithLatestFileModifications(
 
       // not adding the original file to extractedLayers
       // and removing it from the set since it can be found in consecutive layers
-      if (removedFilesToIgnore.has(filename)) {
+      if (
+        removedFilesToIgnore.has(filename) ||
+        Array.from(removedFilesToIgnore).some((removedFile) =>
+          filename.startsWith(removedFile),
+        )
```
rmutuleanu, line 203: What happens if there are multiple files in the deleted directory? 

It looks like this would work only for the first file since we have `removedFilesToIgnore.delete(filename);` when handling it.

rmutuleanu, line 203: Should we still try to remove the `filename` from the `removedFilesToIgnore` set in case the directory was removed and not the file (basically if the if condition is `true` because of the second part)?

lib/extractor/index.ts
```diff
@@ -196,7 +196,12 @@ function layersWithLatestFileModifications(
 
       // not adding the original file to extractedLayers
       // and removing it from the set since it can be found in consecutive layers
-      if (removedFilesToIgnore.has(filename)) {
+      if (
+        removedFilesToIgnore.has(filename) ||
+        Array.from(removedFilesToIgnore).some((removedFile) =>
```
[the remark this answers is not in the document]
SteveShani: That's exactly the type of edge cases I want brought up to feel comfortable with this change.

This is an edge case. In case both the folder and file are in the same directory this would happen.
Since `/app/abcEfg` start with `/app/abc`. I'll make a more strict comparison

lib/extractor/index.ts
SteveShani: The point was "if the file was in a directory that was removed, remove that file". I do see an issue now though, where where if the folder was created again in the next layer, it won't be registered

SteveShani: The issue that I thought about turns out is handled by the fact that we process the layers from bottom to top. I added a test to showcase this

lib/extractor/index.ts
```diff
@@ -196,7 +197,14 @@ function layersWithLatestFileModifications(
...
+          isFileInFolder(removedFile, filename),
+        )
+      ) {
+        // If the file is in a deleted folder, this next line won't work.
```
rmutuleanu, line 206: Could we have a separate condition for files within deleted folders so that it is clear that we don't want to run `removedFilesToIgnore.delete(filename);` for those cases?

lib/extractor/index.ts
```diff
@@ -196,7 +197,14 @@ function layersWithLatestFileModifications(
 
       // not adding the original file to extractedLayers
       // and removing it from the set since it can be found in consecutive layers
-      if (removedFilesToIgnore.has(filename)) {
+      if (
+        removedFilesToIgnore.has(filename) ||
```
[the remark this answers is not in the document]
SteveShani: > Would it make sense to have a separate set for folders instead of using removedFilesToIgnore?
Ideally yes, but there is no simple way to know whether a file or folder was deleted at the point where we add it to the set.

I didn't find any other references to that set in the code, so I assume not.

lib/extractor/index.ts
SteveShani: I edited it to include `continue` statements instead of if-else chains

test/system/deleted-files.spec.ts
```diff
@@ -65,3 +65,78 @@ describe("deleted files", () => {
     ).toEqual(1);
   });
 });
+
+describe("Deleted Folders", () => {
+  const getFixture = (fixturePath) =>
+    path.join(__dirname, "../fixtures/docker-archives", fixturePath);
+
+  /*
+    The image in the example below was built with the following Dockerfile:
```
gergo-papp, line 74: this is unrelated to your change, but I think it's important:

could we submit the Dockerfiles to the test fixtures and maybe even generate the image tars on-demand? It would make these tests a lot more maintainable. Right now re rely on comments to understand how an image tar was generated

SteveShani: I do think it's a little out of scope.

Regarding generating the test fixtures on demand, that's not a bad idea. It would require adding docker as a dependency though. Do you think that's worth it?

gergo-papp (approved):

tommyknows (approved):
To me this makes sense, seems like a simple oversight when we initially implemented this....

test/system/deleted-files.spec.ts
tommyknows: i agree that this makes sense, and is partially already done for example for the [Go binary tests](https://github.com/snyk/snyk-docker-plugin/blob/fcb30861bbc2075bf364abaa719ea1163cd7bdff/test/fixtures/go-binaries/source/build.sh#L1).

(The test itself runs nightly and generates fresh Go binaries based on `go:latest` to ensure new releases don't break: https://github.com/snyk/snyk-docker-plugin/blob/fcb30861bbc2075bf364abaa719ea1163cd7bdff/.circleci/config.yml#L209)

gergo-papp: this would obviously not be part of this PR, but I think it's still worth doing... maybe by the next team

PR #569, opened 2024-02-15
fix: adds large file support to bufferToSha1 method

pecodez:
- [ ] Ready for review

#### What does this PR do?

Adds large file support to `bufferToSha1` method.

The `bufferToSha1` method was failing as the `update` method of the `Crypto.Hash` object uses `Buffer.alloc` which has a limitation of 2G. Any JARs bigger than this would cause a `RangeError` in `Node`.

This fix uses the stream interface of `Crypto.Hash` to stream data for large files in order to generate the Sha1 digest.

Scan any container with a Maven project that contains one or more JAR over 2G in size.

https://snyksec.atlassian.net/browse/SUP-1997

### DIFF
--- a/lib/analyzer/applications/java.ts
+++ b/lib/analyzer/applications/java.ts
@@ -43,7 +43,7 @@ export async function jarFilesToScannedResults(
       continue;
     }
 
-    const fingerprints = getFingerprints(
+    const fingerprints = await getFingerprints(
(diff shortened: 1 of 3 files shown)


ChristinaDara (approved):

jonnyowenpowell (approved):
LGTM!

PR #571, opened 2024-02-21
fix: uses options.platform to select correct manifest from OCI layout

pecodez:
- [ ] Ready for review

#### What does this PR do?

Adds support for multiple platforms of OCI archives; previously restricted to `linux/amd64` only.

When selecting the correct manifest from an OCI layout, use the `--platform` option to match against an index or fallback to the only available manifest in the case of single manifest existing in the archive. This ensures multi-platform images saved with `containerd` get the correct manifest/config and sets the `platform` correctly from the associated config. This requires a default to be set, which I think makes the most sense to use `/linux/amd64`.

Test suites added to include fixtures for OCI archives generated with docker post 25.x and `containerd` as they differ very slightly in manifest properties.

test/lib/extractor/extractor.spec.ts

This has become an issue because Docker Engine 25.x now generates an OCI layout when using `docker save` when it previously did not. The OCI Extractor had a limitation to only support the `linux/amd64` platform.

It was also discovered that the OCI Extractor was never setup to correctly handle multi-platform images. These images have multiple manifests and configs, as well as an index, and there was nothing to handle selecting the correct manifests, especially when platform architecture variants existed. The behaviour was always to use the default manifest, which is incorrect.

#### Use Case
A multi platform image that has manifests for `linux/arm/v5`, `linux/arm/v6` and `linux/arm/v7` will not be able to determine the correct manifest and config if `--platform=linux/arm` is supplied.

However, using `--platform/arm64` will currently work as the only variant is `linux/arm64/v8` and so can be selected without the need to explicitly use the variant.

https://snyksec.atlassian.net/browse/SUP-2577
https://snyksec.atlassian.net/browse/SUP-2579
https://snyksec.atlassian.net/browse/SUP-2594
https://snyksec.atlassian.net/browse/SUP-2595
https://snyksec.atlassian.net/browse/SUP-2630
https://snyksec.atlassian.net/browse/SUP-2644
https://snyksec.atlassian.net/browse/SUP-2665


lib/extractor/oci-archive/layer.ts
```diff
@@ -30,18 +31,20 @@ const MEDIATYPE_OCI_MANIFEST_LIST_V1 =
...
   extractActions: ExtractAction[],
+  platform: string,
 ): Promise<ExtractedLayersAndManifest> {
   return new Promise((resolve, reject) => {
     const tarExtractor: Extract = extract();
 
     const layers: Record<string, ExtractedLayers> = {};
     const manifests: Record<string, OciArchiveManifest> = {};
-    let imageConfig: ImageConfig | undefined;
+    const configs: ImageConfig[] = [];
```
pecodez, line 47: Multi-platform images have a config for each platform included in the image.
Note: The index file will always list all platforms supported even if not included in the image

lib/extractor/oci-archive/layer.ts
```diff
@@ -220,3 +231,40 @@ function isImageIndexFile(json: any): boolean {
...
 }
+
+function getBestMatchForPlatform<
+  T extends {
+    platform?: OciPlatformInfo;
+    os?: string;
+    architecture?: string;
+    variant?: string;
+  },
+>(manifests: T[], platform: string): T | undefined {
```
pecodez, line 242: I'm not super happy with this type. The function handles both image manifests and configs and uses the exact same logic to compare platform info, however the shape of manifests and configs is different so I used a generic and extended it to include the different options.

lib/scan.ts
```diff
@@ -19,6 +19,17 @@ export function mergeEnvVarsIntoCredentials(
   options.password = options.password || process.env.SNYK_REGISTRY_PASSWORD;
 }
 
+function sanitizePlatform(options: Partial<PluginOptions>): void {
+  if (options.platform === undefined) {
+    options.platform = "linux/amd64";
+  } else if (
+    !options.platform.startsWith("linux/") ||
```
jonnyowenpowell, line 26: It might be confusing to users that non-linux platforms are rejected as invalid - perhaps the error could, in this case, read that the platform is _unsupported_?

jonnyowenpowell (approved):
LGTM! I've left a few non-blocking comments.

lib/scan.ts
pecodez: Agreed 😄

test/lib/docker.spec.ts
```diff
@@ -51,7 +51,7 @@ describe("docker", () => {
       const result = docker.save(image, destination);
 
       //  rejects with expected error
-      await expect(result).rejects.toThrowError("server error");
+      await expect(result).rejects.toThrowError();
```
pecodez, line 54: Different Docker engines are returning different error strings, for now we will just expect an exception to be thrown and the Promise rejected

.circleci/config.yml
```diff
@@ -142,24 +142,16 @@ jobs:
...
-    <<: *defaults
-    resource_class: medium
-    steps:
-      - checkout
-      - setup_remote_docker
-      - run:
-          name: Use snyk-main npmjs user
-          command: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" >> .npmrc
-      - run: npm ci
-      - run: npm run test-jest
```
pecodez, line 162: unused

pecodez:
Due to breaking changes across Docker engines tests using snapshots are proving flaky as layer digests and file checksums are different even though tests pass after successful scans. Removing snapshots and testing explicitly for expected properties on failing tests.

lib/extractor/oci-archive/layer.ts
pecodez: Sorry, I was not very clear here! When an image has a manifest of type `application/vnd.oci.image.index.v1+json` it has all the available platforms

PR #574, opened 2024-03-04
fix: upgrade eol'd snyk_docker_plugin version - it

jac-moran:
- [ ] Ready for review

#### What does this PR do?
While building registry main on ciecleCI, I found a persistent error, which points to an EOD dependency in `setup_remote_docker` (see [here](https://discuss.circleci.com/t/remote-docker-image-deprecations-and-eol-for-2024/50176)).

Tracing the dependency, I reached this as a registry dependency where one of the EOD'd versions (20.10.23) is being used. This needs to be updated, but please let me know if a different version should be used here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,7 +145,7 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker:
-          version: 20.10.23
+          version: default
       - attach_workspace:
           at: ~/snyk-docker-plugin
       - run: npm run test-jest


gergo-papp (approved):
Looks good but I'd wait for @pecodez's review too. I know there might have been some issues around these versions

pecodez (commented):
While it should be safe to unpin this as I have tested the plugin against a number of different Docker engine versions, specifically the fix for supporting OCI images generated by Docker engine 25.x, I think our issue will the number of tests we have using Jest snapshots to match layer hashes. I've seen these change completely across different Docker engines, with & without using `containerd` under the hood (Docker Desktop).

pecodez:
The failing test will need addressing, its an example of layer hashes not matching after running `docker save` with a newer engine I believe. Updating the snapshot may be enough here.

PR #595, opened 2024-06-21
feat: support platform when pulling image from registry UNIFY-163

adrobuta:
- [x] Ready for review

#### What does this PR do?

Adds support for passing the platform parameter when pulling an image from a registry without using the docker binary

### DIFF
--- a/lib/docker.ts
+++ b/lib/docker.ts
@@ -41,12 +41,30 @@ class Docker {
     imageSavePath: string,
     username?: string,
     password?: string,
+    platform?: string,
   ): Promise<DockerPullResult> {
     const dockerPull = new DockerPull();
+    const platformTokens = platform ? platform.split("/") : undefined;
+    let os: string = "linux";
+    let architecture: string = "amd64";
+    let variant: string | undefined;
+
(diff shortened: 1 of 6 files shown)


lib/docker.ts
```diff
@@ -41,12 +41,16 @@ class Docker {
     imageSavePath: string,
     username?: string,
     password?: string,
+    platform?: string | undefined,
```
danlucian, line 44:
```suggestion
    platform?: string,
```

`undefined` is redundant here as the parameter is already marked as optional with `?`.

test/lib/docker.spec.ts
```diff
@@ -93,6 +95,35 @@ describe("docker", () => {
       });
     }
 
+    test("Pass platform argument when pulling from registry", async () => {
```
danlucian, line 98: Should we move this test to its own `describe` block since the current one relates to the `docker.save(...)` command?

adrobuta: great catch. thanks for the suggestion!

danlucian (approved):

PR #596, opened 2024-06-23
fix: use custom version compare function to support multiple version string compare groups

SteveShani:
This fixes an issue where an exception would occur when a python dependency had a version that did not conform to semver specs, such as 1.18.3.9

- [x] Ready for review

### DIFF
--- a/lib/python-parser/common.ts
+++ b/lib/python-parser/common.ts
@@ -16,3 +16,42 @@ export function specifierValidRange(
   const versionPartsLength = version.split(".").length;
   return versionPartsLength === 2 ? "^" : "~";
 }
+
+// This regex is used to extract the "semver" part from the version string.
+// See the tests for a better understanding (Also the reason why this is exported)
+export const VERSION_EXTRACTION_REGEX = /(?<VERSION>(\d+\.)*\d+).*/;
+
+function compareArrays(v1: number[], v2: number[]) {
+  const max = v1.length > v2.length ? v1.length : v2.length;
+  for (let i = 0; i < max; i++) {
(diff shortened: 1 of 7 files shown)


adrobuta (approved):

lib/python-parser/common.ts
```diff
@@ -16,3 +16,38 @@ export function specifierValidRange(
...
+  }
+
+  const v1 = v1Match
+    .groups!.VERSION.split(".")
+    .map((part) => Number.parseInt(part, 10));
+  const v2 = v2Match
+    .groups!.VERSION.split(".")
+    .map((part) => Number.parseInt(part, 10));
+
+  const max = v1.length > v2.length ? v1.length : v2.length;
```
danlucian, line 43: I would extract the following logic to its function, aiming for a small readability win:

```
function compareArrays(v1: any[], v2: any[]): number {
  const maxLength = Math.max(v1.length, v2.length);

  for (let i = 0; i < maxLength; i++) {
    const element1 = v1[i] || 0;
    const element2 = v2[i] || 0;

    if (element1 < element2) {
      return 1;
    } else if (element1 > element2) {
      return -1;
    }
  }

  return 0;
}
```

The `any` type should be concrete, just used it for the sake of the example.

What do you say?

SteveShani: I can do that. To me the function is readable but that's subjective.

Regarding the concrete type, the right answer would be `Array[Comparable]` but I don't think Typescript has a `Comparable` type. I'll stick with `number` or `string` for now as that's what's likely to appear.

PR #603, opened 2024-09-10
fix: correctly detect files from removed layers UNIFY-243

stefan-citiriga-snyk:
- [ ] Ready for review

#### What does this PR do?

The customer noticed that we are still reporting vulnerabilities from jars that were deleted.

The issue was that in the extracted layers we could find multiple additions of the exact same file. Docker allows the addition of exactly the same file in exaclty the same path multiple times. We parse layers in reverse chronological order, so encountering removals first means those files are 100% removed. For clarifying, removals are marked with ".wh.".
Before this change, we would only take into account a removal for the first instance of that exact same file in the following layers. 

"Getting into" a container created using this image (using docker debug) the existence of these files can be checked. No matter in how many layers we added that one exact same file, one remove is enough to make sure the file cannot be found anymore.

In this comit I have also added 2 tests and 2 lightweight fixtures for those tests.

This also includes a light refactoring, to make the code easier to understand.

lib/extractor/index.ts

1. Have the CLI and snyk-docker-plugin projects in the same folder.
2. Be on this branch in snyk-docker-plugin
3. In CLI/package.json, change the version of snyk-docker-plugin to be "file:../snyk-docker-plugin"
4. Remove node_modules and dist folder and run npm install and npm run build (for the CLI project)
5. For the behaviour after the fix, run:node dist/cli/index.js auth
 then run node dist/cli/index.js container monitor removed-jars-image:latest
  where removed-jars-image is build from this dockerfile:
FROM alfresco/process-services:2.3.9
RUN rm /usr/local/tomcat/webapps/activiti-app/WEB-INF/lib/*
RUN rm /usr/local/tomcat/lib/*

You should be able to find this:
Screenshot 2024-09-10 at 10 12 36

This is what the customer expects and what should happen.

And if you now run snyk container monitor removed-jars-image:latest, you would have the previously seen outcome overridden with this:
image

This last screenshot shows what happened before the change and the removed jars are not taken into account. This is the behaviour that changed due to this fix.

https://snyksec.atlassian.net/browse/UNIFY-243

### DIFF
--- a/lib/extractor/index.ts
+++ b/lib/extractor/index.ts
@@ -203,28 +203,21 @@ function layersWithLatestFileModifications(
     // go over extracted files products found in this layer
     for (const filename of Object.keys(layer)) {
       // if finding a deleted file - trimming to its original file name for excluding it from extractedLayers
-      if (isDeletedFile(filename)) {
+      // + not adding this file
+      if (isWhitedOutFile(filename)) {
         removedFilesToIgnore.add(filename.replace(/.wh./, ""));
+        continue;
       }
-
-      // not adding the original file to extractedLayers
-      // and removing it from the set since it can be found in consecutive layers
+      // not adding previously found to be whited out files to extractedLayers
       if (removedFilesToIgnore.has(filename)) {
-        removedFilesToIgnore.delete(filename);
         continue;
       }
-      if (
-        Array.from(removedFilesToIgnore).some((removedFile) =>
-          isFileInFolder(removedFile, filename),
-        )
-      ) {
+      // not adding path that has removed path as parent
+      if (isFileInARemovedFolder(filename, removedFilesToIgnore)) {
         continue;
       }
-      if (
-        // file was not found + avoid adding deleted files with .wh.
-        !Reflect.has(extractedLayers, filename) &&
-        !isDeletedFile(filename)
-      ) {
+      // file not already in extractedLayers
+      if (!Reflect.has(extractedLayers, filename)) {
         extractedLayers[filename] = layer[filename];
       }
     }
@@ -232,7 +225,7 @@ function layersWithLatestFileModifications(
   return extractedLayers;
 }
 
-export function isDeletedFile(filename: string) {
+export function isWhitedOutFile(filename: string) {
   return filename.match(/.wh./gm);
 }
 
@@ -278,9 +271,18 @@ function getContent(
     : undefined;
 }
 
-function isFileInFolder(folder: string, file: string): boolean {
+function isFileInFolder(file: string, folder: string): boolean {
   const folderPath = path.normalize(folder);
   const filePath = path.normalize(file);
 
   return filePath.startsWith(path.join(folderPath, path.sep));
+}
+
+function isFileInARemovedFolder(
+  filename: string,
+  removedFilesToIgnore: Set<string>,
+): boolean {
+  return Array.from(removedFilesToIgnore).some((removedFile) =>
+    isFileInFolder(filename, removedFile),
+  );
 }
--- a/test/system/deleted-files.spec.ts
+++ b/test/system/deleted-files.spec.ts
@@ -64,6 +64,30 @@ describe("deleted files", () => {
         .length,
     ).toEqual(1);
   });
+
+  /*
+    The following refers to the case where in several layers the exact same files are added in the same exact path.
+    The image in the example below was built with the following Dockerfile:
+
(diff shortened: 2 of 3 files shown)


lib/extractor/index.ts
```diff
@@ -203,27 +203,25 @@ function layersWithLatestFileModifications(
     // go over extracted files products found in this layer
     for (const filename of Object.keys(layer)) {
       // if finding a deleted file - trimming to its original file name for excluding it from extractedLayers
-      if (isDeletedFile(filename)) {
+      if (isWhitedOutFile(filename)) {
```
SteveShani, line 206: If the current file is a whited-out file, wouldn't that mean we can also `continue` as the next statements won't apply for sure?

lib/extractor/index.ts
```diff
@@ -203,27 +203,25 @@ function layersWithLatestFileModifications(
     // go over extracted files products found in this layer
     for (const filename of Object.keys(layer)) {
```
SteveShani, line 204: What do you think about having "if-else" chains here instead of `continues`?

How about extracting code to separate functions and using "return" instead?

lib/extractor/index.ts
```diff
@@ -203,27 +203,25 @@ function layersWithLatestFileModifications(
...
       }
 
-      // not adding the original file to extractedLayers
-      // and removing it from the set since it can be found in consecutive layers
+      // not adding previously found to be whited out files to extractedLayers
       if (removedFilesToIgnore.has(filename)) {
-        removedFilesToIgnore.delete(filename);
         continue;
       }
       if (
```
SteveShani, line 214: I think you could also extract this statement into a separate function with a clear name, such as:

`function isFileInARemovedFolder`

lib/extractor/index.ts
```diff
@@ -203,27 +203,25 @@ function layersWithLatestFileModifications(
...
         )
       ) {
         continue;
       }
       if (
-        // file was not found + avoid adding deleted files with .wh.
+        // file was not found in extractedLayers + avoid adding deleted files with .wh.
         !Reflect.has(extractedLayers, filename) &&
-        !isDeletedFile(filename)
+        !isWhitedOutFile(filename)
```
SteveShani, line 224: I don't think we'd need this if we `continue` after checking if the file is whited-out

lib/extractor/index.ts
```diff
@@ -203,27 +203,25 @@ function layersWithLatestFileModifications(
...
-          isFileInFolder(removedFile, filename),
+          isFileInFolder(filename, removedFile),
         )
       ) {
         continue;
       }
       if (
-        // file was not found + avoid adding deleted files with .wh.
+        // file was not found in extractedLayers + avoid adding deleted files with .wh.
         !Reflect.has(extractedLayers, filename) &&
```
SteveShani, line 223: Did you figure out if this is necessary?

test/system/deleted-files.spec.ts
```diff
@@ -91,6 +116,31 @@ describe("Deleted Folders", () => {
...
+    FROM alpine:latest
+    WORKDIR /app
+    RUN mkdir -p /app/tmp/jar_folder
+    RUN mkdir jar_folder && wget https://repo1.maven.org/maven2/commons-collections/commons-collections/3.2.1/commons-collections-3.2.1.jar && mv commons-collections-3.2.1.jar jar_folder
+    RUN wget https://repo1.maven.org/maven2/commons-collections/commons-collections/3.2.1/commons-collections-3.2.1.jar -P /app/tmp/jar_folder
+    RUN mv /app/tmp/jar_folder/commons-collections-3.2.1.jar /app/jar_folder/
+    RUN rm -rf /app/tmp/jar_folder
+    RUN rm -rf /app/jar_folder/*
+  */
+  it("should exclude removed packages added multiple times before a removal", async () => {
```
SteveShani, line 132: It's not very clear to me at first glance what the difference between this test and the previous text is.

What is the difference between `files` and `packages`?

SteveShani:
Small nitpick, your PR title "fix: handle removal of multiple jar additions" isn't as accurate as it could be. This fix isn't jar specific 🙂

lib/extractor/index.ts
stefan-citiriga-snyk: Condition before "&&" is necessary, condition after "&&" is not so i will take that out.

SteveShani (approved):

PR #612, opened 2024-11-07
fix: avoid possible unhandled promise rejections

adrobuta:
- [x] Ready for review

#### What does this PR do?

Fixes the possible unhandled promise rejection edge case and adds some additional input validation.

### DIFF
--- a/lib/analyzer/applications/node-modules-utils.ts
+++ b/lib/analyzer/applications/node-modules-utils.ts
@@ -42,17 +42,25 @@ async function createSyntheticManifest(
 ): Promise<void> {
   const tempRootManifestPath = path.join(tempRootManifestDir, manifestName);
   debug(`Creating an empty synthetic manifest file: ${tempRootManifestPath}`);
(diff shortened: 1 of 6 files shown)


lib/analyzer/applications/node-modules-utils.ts
```diff
@@ -32,6 +32,10 @@ async function createTempProjectDir(
 const manifestName: string = "package.json";
 
 async function fileExists(path: string): Promise<boolean> {
+  if (!path) {
```
SteveShani, line 35: what is this checking?

SteveShani: Oh, if the string is "".

SteveShani: how come line 39-41 don't handle this edge case?

adrobuta: it does, removing the unnecessary check. thanks!

SteveShani (approved):

PR #624, opened 2024-12-01
[RNTM-746] Collect node app file hierarchy

davidgonor1408:


lib/analyzer/applications/node.ts
```diff
@@ -35,6 +40,7 @@ interface ManifestLockPathPair {
 
 export async function nodeFilesToScannedProjects(
   filePathToContent: FilePathToContent,
+  applicationFilesFlag: boolean,
```
nozik, line 43: Please rename to `collectApplicationFiles`

lib/facts.ts
```diff
@@ -62,6 +62,11 @@ export interface TestedFilesFact {
   data: string[];
 }
 
+export interface ApplicationFilesFact {
+  type: "applicationFiles";
+  data: string[];
```
nozik, line 67: Create an interface like `ApplicationFileInfo` to allow more flexibility across different ecosystems

lib/inputs/node/static.ts
```diff
@@ -2,12 +2,25 @@ import { basename } from "path";
 import { ExtractAction } from "../../extractor/types";
 import { streamToString } from "../../stream-utils";
 
-const nodeAppFiles = ["package.json", "package-lock.json", "yarn.lock"];
-const deletedAppFiles = nodeAppFiles.map((file) => ".wh." + file);
+const nodeAppFilePatterns = [
+  /^package\.json$/,
```
nozik, line 6: Why not check explicitly if the files are package.json / package-lock.json / etc.? Why do we need the regex?

test/system/application-scans/node.spec.ts
```diff
@@ -619,13 +621,48 @@ describe("node application files grouping", () => {
...
+      "/srv/dist/README.md",
+    ];
+    const fileByDirGroups = nodeUtils.groupFilesByDirectory(nodeAppFiles);
+
+    const [appFilesRootDir, appFiles] =
+      nodeUtils.filterAppFiles(fileByDirGroups);
+
+    expect(appFilesRootDir).toBe("/srv/dist");
+    expect(appFiles.length).toBe(8);
+    expect(appFiles).toEqual([
```
nozik, line 648: Don't we want to remove the commit prefix before reporting the data?

PR #637, opened 2025-01-02
fix: common extracted dir bug

davidgonor1408:


lib/analyzer/applications/runtime-common.ts
```diff
@@ -37,6 +49,9 @@ export function getAppFilesRootDir(
     }
   });
 
+  if (rootDir === "") {
```
nozik, line 52: Why do we need this `if`? Can't it be handled more elegantly?

nozik (approved):

SteveShani (approved):

PR #641, opened 2025-01-23
fix: change signature and pull node and python data

prsnca:


test/system/application-scans/node.spec.ts
```diff
@@ -216,21 +217,18 @@ describe("node application scans", () => {
...
-          fileName.endsWith("/package-lock.json") ||
-          fileName.endsWith("/yarn.lock"),
-      ).toBeTruthy();
-      expect("node-app-files" in result.extractedLayers[fileName]).toBeTruthy();
-    });
+    const applicationFilesFact: ApplicationFilesFact = result.scanResults[0].facts.find(
+        (fact) => fact.type === "applicationFiles",
+      )!.data;
+
+    const fileList = applicationFilesFact.data[0].fileHierarchy.map((fh) => fh.path);
```
nozik, line 230: Why map if you're only counting?

prsnca: this is the model:
```
export interface ApplicationFileInfo {
  path: string;
}
export interface ApplicationFiles {
  fileHierarchy: ApplicationFileInfo[];
  moduleName?: string;
  language: string;
}
```
so in order to pull out the list of files - i have to extract each path by its own...

test/system/application-scans/node.spec.ts
```diff
@@ -216,21 +216,19 @@ describe("node application scans", () => {
...
-      ).toBeTruthy();
-      expect("node-app-files" in result.extractedLayers[fileName]).toBeTruthy();
-    });
+    const applicationFilesFact: ApplicationFilesFact =
+      result.scanResults[0].facts.find(
+        (fact) => fact.type === "applicationFiles",
+      )!.data;
+
+    const fileList = applicationFilesFact[0].fileHierarchy.map((fh) => fh.path);
+    expect(fileList.length).toEqual(4);
```
nozik, line 231: Please validate on the specific file names

prsnca: Done ✌️

nozik (approved):

PR #642, opened 2025-02-01
feat: add kaniko archive scan support

aarlaud:
- [x] Ready for review

#### What does this PR do?

Adds Kaniko image support.

Essentially, simply dealing with the extraction part, the rest just worked.


test/lib/extractor/extractor.spec.ts
```diff
@@ -80,6 +80,27 @@ describe("extractImageContent", () => {
     });
   });
 
+  describe("Kaniko Image Archives (expected to fall back to DockerExtractor)", () => {
```
robguinness-snyk, line 83: What do you mean by "expected to fall back to DockerExtractor"?

test/lib/image-type.spec.ts
```diff
@@ -29,6 +29,15 @@ describe("image-type", () => {
 
       expect(result).toEqual(expectedImageType);
     });
+
+    test("should return oci-archive type given oci-archive image", () => {
```
robguinness-snyk, line 33: should return kaniko-archive?

test/lib/extractor/extractor.spec.ts
aarlaud: cleaning this up, from the test case copy above

test/lib/image-type.spec.ts
aarlaud: same, cleaning up

robguinness-snyk (approved):

PR #652, opened 2025-03-31
feat: support scan with unspecified archive type

adrobuta:
- [x] Ready for review

#### What does this PR do?
 Adds support for scanning archives without specifying the image type. This implies finding the absolute path of the archive file specified and fallback on different type of extractors when decompressing the archive.

The change can be tested by providing only the .tar filename without specifying docker-archive or oci-archive prefix.
`snyk container test archive.tar
`

Snyk can directly scan or monitor a Docker or OCI archive by running the following commands: snyk container test docker-archive:<filename>.tar or snyk container test oci-archive:<filename>.tar.

Examples:
```
snyk container test docker-archive:archive.tar
snyk container test oci-archive:archive.tar
snyk container test kaniko-archive:archive.tar
```
Manually defining the image type adds complexity, increases the risk of errors, and may not account for dynamic updates of the pipelines where the customer deploy snyk.


lib/extractor/index.ts
```diff
@@ -152,6 +149,25 @@ export async function extractImageContent(
...
+    try {
+      return [await extractor.getLayersAndManifest(), extractor];
+    } catch (error) {
+      errorQueue.push(
+        `Fallback extraction failed using ${imageType} extractor`,
+      );
+    }
+  }
+
+  throw new InvalidArchiveError(
```
SteveShani, line 166: I'm not sure we need all of the error messages here. It's implied that if this error is thrown then all extractors failed.

test/system/plugin.spec.ts
```diff
@@ -1,7 +1,54 @@
 import { DepGraph } from "@snyk/dep-graph";
 import * as plugin from "../../lib";
+import { getFixture } from "../util";
 
 describe("plugin", () => {
+  describe("image is scanned when no image type is specified", () => {
```
SteveShani, line 6: Maybe add a test that throws when all extractors fail?

lib/extractor/index.ts
adrobuta: agree! thanks for the suggestion

SteveShani (approved):

PR #655, opened 2025-04-21
feat: collect java class files outside of jar

davidgonor:
- [x] Ready for review

#### What does this PR do?
Supports extraction of class files (java) located outside of a jar. 

lib/inputs/java/static.ts

Runtime team to validate that class files are collected upon CR connector run

Usually java class files can be found in jars/ wars or simply in the image layers not behind any archived file.
We should collect these class files too which is currently not happening.
Note that this logic is relevant only for CR integrations connectors (collectApplicationFiles must be on).

https://snyksec.atlassian.net/browse/RNTM-745

### DIFF
--- a/lib/inputs/java/static.ts
+++ b/lib/inputs/java/static.ts
@@ -8,7 +8,7 @@ const javaClassFileFormats = [".class"];
 
 function jarFilePathMatches(filePath: string): boolean {
   const dirName = path.dirname(filePath);
-  const fileExtension = filePath.slice(-4);
+  const fileExtension = filePath.slice(filePath.lastIndexOf("."));
   return (
     javaArchiveFileFormats.includes(fileExtension) &&
     !ignoredPaths.some((ignorePath) =>
@@ -25,7 +25,7 @@ export const getJarFileContentAction: ExtractAction = {
 
 function classFilePathMatches(filePath: string): boolean {
   const dirName = path.dirname(filePath);
-  const fileExtension = filePath.slice(-6);
+  const fileExtension = filePath.slice(filePath.lastIndexOf("."));
   return (
     javaClassFileFormats.includes(fileExtension) &&
     !ignoredPaths.some((ignorePath) =>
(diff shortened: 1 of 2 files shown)


lib/inputs/java/static.ts
```diff
@@ -18,6 +19,23 @@ function filePathMatches(filePath: string): boolean {
...
+    !ignoredPaths.some((ignorePath) =>
+      dirName.includes(path.normalize(ignorePath)),
+    )
+  );
+}
+
+export const getClassFileContentAction: ExtractAction = {
+  actionName: "class",
+  filePathMatches: classFilePathMatches,
   callback: streamToBuffer,
```
nozik, line 40: Do we really need the contents of the class files, or just their name?

lib/inputs/java/static.ts
```diff
@@ -4,8 +4,9 @@ import { streamToBuffer } from "../../stream-utils";
 
 const ignoredPaths = ["/usr/lib", "gradle/cache"];
```
nozik, line 5: Can you add the .m2 to the ignored paths?

lib/inputs/java/static.ts
```diff
@@ -18,6 +19,23 @@ function filePathMatches(filePath: string): boolean {
...
 export const getJarFileContentAction: ExtractAction = {
   actionName: "jar",
-  filePathMatches,
+  filePathMatches: jarFilePathMatches,
+  callback: streamToBuffer,
+};
+
+function classFilePathMatches(filePath: string): boolean {
+  const dirName = path.dirname(filePath);
+  const fileExtension = filePath.slice(-4);
```
nozik, line 28: Isn't there a util that extracts the extension?

davidgonor1408: Couldn't find one, but it's a good catch, updated the number to 6 '.class'.length

nata7che (approved):

lib/inputs/java/static.ts
nozik: Please use "lastIndexOf('.')" and check if it equals ".class", that would be more elegant

PR #657, opened 2025-05-02
fix!: update snyk-nodejs-lockfile-parser and drop support for EOL node versions

gergo-papp:
- [x] Ready for review

#### What does this PR do?

Fixes vulnerability in cross-spawn

BREAKING CHANGE: Minimum Node.js version raised to 18.


package.json
```diff
@@ -74,11 +74,6 @@
     "tslint-config-prettier": "^1.18.0",
     "typescript": "~4.7.3"
   },
-  "overrides": {
-    "@yarnpkg/core": {
-      "cross-spawn": "^7.0.5"
```
gergo-papp, line 79: this is no longer needed, because `snyk-nodejs-lockfile-parser` pulls in a newer `cross-spawn` package.

This change also drops support for node <18

pecodez:
Breaking changes to a public repo are subject to a specific release strategy, which I am not 100% sure of yet. In this case I would limit this PR to the override for the `cross-spawn` package and the addition of the newer Node versions to `appveyor`. We can raise the question of minimum version support with Product.

gergo-papp:
adding overrides is a hack and should be avoided especially in public repos, so I think this is an extra incentive to update now

PR #679, opened 2025-07-25
feat: migrate inspect to direct API call

bgardiner:
- [x] Ready for review

#### What does this PR do?

Migrate docker inspect calls to be direct API calls to the docker daemon instead of subprocess calls that use the docker CLI. This is a cleaner approach and mitigates various potential failure scenarios that arise from the subprocess call.

https://snyksec.atlassian.net/browse/CN-222

### DIFF
--- a/lib/analyzer/image-inspector.ts
+++ b/lib/analyzer/image-inspector.ts
@@ -23,7 +23,7 @@ async function getInspectResult(
   targetImage: string,
 ): Promise<DockerInspectOutput> {
   const info = await docker.inspectImage(targetImage);
-  return JSON.parse(info.stdout)[0];
+  return info;
 }
 
 function cleanupCallback(imageFolderPath: string, imageName: string) {
--- a/lib/docker.ts
+++ b/lib/docker.ts
@@ -8,6 +8,7 @@ import * as Debug from "debug";
 import * as Modem from "docker-modem";
 import { createWriteStream } from "fs";
 import { Stream } from "stream";
+import { DockerInspectOutput } from "./analyzer/types";
 import * as subProcess from "./sub-process";
 
 export { Docker, DockerOptions };
@@ -23,6 +24,18 @@ interface DockerOptions {
 }
 
 const debug = Debug("snyk");
+
+/**
+ * Type guard to validate that an object conforms to the DockerInspectOutput interface
+ */
+function isValidDockerInspectOutput(data: any): data is DockerInspectOutput {
+  return (
+    data &&
+    typeof data === "object" &&
+    !Array.isArray(data) &&
+    typeof data.Architecture === "string"
+  );
+}
 
 class Docker {
   public static async binaryExists(): Promise<boolean> {
@@ -135,7 +148,38 @@ class Docker {
     });
   }
 
-  public async inspectImage(targetImage: string) {
-    return subProcess.execute("docker", ["inspect", targetImage]);
+  public async inspectImage(targetImage: string): Promise<DockerInspectOutput> {
+    const request = {
+      path: `/images/${targetImage}/json`,
+      method: "GET",
+      statusCodes: {
+        200: true,
+        404: "not found",
+        500: "server error",
+      },
+    };
+
+    debug(`Docker.inspectImage: targetImage: ${targetImage}`);
+
+    const modem = new Modem();
+
+    return new Promise<DockerInspectOutput>((resolve, reject) => {
+      modem.dial(request, (err, data) => {
+        if (err) {
+          return reject(err);
+        }
+
+        // Validate that the response conforms to DockerInspectOutput interface
+        if (!isValidDockerInspectOutput(data)) {
+          return reject(
+            new Error(
+              `Invalid Docker inspect response for image ${targetImage}: expected DockerInspectOutput format`,
+            ),
+          );
+        }
+
+        resolve(data);
+      });
+    });
   }
 }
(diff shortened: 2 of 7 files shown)


lib/docker.ts
```diff
@@ -135,7 +136,29 @@ class Docker {
...
+
+    const modem = new Modem();
+
+    return new Promise<DockerInspectOutput>((resolve, reject) => {
+      modem.dial(request, (err, data) => {
+        if (err) {
+          return reject(err);
+        }
+
+        resolve(data);
```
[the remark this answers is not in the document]
bgardiner: > The inspectImage method creates a new Modem instance without using the DockerOptions (e.g., host, TLS settings) provided to the Docker class, which prevents connecting to remote or TLS-secured Docker daemons.

Leaving as is for now. I think we are expecting the standard DOCKER environment variables to be set to configure most of these docker options and docker-modem respects them (see here: https://github.com/apocas/docker-modem/blob/430827215e6d2f08bde1b4c3955acd3f0c0b560e/lib/modem.js#L20-L75). Side note: Most of the fields on `DockerOptions` are never used. The only place `DockerOptions` is currently used is as an input argument to `pullCli`: in the two invocations of `pullCli`, no options are provided in one case and only `platform` in the other case.

> Additionally, it returns the raw JSON string response from modem.dial without parsing it into the expected DockerInspectOutput object, leading to unparsed data being returned to callers.

`docker-modem` parses the result to JSON: https://github.com/apocas/docker-modem/blob/430827215e6d2f08bde1b4c3955acd3f0c0b560e/lib/modem.js#L350

lib/analyzer/image-inspector.ts
```diff
@@ -23,7 +23,7 @@ async function getInspectResult(
   targetImage: string,
 ): Promise<DockerInspectOutput> {
   const info = await docker.inspectImage(targetImage);
-  return JSON.parse(info.stdout)[0];
+  return info;
```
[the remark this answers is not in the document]
bgardiner: Yes it should. The `DockerInspectOutput` interface [is](https://github.com/snyk/snyk-docker-plugin/blob/c4367e1d8d9341d17fdb5917c937fe11b2040c8d/lib/analyzer/types.ts#L24-L32):
```
export interface DockerInspectOutput {
  Id: string;
  Architecture: string;
  RootFS: {
    Type: string;
    Layers: string[];
  };
}
```

Those match the API docs for `GET /images/<name>/json`: https://docs.docker.com/reference/api/engine/version/v1.51/#tag/Image/operation/ImageInspect

However, I'm still getting used to typescript. What happens if this is not the case. Do I need to do some validation to ensure the json actually has those fields? This same concern would have existed for prior implementation as well (since the stdout of the subprocess call could have been anything).

lib/docker.ts
```diff
@@ -24,6 +25,15 @@ interface DockerOptions {
 
 const debug = Debug("snyk");
 
+/**
+ * Type guard to validate that an object conforms to the DockerInspectOutput interface
+ */
+function isValidDockerInspectOutput(data: any): data is DockerInspectOutput {
+  return (
+    data && typeof data === "object" && typeof data.Architecture === "string"
```
mtstanley-snyk, line 33: might be worth checking if data is an object but not `Array.isArray(data)`, since I think an array would pass the object test, although then you test for a specific field that an array wouldn't have, so maybe this is fine 🤔

lib/docker.ts
```diff
@@ -135,7 +145,38 @@ class Docker {
     });
   }
 
-  public async inspectImage(targetImage: string) {
-    return subProcess.execute("docker", ["inspect", targetImage]);
+  public async inspectImage(targetImage: string): Promise<DockerInspectOutput> {
```
mtstanley-snyk, line 148: I'm assuming we have existing tests that cover this method, and should catch issues introduced by this change?

lib/docker.ts
```diff
@@ -135,7 +145,38 @@ class Docker {
     });
   }
 
-  public async inspectImage(targetImage: string) {
-    return subProcess.execute("docker", ["inspect", targetImage]);
```
mtstanley-snyk, line 139: Do we think there could be a reason we used a call to the binary here instead of implementing this using the docker daemon directly originally?

mtstanley-snyk (commented):
couple small questions, otherwise looks good to me, @parker-snyk are you good with @bgardiner 's response to your q?

bgardiner (commented):
(forgot to post these comments previously)

lib/docker.ts
bgardiner: Good question. @adrobuta Do you have any additional insight here?

I can think of reasons why we would call the binary for the `docker pull` (separate logic in this file to be handled in separate PR), since it would have access to registry authentication information that the user might have previously configured (e.g., when running `docker login`). However, I cannot think of a reason why we would not make the API request here.

adrobuta (dismissed):

lib/docker.ts
bgardiner: added this additional check.

lib/docker.ts
bgardiner: Yes. For example, there are overall tests for `getImageArchive` [here](https://github.com/snyk/snyk-docker-plugin/blob/0e988127a5fe4b8d27f68150b76b70aa7441da26/test/lib/analyzer/image-inspector.spec.ts#L106-L184), which internally calls getInspectResult() which in turn calls docker.inspectImage(). Both the cases of image existing on local docker daemon and it image NOT existing on local docker daemon are tested.

adrobuta (approved):

bgardiner:
> @bgardiner should we merge?

Yes! Thanks for the ping.

PR #704, opened 2025-08-26
fix: update old quay images

parker-snyk:
quay.io deletes previous versions of images as soon as they push a new version. This PR fixes the breaking tests, as they're pointing to now outdated images

We should redo the tests to use a different image that won't get removed every month, but in the meantime this will fix the faililng tests


dan-arpino (approved):
Looks good. I agree though that long term we should look for a better option.

PR #707, opened 2025-09-09
fix: file stream memory allocation bug

parker-snyk:
- [x] Ready for review

#### What does this PR do?
Currently we try to load files fully into memory before we even check if they're a binary (fo go parsing). While rare, this will crash the scanner if the container contains files larger than 2GB. 

This PR will only load the first four bytes first to check if the file has the magic ELF header (which means it's a binary of _some_ sorts)

Ideally we should do more to check, but this should solve the issue for the customer that's affected for the time being

Screenshots:
### Scanning vulnerable go packages
The files are all under 2GB, so these screenshots show that no scanning changes occur from prod -> this pr change
Prod 
image
image

PR changes
image
image

### Scanning containers with large file sizes
Prod (scan returns blank / the scan crashes) Bad
image

PR changes (scan completes, no errors or crashes) Good
image

### DIFF
--- a/test/fixtures/centos-shas.ts
+++ b/test/fixtures/centos-shas.ts
@@ -6,5 +6,5 @@ export const CENTOS_SHAS = {
   stream9:
     "sha256:0cae7e53f997228f0afc523f1bad28f3347dd0216ea8e27977f7af33f626f912",
   stream10:
-    "sha256:798fbe5c40fba2609fae716f5414af6654df21d2552927e85de6a11d20bbff7d",
+    "sha256:8749c1fde6471d2dd94d6b1cee8ac43a070573ce9cc5e85bf59b9e8331bc1974",
 } as const;
--- a/test/system/package-managers/__snapshots__/rpm.spec.ts.snap
+++ b/test/system/package-managers/__snapshots__/rpm.spec.ts.snap
@@ -2350,7 +2350,7 @@ Object {
                       "nodeId": "crypto-policies-scripts@20250905-1.gitc7eb7b2.el10",
                     },
                     Object {
-                      "nodeId": "cryptsetup-libs@2.7.5-2.el10",
+                      "nodeId": "cryptsetup-libs@2.8.1-1.el10",
                     },
                     Object {
                       "nodeId": "curl@8.12.1-2.el10",
@@ -2428,19 +2428,19 @@ Object {
                       "nodeId": "glib2@2.80.4-8.el10",
                     },
                     Object {
-                      "nodeId": "glibc@2.39-59.el10",
+                      "nodeId": "glibc@2.39-63.el10",
                     },
                     Object {
-                      "nodeId": "glibc-common@2.39-59.el10",
+                      "nodeId": "glibc-common@2.39-63.el10",
                     },
                     Object {
-                      "nodeId": "glibc-gconv-extra@2.39-59.el10",
+                      "nodeId": "glibc-gconv-extra@2.39-63.el10",
                     },
                     Object {
-                      "nodeId": "glibc-langpack-en@2.39-59.el10",
+                      "nodeId": "glibc-langpack-en@2.39-63.el10",
                     },
                     Object {
-                      "nodeId": "glibc-minimal-langpack@2.39-59.el10",
+                      "nodeId": "glibc-minimal-langpack@2.39-63.el10",
                     },
                     Object {
                       "nodeId": "gmp@1:6.2.1-12.el10",
@@ -2647,10 +2647,10 @@ Object {
                       "nodeId": "openssl-libs@1:3.5.1-5.el10",
                     },
                     Object {
-                      "nodeId": "p11-kit@0.25.5-7.el10",
+                      "nodeId": "p11-kit@0.25.10-1.el10",
                     },
                     Object {
-                      "nodeId": "p11-kit-trust@0.25.5-7.el10",
+                      "nodeId": "p11-kit-trust@0.25.10-1.el10",
                     },
                     Object {
                       "nodeId": "pam-libs@1.6.1-8.el10",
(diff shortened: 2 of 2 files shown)


lib/go-parser/index.ts
```diff
@@ -131,13 +131,22 @@ async function findGoBinaries(
       const first4Bytes = chunk.toString(encoding, 0, 4);
```
bgardiner, line 131: is it possible we get less than 4 bytes in the first chunk?

lib/go-parser/index.ts
```diff
@@ -131,13 +131,22 @@ async function findGoBinaries(
...
+
         Buffer.from(chunk).copy(buffer, bytesWritten, 0);
         bytesWritten += chunk.length;
+
         // Listen to next chunks only if it's an ELF executable
         stream.addListener("data", (chunk) => {
-          Buffer.from(chunk).copy(buffer, bytesWritten, 0);
-          bytesWritten += chunk.length;
+          if (buffer && bytesWritten < buffer.length) {
+            Buffer.from(chunk).copy(buffer, bytesWritten, 0);
```
bgardiner, line 143: should we be adding a fourth parameter here of `buffer.length - bytesWritten` for safety?

test/unit/go-parser-memory-stream.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+describe("Go parser memory allocation fix", () => {
+  const findGoBinaries = getGoModulesContentAction.callback!;
+  const { filePathMatches } = getGoModulesContentAction;
+
+  describe("Core memory fix - large non-ELF files", () => {
+    it("should not allocate large buffers for non-ELF files", async () => {
+      // This is the main bug we fixed - simulate 5GB non-ELF file
+      const nonElfContent = Buffer.from("This is not an ELF file");
+      const stream = createStreamFromBuffer(nonElfContent);
+      const hugeReportedSize = 5 * 1024 * 1024 * 1024; // 5GB
```
bgardiner, line 25: 😮

test/unit/go-parser-memory-stream.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+      try {
+        const goBinaryBuffer = readFileSync(goBinaryPath);
+        const stream = createStreamFromBuffer(goBinaryBuffer);
+
+        const result = await findGoBinaries(stream, goBinaryBuffer.length);
+
+        // Should process ELF files (may or may not find Go modules)
+        expect(typeof result).toBeDefined();
+      } catch (e) {
+        // If fixture doesn't exist, create a minimal ELF file for testing
```
bgardiner, line 52: why would this be the case? I'm probably mis-understanding, but doesn't this simply mean the test is written incorrectly?

test/unit/go-parser-memory-stream.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+      const nonElfContent = Buffer.from("Not ELF content");
+      const stream = createStreamFromBuffer(nonElfContent);
+
+      const result = await findGoBinaries(stream, 1024 * 1024); // 1MB
+
+      expect(result).toBeUndefined();
+    });
+
+    it("should reject ELF magic split across chunks (current limitation)", async () => {
+      // Our current implementation doesn't handle ELF magic split across chunks
```
bgardiner, line 103: should we fix this in future PR?

lib/go-parser/index.ts
bgardiner: update: I think this is related to the test you wrote that splits ELF magic across headers, but I"ll let you confirm

test/unit/go-parser-memory-stream.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+      expect(filePathMatches("/var/log/app")).toBe(false);
+      expect(filePathMatches("/tmp/file")).toBe(false);
+      expect(filePathMatches("/proc/cpuinfo")).toBe(false);
+    });
+
+    it("should handle the problematic case - large files without extensions", () => {
+      // These would previously cause the memory issue
+      expect(filePathMatches("/app/large-data-file")).toBe(true);
+      expect(filePathMatches("/usr/local/huge-binary")).toBe(true);
+    });
```
bgardiner, line 142: I would argue these cases add no value on top of the `should match files without extensions` ones above.

test/unit/go-parser-memory-stream.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+  });
+
+  describe("Error handling", () => {
+    it("should handle stream errors gracefully", async () => {
```
bgardiner, line 146: nit: throwing an error is graceful 🤨 ? What would non-graceful handling look like?

test/unit/go-parser-memory-stream.spec.ts
parker-snyk: Thoughts?

test/unit/go-parser-memory-stream.spec.ts
parker-snyk: I had AI go over my tests to make sure they were comprehensive. I thought I undid this weird fallback it added. Thanks for catching this

lib/go-parser/index.ts
parker-snyk: From what I can tell, it's very unlikely. The ELF header is always the first 4 bytes, but from what I've read there are very small chances where streaming a tar file results in unaligned bytes. So the first four wouldn't be guaranteed to be the first four of the file?

Let me do a few more hours of research, and I'll come back with a definitive answer hopefully. Either way, the functionality is currently the same as it was before the PR, but I agree we should improve the parsing and checks where possible

PR #750, opened 2026-02-05
fix: resolve platform detection in archives missing index metadata

bgardiner:
Modern Docker versions often produce OCI archives where individual manifests in the index.json lack platform information. This caused the OCI extractor to fail when multiple manifests were present because it couldn't disambiguate them. This fix allows the extractor to resolve platform information from image configuration blobs as a fallback.

- [x] Ready for review

#### What does this PR do?

Modern Docker versions often produce OCI archives where individual manifests in the index.json lack platform information. This caused the OCI extractor to fail when multiple manifests were present because it couldn't disambiguate them. This fix allows the extractor to resolve platform information from image configuration blobs as a fallback.

N/A

Create a tar archive containing multiple manifests from a modern docker daemon with containerd disabled as the image store. Running `snyk container test` on such an image fails with existing public versions of the CLI. Running `snyk container test` with this commit succeeds.

https://snyk.slack.com/archives/C85H8Q12Q/p1769554752582849

### DIFF
--- a/lib/extractor/oci-archive/layer.ts
+++ b/lib/extractor/oci-archive/layer.ts
@@ -51,7 +51,7 @@ export async function extractArchive(
 
     const layers: Record<string, ExtractedLayers> = {};
     const manifests: Record<string, OciArchiveManifest> = {};
-    const configs: ImageConfig[] = [];
+    const configs: Record<string, ImageConfig> = {};
     let mainIndexFile: OciImageIndex;
     const indexFiles: Record<string, OciImageIndex> = {};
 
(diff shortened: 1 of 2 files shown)


adrobuta (approved):

lib/extractor/oci-archive/layer.ts
```diff
@@ -278,11 +288,27 @@ function getImageManifestInfo(
...
+
+      // try to resolve platform from config
+      const manifest = manifestCollection[target.digest];
+      if (manifest) {
+        const config = configs[manifest.config.digest];
+        if (config) {
+          return {
+            os: config.os,
+            architecture: config.architecture,
+          };
```
adrobuta: should we return the variant too here?

PR #766, opened 2026-03-03
fix: dockerfile attribution

d3vco:
- [X] Ready for review

### What does this PR do?

This PR fixes 2 related issues:
1. `--exclude-base-image-vulns` has been disabled in SDP resulting in the flag not being effective in `container monitor` mode.
2. The logic to identify dockerfile-introduced dependencies under-attributes dependencies to the dockerfile. This resulted in vulnerabilities incorrectly being filtered as "base image vulns" and missing dockerfile instruction labels.

#### `buildResponse` orchestration
- Removed `collectDockerfilePkgs`, logic is inlined.
- Expands both versions of dockerfile packages using `getUserInstructionDeps` up front (in place)
- Derives `dockerfilePkgs` as `dockerfileAnalysis?.dockerfilePackages ?? autoDetected…dockerfilePackages` and uses the result to conduct annotation and exclusion (matches [CLI logic](https://github.com/snyk/cli/blob/5074d4ce12a830867482267d4caf1902937460e6/src/lib/snyk-test/run-test.ts#L129-L132)).
- Stores the result of filtering and annotation! (`depsAnalysis.depTree.dependencies = finalDeps`)

#### `excludeBaseImageDeps`
- Removed `extractDockerfileDeps`, filtering is inlined.
- Conducts matching on full name OR source segment. (On main, only exact keys matched)

#### `annotateLayerIds`
- Conducts matching on full name OR source segment.

### How should this be manually tested?

Unit tests have been added to demonstrate the new behavior.

The changes are most easily observed in monitor mode. Creating a small image with packages installed via a run instruction should show those dependencies attributed to the install instruction. Any base image vulns should be filterable, in test or monitor mode.

An example is shown below.

### Any background context you want to provide?

More context provided in this [write-up](https://docs.google.com/document/d/1wK2rSAXeXox_KZHfz7o-noYXFFS3jIG5dnVT3XuXE2U/edit?usp=sharing).

### What are the relevant tickets?

- [CN-568](https://snyksec.atlassian.net/browse/CN-568)
- [CN-846](https://snyksec.atlassian.net/browse/CN-846)

### Screenshots

Current Behavior:
Screenshot 2026-02-27 at 11 44 13 AM

PR Behavior:
Screenshot 2026-02-27 at 11 44 56 AM

[CN-568]: https://snyksec.atlassian.net/browse/CN-568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/response-builder.ts
```diff
@@ -26,14 +26,33 @@ async function buildResponse(
...
+      deps,
+    );
+  }
+
+  const dockerfilePkgs =
+    dockerfileAnalysis?.dockerfilePackages ??
+    depsAnalysis.autoDetectedUserInstructions?.dockerfilePackages;
+
+  if (dockerfilePkgs) {
+    const finalDeps = excludeBaseImageDeps(
```
mtstanley-snyk: just an FYI this seems to be only a partial copy, right? So fields like labels are a shared mutable reference between the `deps` and `finalDeps` objects, I think this is ok for the purpose of the PR but could cause unexpected results in the future.

mtstanley-snyk (commented):
couple comments/qs, otherwise lgtm 🎉. Did we decide on a rollout for this? If not, might be worth a short section in your doc on impact and then can start a thread with team on breaking vs non-breaking change.

My understanding of the impact is:
1. no(?) change to `snyk container test`
2. `snyk container monitor -- exclude-base-...` now excludes things, whereas before the command would have run but not excluded anything
3. monitored container projects from CLI where dockerfile was provided would now attribute a lot more vulns to user installed
4. monitored container projects from CLI without dockerfile should be no(?) change

mtstanley-snyk (dismissed):
🚢

mtstanley-snyk (approved):

PR #779, opened 2026-04-06
refactor(deps): replace mkdirp, uuid, tmp with node builtins

parker-snyk:
- [x] Ready for review

#### What does this PR do?
Removes `mkdirp`, `uuid`, and `tmp` from dependencies and uses native Node built-ins (`fs`, `crypto`, `os`) instead. Since the minimum supported Node version is 20.19, we can safely use features like `fs.mkdirSync(..., { recursive: true })` and `crypto.randomUUID()`.

`lib/analyzer/image-inspector.ts` and `lib/image-save-path.ts` have the main functional changes. `test/system/docker.spec.ts` has a minor test fix to clean up temporary mock tarballs properly.

`npm run test:system` and `npm run test:unit`.

This was prompted by a desire to trim down the dependency tree. Moving off of `tmp` also forced us to be explicit about test cleanup, avoiding silent file leaks in `/tmp` during the test suite.

CN-1003

#### Screenshots
N/A

#### Additional questions
None


lib/analyzer/applications/node-modules-utils.ts
```diff
@@ -22,7 +23,7 @@ interface ScanPaths {
 async function createTempProjectDir(
   projectDir: string,
 ): Promise<{ tmpDir: string; tempProjectRoot: string }> {
-  const tmpDir = await mkdtemp("snyk");
```
jcambier, line 25: info: does adding the "-" in this tmpDir affect/break anything downstream? Otherwise this all looks good to me!

package.json
```diff
@@ -46,17 +46,14 @@
...
     "gunzip-maybe": "^1.4.2",
     "minimatch": "^9.0.0",
-    "mkdirp": "^1.0.4",
     "packageurl-js": "1.2.0",
     "semver": "^7.7.3",
     "shescape": "^2.1.7",
     "snyk-nodejs-lockfile-parser": "^2.2.2",
     "snyk-poetry-lockfile-parser": "1.9.1",
     "snyk-resolve-deps": "^4.9.1",
-    "tar-stream": "^2.2.0",
```
jcambier, line 56: info: Was tar-stream downgraded because of the changes to the other dependencies? Looking at the lock file I'm not sure if it resulted in a change to the installed version?

lib/analyzer/applications/node-modules-utils.ts
parker-snyk: The '-' just makes the temp dirs a little easier to read, so instead of `snykA859JF`, it'll be `snyk-A859JF`. Nothing uses it downstream so it shouldn't break

package.json
parker-snyk: I think this was an error from a merge conflict. Will fix

jcambier (approved):

PR #785, opened 2026-04-08
fix(deps): update node lockfile parser

d3vco:
- [X] Ready for review

#### What does this PR do?
Resolves TypeScript compilation errors caused by the update of `snyk-nodejs-lockfile-parser` to `^2.2.3`. The lockfile parser update introduced type signature changes that became incompatible with the `legacy.depTreeToGraph` function in `@snyk/dep-graph`.

- `lib/analyzer/applications/node.ts`: Look at the modifications to `stripUndefinedLabels`, `depGraphFromNodeModules`, and `buildDepGraphFromDepTree`.
- `test/system/application-scans/node.spec.ts`: Test file adjustments.

1. Run `npm run build` and ensure TypeScript compiles without errors.
2. Run `npm run test` to verify that the application scans still parse node dependencies and build the graph correctly.

**Justification for casting types as `any`:**
The `snyk-nodejs-lockfile-parser` update altered the `DepTreeDep` interface so that the `labels` object now supports an `Alias` type: `{ [key: string]: string | Alias | undefined }`. 
However, `@snyk/dep-graph`'s legacy `DepTree` type expects labels to be strictly `{ [key: string]: string | undefined }`. 

Because of this mismatch between the two libraries' typings, passing the parsed `PkgTree` directly into `legacy.depTreeToGraph` results in TS2345 errors. 

**Why it is necessary:**
Without updating the upstream `@snyk/dep-graph` legacy types (which is deprecated and ideally shouldn't be touched), we have to bridge the gap locally. `stripUndefinedLabels` was updated to handle `Record<string, any>` so it can strip `undefined` fields without crashing on `Alias` values. 

**Why it will not lead to type issues:**
Casting `PkgTree` to `any` at the boundary of `legacy.depTreeToGraph` is safe because `@snyk/dep-graph` processes the tree at runtime without enforcing strict primitive `string` type checks on the label values. The `Alias` information flows through successfully to the `DepGraph` construction without causing runtime exceptions. By restricting the `any` cast strictly to the function call arguments (`pkgTree as any`), we preserve type safety in the rest of the application while appeasing the compiler at the deprecated legacy API boundary.

[CN-1007](https://snyksec.atlassian.net/browse/CN-1007)

#### Screenshots
N/A

#### Additional questions
N/A

[CN-1007]: https://snyksec.atlassian.net/browse/CN-1007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/analyzer/applications/node.ts
+++ b/lib/analyzer/applications/node.ts
@@ -136,7 +136,7 @@ async function depGraphFromNodeModules(
       }
 
       const depGraph = await legacy.depTreeToGraph(
-        pkgTree,
+        pkgTree as any,
         pkgTree.type || "npm",
       );
 
@@ -417,7 +417,7 @@ function stripUndefinedLabels(
   parserResult: lockFileParser.PkgTree,
 ): lockFileParser.PkgTree {
   const optionalLabels = parserResult.labels;
-  const mandatoryLabels: Record<string, string> = {};
+  const mandatoryLabels: Record<string, any> = {};
   if (optionalLabels) {
     for (const currentLabelName of Object.keys(optionalLabels)) {
       if (optionalLabels[currentLabelName] !== undefined) {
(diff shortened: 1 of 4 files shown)


ividalATSnyk (approved):
LGTM!

PR #791, opened 2026-04-13
feat: parse provenance attestations

bdemeo12:
- [ ] Ready for review


lib/facts.ts
```diff
@@ -104,6 +105,11 @@ export interface OCIDistributionMetadataFact {
   data: OCIDistributionMetadata;
 }
 
+export interface ProvenanceAttestationsFact {
+  type: "provenanceAttestations";
```
kateeselius: You will also need to add this to the enum here: https://github.com/snyk/snyk-docker-plugin/blob/main/components/common.yaml#L17

lib/extractor/oci-archive/layer.ts
```diff
@@ -517,6 +537,114 @@ function getImageConfig(
...
+    `[provenance-poc] Scanning ${metadata.mainIndexFile!.manifests.length} descriptors in image index`,
+  );
+
+  for (const descriptor of metadata.mainIndexFile!.manifests) {
+    debug(
+      `[provenance-poc] Descriptor: digest=${descriptor.digest}, mediaType=${descriptor.mediaType}, ` +
+        `platform=${JSON.stringify(descriptor.platform)}, annotations=${JSON.stringify(descriptor.annotations)}`,
+    );
+
+    const isAttestationManifest =
```
kateeselius, line 585: is this just for provenance attestation? Are there other types that could end up with unknown architecture and "attestation-manifest" annotation?

lib/extractor/oci-archive/layer.ts
```diff
@@ -517,6 +537,114 @@ function getImageConfig(
   );
 }
 
+function extractProvenanceAttestations(
+  metadata: ArchiveMetadata,
+): ProvenanceAttestation[] {
+  const attestations: ProvenanceAttestation[] = [];
+
+  debug(
+    `[provenance-poc] Scanning ${metadata.mainIndexFile!.manifests.length} descriptors in image index`,
```
kateeselius: would mainIndexFile ever be undefined?

lib/extractor/oci-archive/layer.ts
```diff
@@ -517,6 +537,114 @@ function getImageConfig(
...
+        `[provenance-poc] Could not find blob for attestation manifest ${descriptor.digest}`,
+      );
+      continue;
+    }
+
+    debug(
+      `[provenance-poc] Attestation manifest content: ${JSON.stringify(nestedManifest, null, 2)}`,
+    );
+
+    const attestationManifest = nestedManifest as OciArchiveManifest;
```
kateeselius: Would malformed attestations be an issue here if they can't be casted to OciArchiveManifest?

kateeselius (commented):
Also, you should add your amazing new fact to the enum list! https://github.com/snyk/snyk-docker-plugin/blob/main/components/common.yaml#L17

lib/extractor/oci-archive/layer.ts
bdemeo12: There are other in-toto attestation types, like SBOM, that also use "unknown" with attestation-manifest, but we filter for provenance specific attestations here: https://github.com/snyk/snyk-docker-plugin/blob/452a32ddb6b395625fb80083191b538130b08254/lib/extractor/oci-archive/layer.ts#L574

lib/extractor/oci-archive/layer.ts
bdemeo12: By the time we call this function, mainIndexFile is always defined because the archive has already been parsed, but I can add in an extra check to be safe

lib/extractor/oci-archive/layer.ts
bdemeo12: No, the cast doesn't check anything at runtime, so if a blob is malformed, the `!attestationManifest.layers` check on the next line will catch it and skip it.

lib/extractor/oci-archive/layer.ts
```diff
@@ -523,6 +534,73 @@ function getImageConfig(
...
+      descriptor.annotations?.["vnd.docker.reference.type"] ===
+        "attestation-manifest";
+
+    if (!isAttestationManifest) {
+      continue;
+    }
+
+    const nestedManifest = metadata.rawBlobs[descriptor.digest];
+    if (!nestedManifest) {
+      continue;
```
jcambier, line 601: My interpretation is that we've found a provenance attestation manifest here, but weren't able to locate the corresponding layer. This shouldn't happen, but theoretically, if it did, do you think we should log anything here?

lib/extractor/oci-archive/layer.ts
```diff
@@ -523,6 +534,73 @@ function getImageConfig(
...
+      const provenanceLayer: RawProvenanceAttestation["provenanceLayers"][number] =
+        {
+          digest: layer.digest,
+          mediaType: layer.mediaType,
+          annotations: layer.annotations,
+        };
+
+      if (isProvenanceLayer) {
+        const inTotoBlob = metadata.rawBlobs[layer.digest];
+        if (inTotoBlob) {
```
jcambier: Same question here as above -- log in case anything goes wrong?

lib/extractor/oci-archive/layer.ts
```diff
@@ -523,6 +534,73 @@ function getImageConfig(
...
+    const attestation: RawProvenanceAttestation = {
+      attestationManifestDigest: descriptor.digest,
+      mediaType: descriptor.mediaType,
+      annotations: descriptor.annotations || {},
+      provenanceLayers: [],
+    };
+
+    for (const layer of attestationManifest.layers) {
+      const isProvenanceLayer =
+        layer.annotations?.["in-toto.io/kind"] === "provenance" ||
```
jcambier: info: From the provenance attestation doc:
> There can be multiple attestations attached to this digest, like SBOM or Vuln scans, so filter for provenance attestation by where annotations["[in-toto.io/predicate-type](http://in-toto.io/predicate-type)"] starts with https://slsa.dev/provenance.

Is that step needed here or do we not need to worry about it?

d3vco:
Is this PR review bot comment a real concern?

> [Runtime Crash on Empty Subject](https://github.com/snyk/snyk-docker-plugin/pull/791/changes#diff-c45f6b08d05f7384de277acd5b6fd2a8889313d61a5606f6258cc657e48ccf0fR104-R104) 🟠 [major]
The functions extractFieldsSlsa02 and extractFieldsSlsa10 use non-null assertions and index access statement.subject![0] without verifying the array is present or has elements. While the in-toto spec usually requires a subject, a malformed or empty subject in the attestation will cause a runtime exception (TypeError: Cannot read properties of undefined), crashing the entire image scan.

lib/extractor/oci-archive/layer.ts
d3vco: I am not sure this is even reachable, but there is a potential bug here. At this point, if we haven't already found the image's manifest digest, we pop the first item from the list of manifests and assume that it's the image manifest. However, now that list will also contain attestation manifests and we could theoretically pop one of those and the scan would proceed thinking the attestation manifest is describing the image.

I think now that you have added more context to the `OciArchiveManifest` type we can improve this logic. The right answer might be looping until we find a manifest where `.annotations['vnd.docker.reference.type'] !== 'attestation-manifest'` or something similar.

d3vco, line 39: I believe we still intend to use this limit to limit the size of our scanResult payload, but have we validated that the behavior still works for this new feature (i.e. all attestation layers >2 MB will be dropped)?

I think some of the comments will need to be updated to clarify the intent.

(Also note that this limit is separately defined and enforced in `stream-utils.ts`)

lib/extractor/oci-archive/layer.ts
bdemeo12: I don't this will actually happen, this check is more of a typescript safeguard against undefined in the map look up!

lib/extractor/oci-archive/layer.ts
bdemeo12: We do this: https://github.com/snyk/snyk-docker-plugin/blob/poc/provenance-attestation/lib/extractor/provenance-parser.ts#L160

lib/extractor/types.ts
```diff
@@ -36,12 +36,20 @@ export interface ExtractionResult {
...
   containerConfig?: ContainerConfig | null;
   history?: HistoryEntry[] | null;
+  rawProvenanceAttestations?: RawProvenanceAttestation[];
 }
 
 export interface ExtractedLayers {
   [layerName: string]: FileNameAndContent;
 }
 
+export interface ExtractedLayersAndManifest {
```
d3vco: This type is already defined on L67 (without `rawProvenanceAttestations`). Recommend de-duping.

lib/extractor/oci-archive/layer.ts
jcambier: Cool, thanks for clarifying

lib/extractor/oci-archive/layer.ts
bdemeo12: I'm not sure how possible this is, but I think you're right, it could happen. I changed this to filter out the attestation manifests

d3vco:
Recommend adding context on this new addition to AGENTS.md

lib/extractor/oci-archive/layer.ts
bdemeo12: This is actually the desired behavior! We expect attestations to be much smaller than this: https://docs.google.com/document/d/1c9KDiazpydj6mVFb_PR453OAwDkLwUpbn1JQaC8gJE8/edit?tab=t.0#heading=h.uog5cis0puju

And if they are bigger we should drop them since we are passing these attestations in mem to Registry, and we do not want to send stuff that big!

bdemeo12: How do you think the comment should change ?

lib/analyzer/static-analyzer.ts
```diff
@@ -169,6 +169,7 @@ export async function analyze(
     imageCreationTime,
     containerConfig,
     history,
+    rawProvenanceAttestations,
```
mtstanley-snyk: nit: I think I'd rename the fact emitted to be `provenanceMetadata` since it's not a full attestation and instead is specific metadata collapsed across multiple different schema versions/types and then I would just call this stuff `provenanceAttestation` since it's the unadulterated thing.

lib/extractor/oci-archive/layer.ts
mtstanley-snyk: I don't think we necessarily need to impose this limit for attestations anymore, right? Since we are parsing it down to just a couple fields, the thing we want to apply a size limit on is the resulting fact we emit, but since we've capped the data we collect in the parser we should be good, right? Separately if `snyk-docker-plugin` enforces a 2mb limit on any json file it tries to parse I think that's fine to leave and have it also apply to the attestations too

lib/extractor/oci-archive/layer.ts
```diff
@@ -57,7 +59,8 @@ export async function extractArchive(
   const metadata = await extractMetadata(ociArchiveFilesystemPath);
 
   // Determine which manifest and layers we need
-  const { manifest, imageConfig } = resolveManifestAndConfig(metadata, options);
+  const { manifest, imageConfig, rawProvenanceAttestations } =
+    resolveManifestAndConfig(metadata, options);
```
mtstanley-snyk: Looking through the rest of the code here it might simplify things/make more sense in terms of following the existing pattern if we change this so that in this method we find the manifest and config for the platform-specific image we're currently scanning and then separately just the manifests for the provenance attestations we care about, then follow the same mechanism of pulling the layers for the attestation and passing it to the extract layers function which would then have a separate control flow for handling the intoto layer vs tar archive of a normal image. Let me know if you want to chat about what that would look like.

lib/extractor/oci-archive/layer.ts
```diff
@@ -123,6 +127,7 @@ interface ArchiveMetadata {
   manifests: Record<string, OciArchiveManifest>;
   indexFiles: Record<string, OciImageIndex>;
   configs: ImageConfig[];
+  rawBlobs: Record<string, unknown>;
```
mtstanley-snyk: I think we should avoid this, see above comment about a potential alternative path to take

lib/extractor/oci-archive/layer.ts
```diff
@@ -406,7 +427,7 @@ function getManifest(
   platformInfo: OciPlatformInfo,
 ): OciArchiveManifest | undefined {
   if (!imageIndex) {
-    return manifestCollection[Object.keys(manifestCollection)[0]];
+    return Object.values(manifestCollection).find(isImageManifest);
```
mtstanley-snyk, line 453: why is this change necessary?

lib/extractor/oci-archive/layer.ts
```diff
@@ -522,6 +543,73 @@ function getImageConfig(
...
+function extractProvenanceAttestations(
+  metadata: ArchiveMetadata,
+): RawProvenanceAttestation[] {
+  const attestations: RawProvenanceAttestation[] = [];
+
+  if (!metadata.mainIndexFile) {
+    return attestations;
+  }
+
+  for (const descriptor of metadata.mainIndexFile.manifests) {
```
mtstanley-snyk: does this currently work? my understanding was that `mainIndexFile` is the `index.json` file, no? And you would have to traverse `index.json` -> `image index` -> `manifest`, it seems like we're missing a layer with this logic 🤔, but I'm likely missing something

Shouldn't we just use the same logic that `getManifest` uses to collect all of the manifests and then pick the attestation associated with the image manifest we are currently scanning?

lib/extractor/oci-archive/layer.ts
```diff
@@ -522,6 +543,73 @@ function getImageConfig(
...
+    const isAttestationManifest =
+      descriptor.platform?.architecture === "unknown" &&
+      descriptor.annotations?.["vnd.docker.reference.type"] ===
```
mtstanley-snyk, line 586: don't we need to check `vnd.docker.reference.digest` somewhere? So that we only pull the attestations associated with the specific platform image we care about?

lib/extractor/oci-archive/layer.ts
```diff
@@ -522,6 +543,73 @@ function getImageConfig(
...
+): RawProvenanceAttestation[] {
+  const attestations: RawProvenanceAttestation[] = [];
+
+  if (!metadata.mainIndexFile) {
+    return attestations;
+  }
+
+  for (const descriptor of metadata.mainIndexFile.manifests) {
+    const isAttestationManifest =
+      descriptor.platform?.architecture === "unknown" &&
```
mtstanley-snyk: do we technically need to check this? seems like reference type would be more authoritative? But if both are specified in the spec then seems fine to check both

lib/extractor/oci-archive/layer.ts
```diff
@@ -522,6 +543,73 @@ function getImageConfig(
...
+      descriptor.annotations?.["vnd.docker.reference.type"] ===
+        "attestation-manifest";
+
+    if (!isAttestationManifest) {
+      continue;
+    }
+
+    const nestedManifest = metadata.rawBlobs[descriptor.digest];
```
mtstanley-snyk: do we need this raw blobs object for this? can't we just pull it from the list of manifests we already found?

lib/extractor/types.ts
```diff
@@ -115,9 +124,32 @@ export interface OciPlatformInfo {
...
+  _type?: string;
+  subject?: Array<{
+    name?: string;
+    digest?: Record<string, string>;
+  }>;
+  predicateType?: string;
+  predicate?: Record<string, unknown>;
+}
+
+export interface RawProvenanceAttestation {
```
mtstanley-snyk: nit: in terms of naming I think I'd argue that this is actually just the image manifest that points to provenance attestations inside the actual layers

lib/extractor/oci-archive/layer.ts
```diff
@@ -522,6 +543,73 @@ function getImageConfig(
...
+      attestationManifestDigest: descriptor.digest,
+      mediaType: descriptor.mediaType,
+      annotations: descriptor.annotations || {},
+      provenanceLayers: [],
+    };
+
+    for (const layer of attestationManifest.layers) {
+      const isProvenanceLayer =
+        layer.annotations?.["in-toto.io/kind"] === "provenance" ||
+        layer.mediaType === "application/vnd.in-toto+json";
```
mtstanley-snyk: if the annotation here is missing won't we collect attestations that aren't of type provenance?

mtstanley-snyk: for example the image I inspected had the annotation `"in-toto.io/predicate-type": "https://slsa.dev/provenance/v0.2"`, I wonder if we can fully rely on these annotations or if we also need to check the actual predicate type field in the intoto body? I guess we could change this to collect all attestation types at this point (`layer.mediaType === "application/vnd.in-toto+json"`) and then filter later when parsing the intoto body to collect provenance metadata?

lib/extractor/provenance-parser.ts
```diff
@@ -0,0 +1,210 @@
...
+const MAX_ATTESTATIONS_PER_IMAGE = 10;
+
+export interface DockerfileMetadata {
+  path: string;
+  contents: string | null;
+}
+
+export interface ProvenanceAttestation {
+  buildTimestamp: string | null;
+  commit: string | null;
```
mtstanley-snyk: `buildConfigSourceDigest`?

lib/extractor/provenance-parser.ts
```diff
@@ -0,0 +1,210 @@
...
+  manifestDigest: string;
```
mtstanley-snyk: is this the digest of the attestation manifest? we likely should be more specific with the naming here

lib/extractor/provenance-parser.ts
```diff
@@ -0,0 +1,210 @@
...
+  repoUri: string | null;
```
mtstanley-snyk: `buildConfigSourceUri`?

lib/extractor/provenance-parser.ts
```diff
@@ -0,0 +1,210 @@
+import * as Debug from "debug";
+import { InTotoStatement, RawProvenanceAttestation } from "./types";
+
+const debug = Debug("snyk");
+
+const MAX_ATTESTATIONS_PER_IMAGE = 10;
+
+export interface DockerfileMetadata {
+  path: string;
+  contents: string | null;
```
mtstanley-snyk: https://docs.google.com/document/d/1c9KDiazpydj6mVFb_PR453OAwDkLwUpbn1JQaC8gJE8/edit?disco=AAAB8TQDwVs

lib/extractor/provenance-parser.ts
```diff
@@ -0,0 +1,210 @@
...
+  buildTimestamp: string | null;
+  commit: string | null;
+  manifestDigest: string;
+  repoUri: string | null;
+  builderId: string;
+  buildType: string;
+  dockerfileMetadata: DockerfileMetadata;
+}
+
+interface SlsaPredicate02 {
```
mtstanley-snyk: nit: can we do `SlsaPredicateV0_2` or something similar?

lib/extractor/provenance-parser.ts
```diff
@@ -0,0 +1,210 @@
...
+      source?: {
+        infos?: Array<{ data?: string }>;
+        locations?: Record<string, { data?: string }>;
+      };
+    };
+  };
+  invocation?: {
+    configSource?: {
+      uri?: string;
+      digest?: { sha1?: string };
```
mtstanley-snyk: it seems like this might not always be `sha1` for the digest algo right?

lib/extractor/provenance-parser.ts
```diff
@@ -0,0 +1,210 @@
...
+  manifestDigest: string,
+): ProvenanceAttestation {
+  const buildTimestamp = predicate.metadata?.buildStartedOn || null;
+
+  const buildkitMeta =
+    predicate.metadata?.["https://mobyproject.org/buildkit@v1#metadata"];
+
+  const remoteCommit = predicate.invocation?.configSource?.digest?.sha1;
+  const localCommit = buildkitMeta?.vcs?.revision;
+  const commit = remoteCommit || localCommit || null;
```
mtstanley-snyk: should we indicate in the fact somewhere whether we parsed this as remote or local? Given that the documentation says:

```
Note that, contrary to the buildDefinition.externalParameters.configSource field, BuildKit doesn't verify the vcs values, and as such they can't be trusted and should only be used as a metadata hint.
```

lib/extractor/provenance-parser.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  const builderId = predicate.builder?.id || "";
+  const buildType = predicate.buildType || "";
+
+  const dockerfilePath =
+    predicate.invocation?.configSource?.entryPoint || "Dockerfile";
+
+  const dockerfileKey = "Dockerfile";
+  const dockerfileContents =
+    buildkitMeta?.source?.infos?.[0]?.data ||
```
mtstanley-snyk: why is this an array? and how can we be sure the first item is what we want? I couldn't find details on this in the spec 🤔

lib/extractor/provenance-parser.ts
mtstanley-snyk: we probably want to include both?

lib/extractor/provenance-parser.ts
```diff
@@ -0,0 +1,210 @@
...
+    buildType,
+    dockerfileMetadata: {
+      path: dockerfilePath,
+      contents: dockerfileContents,
+    },
+  };
+}
+
+function getManifestDigest(statement: InTotoStatement): string | null {
+  const sha256 = statement.subject?.[0]?.digest?.sha256;
```
mtstanley-snyk: do we know this will always be the format? i.e. 1 source and the first is the manifest digest of the container image this attests?

lib/extractor/provenance-parser.ts
```diff
@@ -0,0 +1,210 @@
...
+    return null;
+  }
+
+  const manifestDigest = getManifestDigest(statement);
+  if (!manifestDigest) {
+    debug("[provenance] No valid subject digest in in-toto statement");
+    return null;
+  }
+
+  const predicateType = statement.predicateType;
```
mtstanley-snyk, line 264: should we make this a type as a union of const strings? Then do a switch statement instead of ifs?

lib/extractor/provenance-parser.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  if (predicateType?.includes("provenance/v0.2")) {
```
mtstanley-snyk: should we check for a full string match, why do we want to only check if it contains this substring?

lib/extractor/provenance-parser.ts
```diff
@@ -0,0 +1,210 @@
...
+  return null;
+}
+
+export function parseProvenanceAttestations(
+  attestationManifests: RawProvenanceAttestation[],
+): ProvenanceAttestation[] {
+  const results: ProvenanceAttestation[] = [];
+
+  for (const manifest of attestationManifests) {
+    for (const provenanceLayer of manifest.provenanceLayers) {
```
mtstanley-snyk: will this always produce the exact same subset of attestations for subsequent scans if we hit the limit here?

lib/extractor/types.ts
```diff
@@ -115,9 +124,32 @@ export interface OciPlatformInfo {
 }
 
 export interface OciImageIndex {
+  mediaType?: string;
   manifests: OciManifestInfo[];
 }
 
+export interface InTotoStatement {
+  _type?: string;
+  subject?: Array<{
```
mtstanley-snyk: should we split this inline type definition into its own type? https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md

lib/extractor/provenance-parser.ts
```diff
@@ -0,0 +1,210 @@
...
+): ProvenanceAttestation {
+  const runDetails = predicate.runDetails;
+  const buildDefinition = predicate.buildDefinition;
+
+  const buildTimestamp = runDetails?.metadata?.startedOn || null;
+
+  const remoteCommit =
+    buildDefinition?.externalParameters?.configSource?.digest?.sha1;
+  const localCommit = runDetails?.metadata?.buildkit_metadata?.vcs?.revision;
+  const commit = remoteCommit || localCommit || null;
```
mtstanley-snyk: might be good to include the digest algo in this field, similar to container images: `sha256:abcdef`

test/lib/extractor/provenance-parser.spec.ts
mtstanley-snyk, line 1: should we also have tests around extracting the attestations from image in addition to the parsing logic we're testing here?

mtstanley-snyk (commented):
overall looks good! mostly a lot of qs from me and just one potential alternative design, let me know if you want to chat about any of it!

lib/extractor/oci-archive/layer.ts
bdemeo12: It might be overkill ... its a fallback for when there is no index.json, instead of grabbing the first manifest to avoid accidentally returning an attestation manifest as the image, we select the right one

lib/extractor/oci-archive/layer.ts
bdemeo12: ooooof, good catch.

Fixed it to flatten the indexes and added a vnd.docker.reference.digest check so we only keep attestations for the image manifest we're scanning

lib/extractor/oci-archive/layer.ts
bdemeo12: yeah, its probably overkill! I'll remove it

lib/extractor/oci-archive/layer.ts
bdemeo12: We can!

lib/extractor/types.ts
bdemeo12: I did change this to provenanceAttestation, but that might not be quite right either. What do you think of calling this attestationManifest ?

lib/extractor/oci-archive/layer.ts
bdemeo12: OCI annotations are optional, so its true, we may collect unneeded annotations here, but we do filter later: https://github.com/snyk/snyk-docker-plugin/blob/poc/provenance-attestation/lib/extractor/provenance-parser.ts#L179


I can simplify this though

lib/extractor/provenance-parser.ts
bdemeo12: Do you think sourceImageDigest makes more sense ? I didnt know if buildConfigSourceDigest was easily inferable enough

lib/extractor/provenance-parser.ts
bdemeo12: added!

lib/extractor/provenance-parser.ts
bdemeo12: yup! renamed SlsaPredicate02 → SlsaPredicateV0_2 and SlsaPredicate10 → SlsaPredicateV1_0

lib/extractor/provenance-parser.ts
bdemeo12: I removed it. Github commits don't include the digest algo anyways

lib/extractor/provenance-parser.ts
bdemeo12: I added in a field for this!

lib/extractor/provenance-parser.ts
bdemeo12: I think it is an array: https://github.com/moby/buildkit/blob/0f0aad9e38b86fb11c3996c79dbf0862fd9a91c4/solver/llbsolver/provenance/types/types.go#L47

but you're right, I don't think it is ordered. I changed it to filter for the filename which would be the file name of the Dockerfile.

lib/extractor/provenance-parser.ts
bdemeo12: Same as above, we can use the key!

lib/extractor/provenance-parser.ts
bdemeo12: We can!

lib/extractor/provenance-parser.ts
bdemeo12: I added in a  sort for the attestations, so now they should be!

lib/extractor/provenance-parser.ts
bdemeo12: Sure! I can add it for remote builds, but we dont have this is local builds

lib/extractor/provenance-parser.ts
bdemeo12: I discussed with Steve!

At a later date we will look into sending layer metadata from the attestation through the fact to be used by vulns by layer.

I the Dockerfile is now decoded and parsed using the same path as when a Dockerfile is passed via the CLI: https://github.com/snyk/snyk-docker-plugin/blob/main/lib/dockerfile/index.ts#L35-L48

lib/extractor/oci-archive/layer.ts
```diff
@@ -522,6 +563,105 @@ function getImageConfig(
...
+
+    const attestationManifest = metadata.manifests[descriptor.digest];
+    if (
+      !attestationManifest?.layers ||
+      !Array.isArray(attestationManifest.layers)
+    ) {
+      continue;
+    }
+
+    const attestation: ProvenanceAttestation = {
```
mtstanley-snyk: I don't think this is a blocker, but couldn't we simplify this a little more, isn't this just an `OciArchiveManifest` plus it's digest? Could potentially change this type to be defined as:
```
interface ResolvedProvenanceAttestationManifest {
  manifestDigest: string;
  manifest: OciArchiveManifest;
  intotoStatements: Record<string, IntotoStatement>; // key is layer digest
}
```
then you have something that unifies with existing types and makes the relationships of all the pieces involved a little clearer. Totally up to you, like I said not a blocker at all!

lib/extractor/oci-archive/layer.ts
mtstanley-snyk: A couple things I was thinking:

1. This change doesn't identify manifests that aren't attestations does it? Doesn't the logic match both since attestations are just regular manifests? (we already know everything in that collection is an image manifest)
2. This would be an existing bug, right? Nothing in your PR would cause this logic to now break, right?

I think I'd recommend tackling this in a separate PR if (2) is accurate, and I'd add tests that confirm the existing behavior is broken and that the new logic fixes those tests. The existing logic doesn't even select the manifest for the correct platform also, right?

lib/extractor/types.ts
mtstanley-snyk: left a related comment above https://github.com/snyk/snyk-docker-plugin/pull/791/changes#r3403857913, I think what you have is likely fine, but we can discuss

lib/extractor/oci-archive/layer.ts
mtstanley-snyk: Got it, if we're no longer filtering at this stage and we can't guarantee that this type is only populated with provenance attestations then we should change the naming here to call these just attestations without the provenance prefix, otherwise things can get confusing.

lib/extractor/provenance-parser.ts
mtstanley-snyk: This comment was about the `commit` field, I mainly wanted to make sure that we weren't using git/github specific terminology when it seems like these fields can represent any vcs, if we can confirm that all vcs use commit then it seems fine, but the spec here uses digest `predicate.invocation?.configSource?.digest` so it felt safe to just copy that.

For the manifest digests I think `attestedManifestDigest` and `attestationManifestDigest` seem good to me, what do you think? That way we avoid using `source` to mean a bunch of different things across fields.

lib/extractor/provenance-parser.ts
mtstanley-snyk: Sounds good, one q though, looking at the analyze dockerfile method I don't think this returns what @SteveShani needs for vulns by layer, it looks like this filters to just run instructions and instructions that match a particular regex. Might need to double-check this is the right thing to use.

lib/extractor/provenance-parser.ts
mtstanley-snyk: 🫡, should also rename the methods, right?

lib/extractor/provenance-parser.ts
mtstanley-snyk: What do they include? Does the logic you have still work for github?

lib/extractor/provenance-parser.ts
mtstanley-snyk: 👍 that solves for the digest algo, but what about the other pieces? Do we know the first item here is always the manifest digest of the container image this attests? Might need to do more research on how we can make this logic more robust, or we could just use the digest  we pulled why scanning, we already know the manifest digest that the attestation is for because that's the image we are scanning, we used the reference digest annotation to filter the list of attestations to be for just the image we care about.

lib/extractor/provenance-parser.ts
```diff
@@ -0,0 +1,326 @@
...
+  };
+}
+
+function getSourceImageDigest(statement: InTotoStatement): string | null {
+  const digest = statement.subject?.[0]?.digest;
+  if (!digest) {
+    return null;
+  }
+
+  const [algorithm, value] = Object.entries(digest)[0] ?? []; // Bella todo
```
mtstanley-snyk: reminder to remove todo comment

test/windows/__snapshots__/application-scans.spec.ts.snap
```diff
@@ -2443,6 +2443,25 @@ Object {
...
+          "data": Array [
+            Object {
+              "buildConfigCommit": null,
+              "buildConfigCommitSource": null,
+              "buildConfigSourceUri": null,
+              "buildTimestamp": "2025-10-22T15:17:20.705984586Z",
+              "buildType": "https://mobyproject.org/buildkit@v1",
+              "builderId": "",
+              "dockerfileMetadata": Object {
+                "analysis": null,
```
mtstanley-snyk, line 2458: any ideas on why this is null? Is that expected or did the test catch a bug?

mtstanley-snyk: did you check the rest of the fields here to make sure things are working as expected?

mtstanley-snyk (commented):
just a couple more comments from the previous review to finalize, nothing major though!

lib/extractor/oci-archive/layer.ts
bdemeo12: no, thats a good idea! Changed!

bdemeo12:
In response to this comment #https://github.com/snyk/snyk-docker-plugin/pull/791#discussion_r3404047962

You're right, it doesn't reliably distinguish them from image manifests.

I added a test for this and I created a ticket to work this in a follow up PR:
https://snyksec.atlassian.net/browse/CN-1480

Also, we only extract attestations when there's an image index! So if there is no index, there is no attestation. When there is an index, attestations are matched to the images manifest digest. We only explore one platform per scan, so this wont affect anything.

bdemeo12:
replying to https://github.com/snyk/snyk-docker-plugin/pull/791#discussion_r3404133819


I renamed these to remove the reference to provenance, since the collection can hold any attestation type.

bdemeo12:
replying to https://github.com/snyk/snyk-docker-plugin/pull/791#discussion_r3404250478

This will not include the fields needed for vulns by layer. We decided to do this in a follow up PR. 

Jira ticket: https://snyksec.atlassian.net/browse/CN-1481

test/windows/__snapshots__/application-scans.spec.ts.snap
bdemeo12: Yes, this is expected! This is testing an image that has an attestation built with mode=min, which will not have dockerfile contents. The populated path is covered by the unit tests in provenance-parser.spec.ts!

test/windows/__snapshots__/application-scans.spec.ts.snap
```diff
@@ -2443,6 +2443,25 @@ exports[`go binaries scanning with esbuild npm path that previously failed exten
           "data": "Alpine Linux v3.22",
           "type": "imageOsReleasePrettyName",
         },
+        {
+          "data": [
```
kateeselius, line 2447: Why did only this snapshot change? Should we add a non-windows test?

kateeselius (commented):
LGTM - 🚢  one question regarding snapshots but otherwise, I think it is ready. Have you been able to run it locally e2e to test? Ie. cli (using this version of SDP) --> registry --> CMD to see the fact get persisted? Its a pain, but it is good to verify all the pieces work together

test/windows/__snapshots__/application-scans.spec.ts.snap
bdemeo12: the `go-binaries.tar` is the only fixture with a provenance attestation (the fact is only emitted when one exists!)

I didnt add a non-windows test because this isnt windows specific behaviour !

kateeselius (approved):
🚢 LGTM

PR #804, opened 2026-04-28
feat: add phase-level latency metrics for container scans

bdemeo12:
- [ ] Ready for review

#### What does this PR do?

Adds per-phase latency timing to the container scan flow. Each major phase (image extraction, OS package analysis, JAR/Go/Node/PHP/Python analysis, binary detection, dep tree building) is timed with Date.now() and returned on the PluginResponse via a new optional analytics field. A companion CLI change wires these timings into CLI analytics metadata as containerPluginTimings.

The container plugin has no latency visibility, this PR is to collect latency metrics and surface them in the CLI analytics.

https://snyksec.atlassian.net/browse/CN-1170
https://snyksec.atlassian.net/browse/CN-1054

### DIFF
--- a/lib/scan-payload-metrics.ts
+++ b/lib/scan-payload-metrics.ts
@@ -1,0 +1,22 @@
+import type { ScanResult } from "./types";
+
+export interface ScanPayloadMetrics {
+  scanResultCount: number;
+  applicationScanResultCount: number;
+  scanResultPayloadBytes: number[];
+  totalScanResultsPayloadBytes: number;
+}
+
+export function computeScanPayloadMetrics(
+  scanResults: ScanResult[],
+): ScanPayloadMetrics {
+  const bytes = (v: unknown) =>
+    Buffer.byteLength(JSON.stringify(v), "utf8");
+
+  return {
+    scanResultCount: scanResults.length,
+    applicationScanResultCount: Math.max(0, scanResults.length - 1),
+    scanResultPayloadBytes: scanResults.map(bytes),
+    totalScanResultsPayloadBytes: bytes(scanResults),
+  };
+}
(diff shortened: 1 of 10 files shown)


kateeselius (commented):
Possible other metrics to include: 

(probably in response-builder.ts) 
- number of scan results (os+ n application scan results) 
- size of each scan result + overall size of the payload

lib/scan-payload-metrics.ts
```diff
@@ -0,0 +1,21 @@
+import type { ScanResult } from "./types";
+
+export interface ScanPayloadMetrics {
```
kateeselius, line 3: this is epic - so helpful!

lib/types.ts
```diff
@@ -84,9 +84,15 @@ export type FactType =
...
 
+export interface PluginAnalytics {
+  name: string;
+  data: unknown;
+}
+
 export interface PluginResponse {
   /** The first result is guaranteed to be the OS dependencies scan result. */
   scanResults: ScanResult[];
+  analytics?: PluginAnalytics[];
```
kateeselius, line 95: does adding a field here have implications for downstream consumers (k8s-monitor, dra, cli? / other users?)

would we consider adding the timings as a fact instead?

kateeselius (commented):
Looks good - one question is about adding the analytics field to the plugin response - would this cause problems for downstream callers? 

Have we tested with cli?

lib/types.ts
bdemeo12: I don't think we need to do that! Analytics is optional so it won't break any consumers. Also, IaC has a pattern of collecting per-phase latency metrics and sending them through CLI analytics, which is what this is based off of! 

I tested this as did the CLI team: https://github.com/snyk/cli/pull/6752

kateeselius: got it - cool! Makes sense

kateeselius (approved):
LGTM

PR #858, opened 2026-06-09
feat: add per-layer OS package diff attribution (algorithm + unit tests)

SteveShani:
- [ ] Ready for review

#### What does this PR do?

First PR in the vulns-by-layer stack. It adds the pure layer-attribution algorithm and its unit tests, with no wiring into the scan path yet — so this change is purely additive and `scan()` output is byte-identical to today.

`lib/analyzer/layer-attribution.ts` computes, for each OS package that survives in the final image, the diffID of the layer that introduced it. It does this by walking the per-layer package DBs (apk / apt / rpm / chisel) and diffing consecutive snapshots:

- `computeOsPackageManagerLayerAttribution` — per-ecosystem diff producing a `<source>/<binary>@<version>` → diffID map for surviving packages. Distinguishes "layer has no package DB" (skip, state unchanged) from "empty DB" (wipe, so reinstalls re-attribute to the latest layer), and drops packages removed and not reinstalled.
- `computeOsLayerAttribution` — image-wide orchestrator that dedupes by ecosystem, merges the per-PM maps, and returns any non-fatal `warnings` (per-ecosystem failure or cross-ecosystem key collision) as data alongside the map.
- `checkHistoryAlignment` — returns a user-facing warning when the image's non-empty `history` count doesn't match `rootfs.diff_ids[]`, i.e. when the backend's later `diffID → createdBy` join can't be done reliably. Attribution itself is unaffected; this only flags that instruction text may be missing in the UI.

Supporting changes: exports `depFullName` from `lib/dependency-tree` so attribution keys match dep-graph node names exactly (the join the response builder will rely on in a later PR), and adds the `IntroducingLayerByPackage` type.

Keys are deliberately produced via the shared `depFullName` so they line up 1:1 with dep-graph node names — that's the load-bearing correctness decision.

Read it bottom-to-top to follow the call flow: `computeOsLayerAttribution` (orchestrator) → `computeOsPackageManagerLayerAttribution`(per-ecosystem diff) → `parseLayerOsPackages` → the small helpers. `checkHistoryAlignment` is independent and can be read separately. The unit tests are grouped in the same order as the public functions and double as the edge-case spec.

#### Additional questions


SteveShani:
* **#860** <a href="https://app.graphite.com/github/pr/snyk/snyk-docker-plugin/860?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#859** <a href="https://app.graphite.com/github/pr/snyk/snyk-docker-plugin/859?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#858** <a href="https://app.graphite.com/github/pr/snyk/snyk-docker-plugin/858?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/snyk/snyk-docker-plugin/858?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

lib/analyzer/layer-attribution.ts
```diff
@@ -0,0 +1,376 @@
...
+        await apkAnalyze(targetImage, getApkDbFileContent(layer)),
+      );
+    }
+
+    case AnalysisType.Apt: {
+      if (!layerHasAction(layer, getDpkgFileContentAction.actionName)) {
+        return null;
+      }
+      return pkgKeySetFromAnalyses(
+        await aptAnalyze(targetImage, getAptDbFileContent(layer), osRelease),
```
mtstanley-snyk, line 144: do we also need to handle distroless apt here?

lib/analyzer/layer-attribution.ts
```diff
@@ -0,0 +1,376 @@
...
+ * The top-level `Promise.all` in `static-analyzer.ts` produces one
+ * `ImagePackagesAnalysis` per DB *format* — e.g. RPM BDB/NDB and RPM SQLite
+ * are separate results, both tagged `AnalysisType.Rpm`; regular APT and
+ * distroless APT are separate results, both tagged `AnalysisType.Apt`. The
+ * per-PM helper is keyed on `AnalysisType` and already reads every DB
+ * format for that ecosystem per layer, so we must call it once per
+ * *ecosystem*, not once per *result* — otherwise duplicate ecosystem-level
+ * invocations would write the same key twice.
+ *
+ * Cross-ecosystem key collisions are not expected in practice — each OS
```
mtstanley-snyk: Makes sense that having 2 separate ecosystems in the same image is unlikely but we do correctly produce 2 subtrees currently in the dep graph, right? So seems like we should account for this, which I believe you are, but I'm not fully following why the dep name key we're using is unlikely to have collisions across ecosystems? Would it be safer to include the purl or something to ensure this doesn't happen?

Just looking to understand why we're certain this `PM produces a disjoint `<source>/<binary>` namespace` is the case.

mtstanley-snyk (commented):
Just a couple qs, otherwise looks great! Nice work 🚀

lib/analyzer/layer-attribution.ts
SteveShani: That is missing, you're right. However, looking into it, the implementation for distroless apt is different than normal apt. The standard DB file parsing and comparison won't work with distroless. If you're okay with it, I would rather create a ticket and defer attribution support for distroless to a separate PR.

lib/analyzer/layer-attribution.ts
SteveShani: > we do correctly produce 2 subtrees currently in the dep graph, right?

So turns out, we don't.
Looking at lib/parser/index.ts:
```typescript
export function parseAnalysisResults(
  targetImage: string,
  analysis: StaticPackagesAnalysis,
): AnalysisInfo {
  let analysisResult = analysis.results.filter((res) => {
    return res.Analysis && res.Analysis.length > 0;
  })[0];
```

We do get an analysis for each ecosystem / package manager, but then only take the first option.
From static analyzer:
```
results = await Promise.all([
      apkAnalyze(targetImage, apkDbFileContent),
      aptAnalyze(targetImage, aptDbFileContent, osRelease),
      rpmAnalyze(...),
      mapRpmSqlitePackages(...),
      aptDistrolessAnalyze(targetImage, distrolessAptFiles, osRelease),
      chiselAnalyze(targetImage, chiselPackages),
    ]);
```
This would be the "order of importance".

In the PR, I included something I didn't need to. I actually simplified the code a little bit since multi package managers aren't needed anymore.

Regarding the PURL, the reason I kept the key as <depFullName>@<version> is that it is the dep-graph node's name, so annotation is a direct key lookup against nodes we already hold, with no extra derivation. Given the single-ecosystem guarantee above, that name is already unambiguous within the graph we're annotating, so a purl wouldn't add safety here.

lib/analyzer/layer-attribution.ts
```diff
@@ -0,0 +1,365 @@
...
+        analyses.push(
+          await rpmAnalyze(
+            targetImage,
+            [...bdbPkgs, ...ndbPkgs],
+            redHatRepositories,
+            osRelease,
+          ),
+        );
+      }
+      if (hasSqlite) {
```
jcambier, line 180: info: Looking at the ExtractActions for rpm, it looks like `"/usr/lib/sysimage/rpm/Packages.db"` is detected by both the `"rpm-sqlite-db"` action and the `"rpm-ndb"` action. For images with this file, do you think there would be any problems if both `hasNdb` and `hasSqlite` were true?

SteveShani: Huh interesting. It's a bit odd that the rpm-sqlite-db action even looks at the `/usr/lib/sysimage/rpm/Packages.db` file. It would make sense to remove it from the list, but I'm not sure why it was originally added. In any case, it shouldn't be a problem. Since it won't be able to parse it, and won't return any packages in that instance, only the correct action matches and populates the package name -> layer ID set.

jcambier (approved):

PR #861, opened 2026-06-10
feat: add apkPackageOwnership for Chainguard applications

LSTu5:
- [x] Ready for review

## What does this PR do?

Adds a new `apkPackageOwnership` fact to application `ScanResult`s on Wolfi/Chainguard images. For each detected app dependency, the plugin resolves evidence paths (`identity.targetFile`, `testedFiles`, `jarFingerprints`) to the APK package that owns those files on disk.

**Supporting changes:**

- **APK DB parsing** — extend `apk.ts` to parse `F:`/`R:`/`M:` records from the installed database into per-package `Files` and `Directories` lists.
- **Ownership resolution** — new `apk-ownership.ts` builds a path index (exact file map + directory trie), canonicalizes paths through symlinks, and picks a consistent owning package across all evidence paths.
- **Symlink plumbing** — extractors now collect per-layer symlinks and pass them through static analysis into `buildResponse`.
- **Scope** — only runs on Wolfi/Chainguard distros (`osRelease.name` is `wolfi` or `chainguard`); skipped when evidence paths cannot be resolved or owners disagree ambiguously.

This is a **public API change**: `apkPackageOwnership` is a new `FactType` consumed downstream for app/OS vulnerability reconciliation.

## Where should the reviewer start?

1. `lib/analyzer/package-managers/apk-ownership.ts` — core ownership logic (`buildApkPathIndex`, `resolveApkOwnership`, distro gating)
2. `lib/response-builder.ts` (~lines 279–295) — where the fact is attached to app scan results
3. `lib/analyzer/package-managers/apk.ts` — installed-db file-list parsing (`F`/`R`/`M` records)
4. `test/unit/apk-ownership.spec.ts` and `test/unit/response-builder-ownership.spec.ts` — expected matching behavior (exact vs directory match, symlink canonicalization, multi-path consistency)
5. `lib/analyzer/applications/evidence-paths.ts` — which paths are considered evidence

### Prerequisites

- Node 20+ (`npm install`; run `npm run build` only if exercising `dist/` directly)
- **System test:** running Docker daemon (containerd image store disabled — see `test/README.md`)

### Quick run (all PR-specific tests)

Use explicit file paths — pattern args like `npm run test:unit -- apk-ownership` match unrelated suites.

```bash
npx jest \
  test/unit/apk-ownership.spec.ts \
  test/unit/response-builder-ownership.spec.ts \
  test/unit/path-canonicalization.spec.ts \
  test/unit/evidence-paths.spec.ts \
  test/unit/apk-file-list.spec.ts \
  test/system/application-scans/chainguard-ownership.spec.ts
```

## Background

Chainguard/Wolfi apps are often installed as APK packages. Downstream needs to know which OS package owns an app's files to reconcile app and OS vulnerabilities. Ownership is only emitted when all evidence paths resolve to the same package.

## Tickets
CN-1395

### DIFF
--- a/lib/analyzer/package-managers/apk-ownership.ts
+++ b/lib/analyzer/package-managers/apk-ownership.ts
@@ -111,24 +111,20 @@ function resolveDirectoryOwner(
     node = child;
     if (node.owners.length > 0) {
       best = {
-        owner: pickDirectoryOwner(node.owners),
+        owner: pickExactOwner(node.owners),
         matchKind: "directory",
         prefixLength: i + 1,
       };
     }
   }
 
-  if (!best && node.owners.length > 0) {
-    best = {
-      owner: pickDirectoryOwner(node.owners),
-      matchKind: "directory",
-      prefixLength: segments.length,
-    };
-  }
-
   return best;
 }
 
+/**
+ * Resolve APK package ownership for app evidence paths on Wolfi/Chainguard images.
+ * Every evidence path must match an owner; we skip the fact rather than guess.
+ */
 export function resolveApkOwnership(
   evidencePaths: string[],
   packages: AnalyzedPackageWithVersion[],
@@ -145,6 +141,7 @@ export function resolveApkOwnership(
   for (const evidencePath of evidencePaths) {
     const normalized = normalizeAbsolutePath(evidencePath);
     const match = resolveOwnerForEvidencePath(normalized, index, symlinkGraph);
+    // Require every evidence path to resolve; partial matches are not emitted.
     if (!match) {
       return undefined;
     }
@@ -190,6 +187,7 @@ function pickBestOwnerAcrossPaths(
 
   for (const match of matches) {
     const key = `${match.owner.Name}@${match.owner.Version}`;
+    // Prefer exact file matches over directory-prefix matches when paths disagree.
     const exactBonus = match.matchKind === "exact" ? 1000 : 0;
     const score = exactBonus + match.prefixLength;
     const existing = scores.get(key);
@@ -221,12 +219,6 @@ function pickExactOwner(
   }
   const originMatch = owners.find((o) => o.Name === o.Source);
   return originMatch ?? owners[0];
-}
-
-function pickDirectoryOwner(
-  owners: AnalyzedPackageWithVersion[],
-): AnalyzedPackageWithVersion {
-  return pickExactOwner(owners);
 }
 
 function addToOwnerMap(
--- a/lib/analyzer/package-managers/apk-ownership.ts
+++ b/lib/analyzer/package-managers/apk-ownership.ts
@@ -1,0 +1,270 @@
+import { SymlinkMap } from "../../extractor/types";
+import {
+  AnalysisType,
+  AnalyzedPackageWithVersion,
+  ImageAnalysis,
+  ImagePackagesAnalysis,
+  OSRelease,
+} from "../types";
+import {
+  canonicalizePath,
+  normalizeAbsolutePath,
+  SymlinkGraph,
+} from "./path-canonicalization";
+
+export interface ApkPackageOwnership {
+  distroId: string;
+  packageName: string;
+  packageVersion: string;
+  originPackage: string;
+  evidencePaths: string[];
+}
+
+export type MatchKind = "exact" | "directory";
+
+export interface PathOwnerMatch {
+  owner: AnalyzedPackageWithVersion;
+  matchKind: MatchKind;
+  prefixLength: number;
+}
+
+interface DirectoryTrieNode {
+  owners: AnalyzedPackageWithVersion[];
+  children: Map<string, DirectoryTrieNode>;
+}
+
+export interface ApkPathIndex {
+  exactFileOwners: Map<string, AnalyzedPackageWithVersion[]>;
+  directoryTrie: DirectoryTrieNode;
+}
+
+const CHAINGUARD_DISTROS = new Set(["wolfi", "chainguard"]);
+
+export function isChainguardDistro(osRelease: OSRelease): boolean {
+  return CHAINGUARD_DISTROS.has(osRelease.name);
+}
+
+export function toSymlinkGraph(symlinks?: SymlinkMap): SymlinkGraph {
+  const graph: SymlinkGraph = new Map();
+  if (!symlinks) {
+    return graph;
+  }
+  for (const [symlinkPath, target] of Object.entries(symlinks)) {
+    graph.set(normalizeAbsolutePath(symlinkPath), target);
+  }
+  return graph;
+}
+
+export function buildApkPathIndex(
+  packages: AnalyzedPackageWithVersion[],
+  symlinkGraph: SymlinkGraph,
+): ApkPathIndex {
+  const exactFileOwners = new Map<string, AnalyzedPackageWithVersion[]>();
+  const directoryTrie: DirectoryTrieNode = { owners: [], children: new Map() };
+
+  for (const pkg of packages) {
+    for (const filePath of pkg.Files ?? []) {
+      const canonical = canonicalizePath(filePath, symlinkGraph);
+      addToOwnerMap(exactFileOwners, canonical, pkg);
+    }
+
+    for (const dirPath of pkg.Directories ?? []) {
+      const canonical = canonicalizePath(dirPath, symlinkGraph);
+      insertDirectoryOwner(directoryTrie, canonical, pkg);
+    }
+  }
+
+  return { exactFileOwners, directoryTrie };
+}
+
+export function resolveOwnerForEvidencePath(
+  evidencePath: string,
+  index: ApkPathIndex,
+  symlinkGraph: SymlinkGraph,
+): PathOwnerMatch | undefined {
+  const canonical = canonicalizePath(evidencePath, symlinkGraph);
+  const exactOwners = index.exactFileOwners.get(canonical);
+  if (exactOwners && exactOwners.length > 0) {
+    return {
+      owner: pickExactOwner(exactOwners),
+      matchKind: "exact",
+      prefixLength: canonical.length,
+    };
+  }
+
+  return resolveDirectoryOwner(canonical, index);
+}
+
+function resolveDirectoryOwner(
+  canonicalPath: string,
+  index: ApkPathIndex,
+): PathOwnerMatch | undefined {
+  const segments = canonicalPath.split("/").filter(Boolean);
+  let node = index.directoryTrie;
+  let best: PathOwnerMatch | undefined;
+
+  for (let i = 0; i < segments.length; i++) {
+    const child = node.children.get(segments[i]);
+    if (!child) {
+      break;
+    }
+    node = child;
+    if (node.owners.length > 0) {
+      best = {
+        owner: pickDirectoryOwner(node.owners),
+        matchKind: "directory",
+        prefixLength: i + 1,
+      };
+    }
+  }
+
+  if (!best && node.owners.length > 0) {
+    best = {
+      owner: pickDirectoryOwner(node.owners),
+      matchKind: "directory",
+      prefixLength: segments.length,
+    };
+  }
+
+  return best;
+}
+
+export function resolveApkOwnership(
+  evidencePaths: string[],
+  packages: AnalyzedPackageWithVersion[],
+  osRelease: OSRelease,
+  symlinkGraph: SymlinkGraph,
+): ApkPackageOwnership | undefined {
+  if (!isChainguardDistro(osRelease) || evidencePaths.length === 0) {
+    return undefined;
+  }
+
+  const index = buildApkPathIndex(packages, symlinkGraph);
+  const perPathMatches: PathOwnerMatch[] = [];
+
+  for (const evidencePath of evidencePaths) {
+    const normalized = normalizeAbsolutePath(evidencePath);
+    const match = resolveOwnerForEvidencePath(normalized, index, symlinkGraph);
+    if (!match) {
+      return undefined;
+    }
+    perPathMatches.push(match);
+  }
+
+  const owner = pickConsistentOwner(perPathMatches);
+  if (!owner) {
+    return undefined;
+  }
+
+  return {
+    distroId: osRelease.name,
+    packageName: owner.Name,
+    packageVersion: owner.Version,
+    originPackage: owner.Source ?? owner.Name,
+    evidencePaths,
+  };
+}
+
+function pickConsistentOwner(
+  matches: PathOwnerMatch[],
+): AnalyzedPackageWithVersion | undefined {
+  const ownerKey = (pkg: AnalyzedPackageWithVersion) =>
+    `${pkg.Name}@${pkg.Version}`;
+
+  const firstKey = ownerKey(matches[0].owner);
+  const allSame = matches.every((m) => ownerKey(m.owner) === firstKey);
+  if (allSame) {
+    return matches[0].owner;
+  }
+
+  return pickBestOwnerAcrossPaths(matches);
+}
+
+function pickBestOwnerAcrossPaths(
+  matches: PathOwnerMatch[],
+): AnalyzedPackageWithVersion | undefined {
+  const scores = new Map<
+    string,
+    { pkg: AnalyzedPackageWithVersion; score: number }
+  >();
+
+  for (const match of matches) {
+    const key = `${match.owner.Name}@${match.owner.Version}`;
+    const exactBonus = match.matchKind === "exact" ? 1000 : 0;
+    const score = exactBonus + match.prefixLength;
+    const existing = scores.get(key);
+    if (!existing || score > existing.score) {
+      scores.set(key, { pkg: match.owner, score });
+    }
+  }
+
+  const entries = [...scores.values()];
+  if (entries.length === 0) {
+    return undefined;
+  }
+
+  entries.sort((a, b) => b.score - a.score);
+  const topScore = entries[0].score;
+  const topEntries = entries.filter((e) => e.score === topScore);
+  if (topEntries.length === 1) {
+    return topEntries[0].pkg;
+  }
+
+  return undefined;
+}
+
+function pickExactOwner(
+  owners: AnalyzedPackageWithVersion[],
+): AnalyzedPackageWithVersion {
+  if (owners.length === 1) {
+    return owners[0];
+  }
+  const originMatch = owners.find((o) => o.Name === o.Source);
+  return originMatch ?? owners[0];
+}
+
+function pickDirectoryOwner(
+  owners: AnalyzedPackageWithVersion[],
+): AnalyzedPackageWithVersion {
+  return pickExactOwner(owners);
+}
+
+function addToOwnerMap(
+  map: Map<string, AnalyzedPackageWithVersion[]>,
+  canonicalPath: string,
+  pkg: AnalyzedPackageWithVersion,
+): void {
+  const existing = map.get(canonicalPath) ?? [];
+  existing.push(pkg);
+  map.set(canonicalPath, existing);
+}
+
+function insertDirectoryOwner(
+  root: DirectoryTrieNode,
+  dirPath: string,
+  pkg: AnalyzedPackageWithVersion,
+): void {
+  const segments = dirPath.split("/").filter(Boolean);
+  let node = root;
+  for (const segment of segments) {
+    let child = node.children.get(segment);
+    if (!child) {
+      child = { owners: [], children: new Map() };
+      node.children.set(segment, child);
+    }
+    node = child;
+  }
+  node.owners.push(pkg);
+}
+
+export function getApkPackagesFromResults(
+  results?: ImageAnalysis[],
+): AnalyzedPackageWithVersion[] {
+  if (!results) {
+    return [];
+  }
+  const apkResult = results.find((r) => r.AnalyzeType === AnalysisType.Apk) as
+    | ImagePackagesAnalysis
+    | undefined;
+  return apkResult?.Analysis ?? [];
+}
--- a/lib/analyzer/package-managers/apk-ownership.ts
+++ b/lib/analyzer/package-managers/apk-ownership.ts
@@ -1,3 +1,4 @@
+import * as Debug from "debug";
 import { SymlinkMap } from "../../extractor/types";
 import {
   AnalysisType,
@@ -25,6 +26,7 @@ export type MatchKind = "exact" | "directory";
 export interface PathOwnerMatch {
   owner: AnalyzedPackageWithVersion;
   matchKind: MatchKind;
+  /** Number of matched path segments; a deeper directory match outranks a shallower one. */
   prefixLength: number;
 }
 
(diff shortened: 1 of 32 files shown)


lib/analyzer/package-managers/path-canonicalization.ts
```diff
@@ -0,0 +1,88 @@
...
+    current = normalizeSymlinkTarget(current, linkTarget);
+  }
+
+  return current;
+}
+
+/**
+ * Merge symlink maps from multiple layers. Later entries override earlier ones.
+ */
+export function mergeSymlinkGraphs(layers: SymlinkGraph[]): SymlinkGraph {
```
parker-snyk: I may be overlooking something, but where is this being called?

lib/analyzer/package-managers/apk-ownership.ts
```diff
@@ -0,0 +1,270 @@
...
+      best = {
+        owner: pickDirectoryOwner(node.owners),
+        matchKind: "directory",
+        prefixLength: i + 1,
+      };
+    }
+  }
+
+  if (!best && node.owners.length > 0) {
+    best = {
```
parker-snyk: This might be unreachable, if `node.owners.length` > 0, `best` will be set from line 113 it looks like?

lib/analyzer/package-managers/apk-ownership.ts
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  const index = buildApkPathIndex(packages, symlinkGraph);
+  const perPathMatches: PathOwnerMatch[] = [];
+
+  for (const evidencePath of evidencePaths) {
+    const normalized = normalizeAbsolutePath(evidencePath);
+    const match = resolveOwnerForEvidencePath(normalized, index, symlinkGraph);
+    if (!match) {
+      return undefined;
```
parker-snyk, line 159: I'm still going through this, so I might find the answer in a bit - but do we want to return undefined if there's ever not a match? In other words, is it possible to have a match for one evidencePath, but not another, and if so, do we want to get the owner of the one that DOES exist?

lib/analyzer/package-managers/apk-ownership.ts
```diff
@@ -0,0 +1,270 @@
...
+  matches: PathOwnerMatch[],
+): AnalyzedPackageWithVersion | undefined {
+  const scores = new Map<
+    string,
+    { pkg: AnalyzedPackageWithVersion; score: number }
+  >();
+
+  for (const match of matches) {
+    const key = `${match.owner.Name}@${match.owner.Version}`;
+    const exactBonus = match.matchKind === "exact" ? 1000 : 0;
```
parker-snyk: What is this for?

lib/analyzer/package-managers/apk-ownership.ts
```diff
@@ -0,0 +1,270 @@
...
+  owners: AnalyzedPackageWithVersion[],
+): AnalyzedPackageWithVersion {
+  if (owners.length === 1) {
+    return owners[0];
+  }
+  const originMatch = owners.find((o) => o.Name === o.Source);
+  return originMatch ?? owners[0];
+}
+
+function pickDirectoryOwner(
```
parker-snyk: Whats the purpose for this function, it looks like it's just calling `pickExactOwner`?

lib/analyzer/package-managers/apk.ts
```diff
@@ -15,55 +15,98 @@ export function analyze(
...
-  }
-  return pkgs;
-}
+    if (line.length < 2) {
+      continue;
+    }
+    const key = line.charAt(0);
+    const value = line.substr(2).trim();
 
-function parseLine(
```
parker-snyk, line 27: Nice change merging in parseLine to the caller

parker-snyk (changes requested):
Good start! Also, minor nit, consider adding comments in some places. There are a couple files with no comments, where it might be beneficial to explain what's going on or why we're taking a certain approach

lib/extractor/index.ts
```diff
@@ -226,6 +228,22 @@ export function getUserInstructionLayersFromConfig(imageConfig) {
...
+  symlinkLayers?: SymlinkMap[],
+): SymlinkMap | undefined {
+  if (!symlinkLayers || symlinkLayers.length === 0) {
+    return undefined;
+  }
+
+  const merged: SymlinkMap = {};
+  for (const layer of symlinkLayers) {
+    for (const [symlinkPath, target] of Object.entries(layer)) {
+      merged[symlinkPath] = target;
```
parker-snyk: I may be overlooking something, but note the layers arrive here newest-first; the extractors `.reverse()` before returning (the OCI one documents "from last to first"), and `layersWithLatestFileModifications` just below is built around that: it only inserts a path it hasn't seen yet, so the newest layer wins. This loop overwrites instead. If the same symlink path exists in two layers, which one ends up in the merged map - and is that the one that exists in the running container?

It also makes me wonder whether we need a second merge with its own semantics at all. Symlinks are tar entries like everything else this pipeline already handles, so is there a way for them to ride through the existing per-layer record instead of a parallel one?

Related: the existing function also does some bookkeeping with `isWhitedOutFile` that has no equivalent here. What happens to a symlink that a later layer deletes?

A two-layer test image where the upper layer replaces - and another where it deletes. A base-layer symlink would settle the behavior questions either way.

lib/extractor/layer.ts
```diff
@@ -22,17 +22,29 @@ const debug = Debug("snyk");
...
     const result: ExtractedLayers = {};
+    const symlinks: SymlinkMap = {};
     const tarExtractor: Extract = extract();
 
     tarExtractor.on("entry", async (headers, stream, next) => {
-      if (headers.type === "file") {
-        const absoluteFileName = path.join(path.sep, headers.name);
+      const absoluteFileName = path.join(path.sep, headers.name);
+
+      if (headers.type === "symlink" || headers.type === "link") {
```
parker-snyk: Symlinks and hard links use different rules for joining. A symlinks path join rule is to join the links directory and the target. A hard link's rule is to join the system root (/) with the target. Since both are being added to an array, but we're only adding the linkTarget, we're losing infomation about the type of the link that we are passing. So we should consider either tracking the type or performing the path join here while we know the type

lib/analyzer/package-managers/apk.ts
```diff
@@ -15,55 +15,98 @@ export function analyze(
...
-      break;
-    case "V": // Version
-      curPkg.Version = value;
-      break;
-    case "p": // Provides
-      for (let name of value.split(" ")) {
-        name = name.split("=")[0];
-        curPkg.Provides.push(name);
+    switch (key) {
+      case "P": {
```
parker-snyk, line 32: Nit - no reason to remove the comments that were originally here, ie // Package, // Version, etc

lib/analyzer/package-managers/apk.ts
```diff
@@ -15,55 +15,98 @@ export function analyze(
...
+        if (curPkg) {
+          curPkg.Source = value;
+        }
+        break;
+      case "F": {
+        if (!curPkg) {
+          break;
+        }
+        dirStack = value.split("/").filter(Boolean);
+        const dirPath = stackToAbsolutePath(dirStack);
```
parker-snyk: I may be following this incorrectly, but it looks like we're splitting by `/` and then immediately joining by `/` after?

parker-snyk (changes requested):

lib/analyzer/package-managers/path-canonicalization.ts
LSTu5: Not called, will remove

lib/analyzer/package-managers/apk-ownership.ts
LSTu5: removed, unreachable

lib/analyzer/package-managers/apk-ownership.ts
LSTu5: remove

lib/analyzer/package-managers/apk.ts
LSTu5: Added back

lib/extractor/index.ts
```diff
@@ -226,6 +231,62 @@ export function getUserInstructionLayersFromConfig(imageConfig) {
...
+    }
+
+    // Register this layer's whiteouts only after its own entries are merged:
+    // they delete paths from older layers, not from this layer or newer ones.
+    const fileLayer = fileLayers?.[i];
+    if (fileLayer) {
+      for (const filename of Object.keys(fileLayer)) {
+        if (isWhitedOutFile(filename)) {
+          removedPathsToIgnore.add(filename.replace(/.wh./, ""));
+        }
```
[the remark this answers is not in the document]
parker-snyk: This is correct, we have another PR pending for some related whiteout stuff, but we should treat the `.` as a literal, otherwise you can match to non white out files

lib/response-builder.ts
```diff
@@ -261,6 +267,33 @@ async function buildResponse(
...
     appDepsScanResult.facts.push(appPluginVersionFact);
 
+    const osRelease = depsAnalysis.osRelease ?? depsAnalysis.depTree.targetOS;
+    if (osRelease?.prettyName) {
+      const imageOsReleasePrettyNameFact: facts.ImageOsReleasePrettyNameFact = {
+        type: "imageOsReleasePrettyName",
+        data: osRelease.prettyName,
+      };
+      appDepsScanResult.facts.push(imageOsReleasePrettyNameFact);
+    }
```
parker-snyk, line 277: This emits `imageOsReleasePrettyName` on every app scan result for every distro, not just Chainguard — it's where almost all the snapshot churn in this PR comes from. The OS scan result already carries the same fact (line 197 above), and the ownership fact already includes `distroId`.

Per AGENTS.md, changing the facts on a ScanResult is a contract change for every consumer (CLI, kubernetes-monitor, registry agent, ...), and the PR description only calls out `apkPackageOwnership` as the API change. Can we either gate this to Chainguard/Wolfi or drop it and rely on the OS result's fact? If downstream genuinely needs it on app results for all distros, it should at least be listed in the description as a second API change.

*— AI-assisted*

lib/response-builder.ts
```diff
@@ -261,6 +267,33 @@ async function buildResponse(
...
+    const evidencePaths = extractEvidencePaths(appDepsScanResult);
+    const apkPackages = getApkPackagesFromResults(depsAnalysis.results);
+    const ownership =
+      osRelease &&
+      resolveApkOwnership(
+        evidencePaths,
+        apkPackages,
+        osRelease,
+        toSymlinkGraph(depsAnalysis.symlinks),
+      );
```
parker-snyk, line 288: This block runs once per app scan result, and `resolveApkOwnership` rebuilds the whole path index from scratch on every call — `buildApkPathIndex` re-canonicalizes every file of every APK package, and `toSymlinkGraph` rebuilds the Map too. A Chainguard image with a pile of go binaries or jar fingerprints redoes all that work N times for identical inputs.

Worth hoisting: build the symlink graph and the index once before the loop (behind the `isChainguardDistro` check) and pass them into the resolver.

*— AI-assisted*

test/lib/extractor/layer.spec.ts
```diff
@@ -14,7 +15,7 @@ describe("layer extractor: layer contain bad elf file", () => {
       getFixture("extracted-layers/layer_with_bad_elf_file.tar"),
     );
     const p = extractImageLayer(stream, staticAnalysisActions);
-    expect(p).toMatchObject({});
+    expect(p).resolves.toMatchObject({ extractedLayers: {}, symlinks: {} });
```
parker-snyk, line 18: Two problems with this assertion (and the same one on line 30):

1. It isn't awaited. `expect(p).resolves...` returns a promise, and since the test callback isn't async and doesn't return it, jest finishes the test without ever waiting for the assertion to run.
2. `toMatchObject({ extractedLayers: {}, symlinks: {} })` is vacuous — an empty object in `toMatchObject` matches *any* object, so this passes even if `extractedLayers` came back full of entries.

The old `expect(p).toMatchObject({})` was equally broken, so this was a chance to actually fix it. Suggest making the test async and doing `await expect(p).resolves.toEqual({ extractedLayers: {}, symlinks: {} })`, which asserts both shape and emptiness.

*— AI-assisted*

lib/go-parser/index.ts
```diff
@@ -199,8 +199,12 @@ export async function goModulesToScannedProjects(
         type: "depGraph",
         data: depGraph,
       };
+      const testedFilesFact: TestedFilesFact = {
+        type: "testedFiles",
+        data: [filePath],
+      };
       scanResults.push({
-        facts: [depGraphFact],
+        facts: [depGraphFact, testedFilesFact],
```
parker-snyk, line 207: `data: [filePath]` is always the same value as `identity.targetFile` a few lines down, and `extractEvidencePaths` already collects `targetFile`, so ownership resolution doesn't actually need this fact. Meanwhile it changes the output for every consumer scanning go binaries (see the gomodules/sles15/windows snapshot churn).

If the intent is consistency with how node/php emit `testedFiles`, that's reasonable — but then it belongs in the PR description as an output change. Otherwise I'd drop it.

*— AI-assisted*

test/system/application-scans/chainguard-ownership.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  it("does not attach apkPackageOwnership on OS-only Chainguard images", async () => {
+    const image =
+      "chainguard/bash@sha256:642933df66209814502599053ca3dfa97cccf847badc4219d2b1fd6565f6559a";
+    const pluginResult = await scan({
+      path: image,
+      platform: "linux/amd64",
+    });
+
+    const appResults = pluginResult.scanResults.slice(1);
```
parker-snyk, line 26: Two things here. `slice(1)` assumes the OS result is always index 0 without asserting it, so a reorder upstream would quietly turn this into a test of nothing — finding app results by identity type (or asserting result 0 is the OS one) would be safer.

Bigger picture: this is the only system test for the feature and it covers the negative case. Nothing end-to-end proves an `apkPackageOwnership` fact actually gets attached — the positive path only has unit coverage with hand-built fixtures. Could we add a positive case against a Chainguard image that ships an app as an APK package (one of their language images, pinned by digest like this one)?

*— AI-assisted*

lib/extractor/index.ts
```diff
@@ -226,6 +231,62 @@ export function getUserInstructionLayersFromConfig(imageConfig) {
...
+      }
+    }
+
+    // Register this layer's whiteouts only after its own entries are merged:
+    // they delete paths from older layers, not from this layer or newer ones.
+    const fileLayer = fileLayers?.[i];
+    if (fileLayer) {
+      for (const filename of Object.keys(fileLayer)) {
+        if (isWhitedOutFile(filename)) {
+          removedPathsToIgnore.add(filename.replace(/.wh./, ""));
```
parker-snyk, line 272: The dots in `/.wh./` aren't escaped, so this matches any `<char>wh<char>` sequence, not just a literal `.wh.` marker. A real file at `/app/swhole` matches (on `swho`), the replace turns it into `/app/le`, and that bogus path lands in `removedPathsToIgnore` — after which `isSymlinkInARemovedFolder` will hide any symlink under `/app/le`.

I realize this mirrors the existing file-merge code (line 303 has the same pattern), so the regex bug is pre-existing. But the new symlink path prefix-matches against the stripped values, which widens the blast radius beyond what the file merge had. A literal `filename.replace(".wh.", "")` plus an anchored check on the basename would fix the new code; cleaning up the old occurrence while you're in here would be a bonus.

*— AI-assisted*

lib/analyzer/package-managers/path-canonicalization.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+function normalizeSymlinkTarget(basePath: string, linkTarget: string): string {
+  const target = linkTarget.replace(/\\/g, "/");
+  if (target.startsWith("/")) {
+    return path.normalize(target);
+  }
+  const baseDir = path.dirname(basePath);
+  return path.normalize(path.join(baseDir, target));
+}
```
parker-snyk, line 54: nit: this is the same math as `absoluteLinkTarget` in `lib/extractor/layer.ts` (absolute target gets normalized, relative target resolves against the link's own directory). Two copies of symlink-target resolution will drift apart eventually — worth extracting one shared helper.

*— AI-assisted*

lib/analyzer/package-managers/apk-ownership.ts
```diff
@@ -0,0 +1,263 @@
...
+  index: ApkPathIndex,
+  symlinkGraph: SymlinkGraph,
+): PathOwnerMatch | undefined {
+  const canonical = canonicalizePath(evidencePath, symlinkGraph);
+  const exactOwners = index.exactFileOwners.get(canonical);
+  if (exactOwners && exactOwners.length > 0) {
+    return {
+      owner: pickExactOwner(exactOwners),
+      matchKind: "exact",
+      prefixLength: canonical.split("/").filter(Boolean).length,
```
parker-snyk, line 95: nit: `prefixLength` on exact matches is never read — `pickBestOwnerAcrossPaths` short-circuits on `matchKind === "exact"` and only uses `prefixLength` to rank directory matches. Computing it here implies it participates in ranking when it doesn't. Either a comment saying it's unused for exact matches, or restructuring so only directory matches carry it, would make the intent clearer.

*— AI-assisted*

lib/analyzer/package-managers/apk.ts
```diff
@@ -15,55 +15,97 @@ export function analyze(
...
+          curPkg.Directories!.push(currentDir);
+        }
+        break;
+      }
+      case "R": {
+        // File name relative to the current F: directory
+        if (!curPkg || !currentDir) {
+          break;
+        }
+        curPkg.Files!.push(`${currentDir}/${value}`);
```
parker-snyk, line 94: nit: the `curPkg.Files!` / `curPkg.Directories!` assertions (here and on lines 84/102) work around the fields being optional on `AnalyzedPackage`, even though this parser always initializes them in the `P:` case. Capturing them in typed locals when the package is created would drop the `!`s and let the compiler do the checking.

*— AI-assisted*

parker-snyk (commented):
Did a pass over this with AI assistance (Claude) — inline comments on the specific lines, roll-up below.

Overall the core logic looks solid. Skipping ownership rather than guessing when evidence paths disagree is the right call given this drives downstream suppression, and the unit coverage on the matching rules is good. Most of what I found is around the edges of the change.

## Blockers

None.

## Critical

1. **`imageOsReleasePrettyName` is emitted on every app scan result for every distro** (`lib/response-builder.ts:270-277`). It duplicates the fact already on the OS result, it's an undocumented contract change for all downstream consumers, and it's the source of most of the snapshot churn in this PR. Gate it to Chainguard, drop it, or document it as a second API change.
2. **The APK path index and symlink graph are rebuilt for every app scan result** (`lib/response-builder.ts:279-288`). `buildApkPathIndex` re-canonicalizes every file of every APK package on each call. Build them once before the loop and pass them in.
3. **The reworked assertions in `test/lib/extractor/layer.spec.ts:18,30` don't assert anything.** They're unawaited, and `toMatchObject` with empty-object values matches any result. They pass no matter what `extractImageLayer` returns.

## Nice to have

4. The new `testedFiles` fact on go scan results duplicates `identity.targetFile` exactly, and ownership resolution doesn't need it — but it changes output for every consumer scanning go binaries (`lib/go-parser/index.ts:202`). Justify it in the description or drop it.
5. System coverage only proves the negative case. Add a positive system test that asserts an `apkPackageOwnership` fact actually gets attached on a Chainguard app image; also `slice(1)` in the existing test assumes result ordering without checking it (`test/system/application-scans/chainguard-ownership.spec.ts:26`).
6. The whiteout regex `/.wh./` has unescaped dots, so filenames like `/app/swhole` get treated as whiteouts. It's copied from the existing file-merge code, but the new symlink merge prefix-matches against the bogus stripped paths, which makes a false match worse there (`lib/extractor/index.ts:271-272`).

## Nits

7. `normalizeSymlinkTarget` in `path-canonicalization.ts` duplicates `absoluteLinkTarget` from `lib/extractor/layer.ts` — extract a shared helper.
8. `prefixLength` is computed for exact matches in `apk-ownership.ts:95` but never used in ranking them.
9. The `Files!` / `Directories!` non-null assertions in the apk parser are avoidable with typed locals.

Items 1-3 are the ones I'd like addressed before merge; the rest are take-it-or-leave-it.

LSTu5:
Closing in favor of a stacked split — this PR grew to ~1,530 lines across four separable concerns, so it's now five single-topic PRs that build on each other (review bottom-up):

1. #869 — `feat: parse apk file ownership records` (apk.ts F:/R:/M: parsing; two commits: refactor then feature)
2. #870 — `feat: capture and merge layer symlinks during extraction`
3. #871 — `feat: add apk ownership resolution library` (new files only, inert)
4. #872 — `feat: emit apkPackageOwnership fact for chainguard images` (the new FactType / public contract)
5. #873 — `feat: emit imageOsReleasePrettyName on app scan results` (droppable topper; the one all-distro output change)

All review feedback from this PR is already addressed in the stack. Mapping the threads:

| Thread here | Resolution | Now lives in |
| --- | --- | --- |
| [Merge loop overwrites / layer order / whiteouts](https://github.com/snyk/snyk-docker-plugin/pull/861#discussion_r3389610095) | Newest-first, first-seen-wins, OCI whiteout semantics (incl. same-layer usrmerge); two-layer replace/delete/base-layer test images added; why-a-parallel-map explained in the PR body | #870 |
| [Symlink vs hard link join rules](https://github.com/snyk/snyk-docker-plugin/pull/861#discussion_r3390090719) | Hard links excluded (alternate names, not redirects); targets resolved to absolute paths at extraction time | #870 |
| [split by `/` then join by `/`](https://github.com/snyk/snyk-docker-plugin/pull/861#discussion_r3390294057) | Fake stack replaced with a plain `currentDir` string; output byte-identical for real APK DBs | #869 |
| [`exactBonus = 1000` — what is this for?](https://github.com/snyk/snyk-docker-plugin/pull/861#discussion_r3389071739) | Score-mixing replaced with explicit rules: exact match > directory match, deepest directory wins, ambiguity → no owner | #871 |
| [Partial matches — return undefined if one path doesn't match?](https://github.com/snyk/snyk-docker-plugin/pull/861#discussion_r3389057732) | Deliberate and spec-required: evidence must be *wholly contained* ([Chainguard scanner spec](https://github.com/chainguard-dev/vulnerability-scanner-support/blob/main/docs/scanning_implementation.md)); ownership drives downstream suppression, so partial attribution could hide real vulns in user-added software. JSDoc cites the spec; mixed case locked with a test | #871 |
| "Does this cover npm/node as well?" | Yes — ownership is ecosystem-agnostic, driven by evidence paths (`identity.targetFile`, `testedFiles`, jar locations); node already emits these | #871 (`evidence-paths.ts`) |

One descope vs this branch: the go-parser `testedFiles` fact is dropped — Go scan results already carry the binary path in `identity.targetFile`, which the evidence collector reads, so it added no signal for ownership (it can return as a tiny standalone PR if we want it for node-parity).

Branch `chainguard-apk` stays untouched as the reference for comparing the stack against this diff.

PR #869, opened 2026-06-12
feat: cn-1395 parse apk file ownership records

LSTu5:

**Stack 1/5** — `main` ← **this PR** ← layer symlinks ← ownership library ← ownership fact ← prettyName. Replaces #861 (split per review feedback; see closing comment there for the thread map).

- [x] Ready for review

#### What does this PR do?

Parses the `F:`/`R:`/`M:` file-list records of the APK installed database into new optional `Files`/`Directories` fields on `AnalyzedPackage`. Two commits: a behavior-preserving refactor (inline `parseLine` into `parseFile`, typed `curPkg`, null guards), then the new record handling.

**Why:** Chainguard's scanner spec resolves app-dependency ownership by checking that an app's file paths are *wholly contained* in an APK package's declared paths. This PR supplies the "declared paths" side. It is inert on its own — nothing consumes the new fields until the ownership-library PR higher in the stack.

`lib/analyzer/package-managers/apk.ts`, commit by commit — the refactor commit changes no behavior; the feat commit is only the three new record types and the `currentDir` tracking (a plain string: `F:` lines are full root-relative paths, so there is no push/pop stack semantics).

`npx jest test/unit/apk-file-list.spec.ts` (parses the alpine 3.7 fixture DB and asserts known file/dir paths), plus `npx jest test/system/package-managers/apk.spec.ts` — the full pluginResult snapshots for alpine and chainguard/bash are unchanged, proving the new fields don't leak into existing output.

Customer escalation: scans of Chainguard images report hundreds of findings for CVEs Chainguard already patched, because app dependencies are matched against generic upstream data with no Chainguard context. The fix (per Chainguard's scanner spec) is an ownership model; this stack implements the plugin side of it.

CN-1395

#### Screenshots

N/A

#### Additional questions

N/A

### DIFF
--- a/lib/analyzer/package-managers/apk.ts
+++ b/lib/analyzer/package-managers/apk.ts
@@ -15,55 +15,62 @@ export function analyze(
   });
 }
 
-function parseFile(text: string): AnalyzedPackageWithVersion[] {
+export function parseFile(text: string): AnalyzedPackageWithVersion[] {
   const pkgs: AnalyzedPackageWithVersion[] = [];
-  let curPkg: any = null;
-  for (const line of text.split("\n")) {
-    curPkg = parseLine(line, curPkg, pkgs);
-  }
-  return pkgs;
-}
+  let curPkg: AnalyzedPackageWithVersion | null = null;
 
-function parseLine(
-  text: string,
-  curPkg: AnalyzedPackageWithVersion,
-  pkgs: AnalyzedPackageWithVersion[],
-) {
-  const key = text.charAt(0);
-  const value = text.substr(2).trim();
-  switch (key) {
-    case "P": // Package
-      curPkg = {
-        Name: value,
-        Version: "",
-        Source: undefined,
-        Provides: [],
-        Deps: {},
-        AutoInstalled: undefined,
-      };
-      pkgs.push(curPkg);
-      break;
-    case "V": // Version
-      curPkg.Version = value;
-      break;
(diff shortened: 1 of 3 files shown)


lib/analyzer/package-managers/apk.ts
```diff
@@ -15,55 +15,97 @@ export function analyze(
...
-  }
-  return pkgs;
-}
+    if (line.length < 2) {
+      continue;
+    }
+    const key = line.charAt(0);
+    const value = line.substr(2).trim();
 
-function parseLine(
```
parker-snyk, line 27: Does this need to be removed as a function? It looks like a lot of the diff is due to indentation only, which makes it harder to follow what, if anything, changed

parker-snyk: Nevermind, I got it

parker-snyk (changes requested):
Please just confirm if the bot is correct on:

~~The use of .includes() on curPkg.Directories inside the line-processing loop (for both 'F' and 'M' records) introduces O(N^2) complexity for directory deduplication within a single package. For packages with many directories, this will be significantly slower than using a Set during the parsing phase and converting to an array at the end. Since apk databases can be large, this may noticeably increase scan times for complex images.~~

Falsely flagged by the LLM ^

~~Other than possible perf impacts~~ this looks good

parker-snyk (approved):

PR #870, opened 2026-06-12
feat: cn-1395 capture and merge layer symlinks during extraction

LSTu5:

**Stack 2/5** — `main` ← apk file lists ← **this PR** ← ownership library ← ownership fact ← prettyName. Replaces the extractor portion of #861.

- [x] Ready for review

#### What does this PR do?

Records every symlink while the extractor streams each layer's tar entries, and merges the per-layer maps into a single view of the running container, exposed as `ExtractionResult.symlinks` → `StaticAnalysis.symlinks`.

**Why:** Chainguard images are usr-merged — the APK database declares `/usr/bin/node` while app evidence may say `/bin/node`, because `/bin → usr/bin` is a symlink. Ownership resolution (next PR in the stack) needs the image's symlink map to canonicalize both sides to the same form.

Design decisions, each from #861 review feedback:

- **Hard links are excluded** ([thread](https://github.com/snyk/snyk-docker-plugin/pull/861#discussion_r3390090719)): a hard link is an alternate name for the same inode, not a redirect — both names are real files and both appear in APK file lists, so they don't belong in a redirect map. Symlink targets are resolved to **absolute image paths at extraction time**, while the entry type and its directory are still known (relative targets resolve against the link's own directory; tar names are always forward-slash, so this is posix math on every platform).
- **Merge semantics mirror `layersWithLatestFileModifications`** ([thread](https://github.com/snyk/snyk-docker-plugin/pull/861#discussion_r3389610095)): layers arrive newest-first, first-seen wins. A whiteout hides paths only in *lower* (older) layers per the OCI spec — so a symlink re-created above a deletion survives, and a layer that whiteouts a directory while creating a symlink at the same path (the usrmerge pattern, exactly the `/bin → usr/bin` case this feature exists for) keeps its own symlink. A separate merge (rather than riding through `ExtractedLayers`) is deliberate: that structure only records files matching extract-action patterns, while symlinks must be collected for every path, and `getContent()` iterates all of its keys.

`lib/extractor/layer.ts` (capture + `absoluteLinkTarget`), then `symlinksWithLatestModifications` in `lib/extractor/index.ts` (merge semantics). The rest is plumbing the per-layer maps through both archive extractors.

`npx jest test/lib/extractor/` — includes in-memory merge tests for replace/delete/re-create/same-layer-whiteout/folder-whiteout, an in-memory tar test for capture (absolute targets, hard links ignored), and synthetic two-layer docker-archive tests asserting the merged map end to end.

Nothing consumes `StaticAnalysis.symlinks` until the ownership-fact PR higher in the stack; existing output (including all snapshots) is unchanged.

CN-1395

#### Screenshots

N/A

#### Additional questions

N/A

### DIFF
--- a/lib/analyzer/static-analyzer.ts
+++ b/lib/analyzer/static-analyzer.ts
@@ -171,6 +171,7 @@ export async function analyze(
     manifestLayers,
     extractedLayers,
     orderedLayers,
+    symlinks,
     rootFsLayers,
     autoDetectedUserInstructions,
     platform,
@@ -401,6 +402,7 @@ export async function analyze(
     imageCreationTime,
     containerConfig,
     history,
+    symlinks,
     timings,
   };
 }
--- a/lib/extractor/oci-archive/layer.ts
+++ b/lib/extractor/oci-archive/layer.ts
@@ -8,10 +8,12 @@ import { getErrorMessage } from "../../error-utils";
 import { streamToJson } from "../../stream-utils";
 import { PluginOptions } from "../../types";
 import { decompressMaybe } from "../decompress-maybe";
-import { extractImageLayer } from "../layer";
+import {
+  extractImageLayer,
+  LayerExtractionResult as ImageLayerExtractionResult,
+} from "../layer";
 import {
   ExtractAction,
-  ExtractedLayers,
   ExtractedLayersAndManifest,
   ImageConfig,
   OciArchiveManifest,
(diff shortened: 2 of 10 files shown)


lib/analyzer/static-analyzer.ts
```diff
@@ -35,7 +35,6 @@ import {
   getJarFileContentAction,
   getUsrLibJarFileContentAction,
 } from "../inputs/java/static";
-import { getDotnetAppFileContentAction } from "../inputs/dotnet/static";
```
parker-snyk, line 38: I think the agent split the PR without pulling in the latest main branch / rebasing. As a result it seems to be attempting to undo some of Bellas changes

lib/extractor/oci-archive/layer.ts
```diff
@@ -8,10 +8,12 @@ import { getErrorMessage } from "../../error-utils";
 import { streamToJson } from "../../stream-utils";
 import { PluginOptions } from "../../types";
 import { decompressMaybe } from "../decompress-maybe";
-import { extractImageLayer } from "../layer";
+import {
+  extractImageLayer,
+  LayerExtractionResult as ImageLayerExtractionResult,
```
parker-snyk, line 13: Why are we importing it with a different name?

parker-snyk (approved):

PR #871, opened 2026-06-12
feat: cn-1395 add apk ownership resolution library

LSTu5:

**Stack 3/5** — `main` ← apk file lists ← layer symlinks ← **this PR** ← ownership fact ← prettyName. Replaces the ownership-model portion of #861.

- [x] Ready for review

#### What does this PR do?

Adds the ownership resolution library — three new files, nothing wired in yet:

- `lib/analyzer/applications/evidence-paths.ts` — collects an app scan result's filesystem evidence: `identity.targetFile`, `testedFiles` facts, jar fingerprint locations.
- `lib/analyzer/package-managers/path-canonicalization.ts` — resolves paths through the image symlink map (chains, cycles, relative targets) so `/bin/node` matches APK records at `/usr/bin/node` on usr-merged images.
- `lib/analyzer/package-managers/apk-ownership.ts` — indexes APK `Files`/`Directories` (exact-file map + directory trie) and resolves which package owns a set of evidence paths.

**Why the rules are what they are:**

- **All-or-nothing containment** ([thread](https://github.com/snyk/snyk-docker-plugin/pull/861#discussion_r3389057732)): per [Chainguard's scanner spec](https://github.com/chainguard-dev/vulnerability-scanner-support/blob/main/docs/scanning_implementation.md), evidence paths must be *wholly contained* in the owning package's declared paths, and unowned dependencies "should be included in the final scan result". The result drives downstream finding **suppression**, so a partially-owned app (e.g. one jar apk-managed, one COPY'd in by the user) must keep its findings — guessing would hide real vulnerabilities in user-added software. Skipped paths are debug-logged.
- **Explicit disagreement rules** ([thread](https://github.com/snyk/snyk-docker-plugin/pull/861#discussion_r3389071739)): when evidence paths point at different packages, an owner backed by an exact file match outranks owners inferred from a parent directory; among directory-only matches the deepest prefix wins; still-split evidence yields no owner. (This replaces an earlier score-based tiebreak with a magic exact-match bonus.)

`apk-ownership.ts` top to bottom — `buildApkPathIndex`, `resolveOwnerForEvidencePath`, then `resolveApkOwnership` and the selection helpers. Then `canonicalizePath` in `path-canonicalization.ts`.

`npx jest test/unit/apk-ownership.spec.ts test/unit/path-canonicalization.spec.ts test/unit/evidence-paths.spec.ts` — covers exact/directory resolution, symlink canonicalization, partial-match refusal, exact-over-directory preference, split exact matches, depth ties.

Inert until the next PR wires it into `buildResponse`. Only consumes types introduced lower in the stack (`Files`/`Directories`, `SymlinkMap`).

CN-1395

#### Screenshots

N/A

#### Additional questions

N/A

### DIFF
--- a/lib/analyzer/applications/evidence-paths.ts
+++ b/lib/analyzer/applications/evidence-paths.ts
@@ -1,0 +1,33 @@
+import { JarFingerprintsFact, TestedFilesFact } from "../../facts";
+import { AppDepsScanResultWithoutTarget } from "./types";
+
+/** Collect filesystem paths from an app scan result used to resolve APK ownership. */
+export function extractEvidencePaths(
+  scanResult: AppDepsScanResultWithoutTarget,
+): string[] {
+  const paths = new Set<string>();
+
+  if (scanResult.identity.targetFile) {
+    paths.add(scanResult.identity.targetFile);
+  }
+
+  for (const fact of scanResult.facts) {
+    if (fact.type === "testedFiles") {
+      const testedFilesFact = fact as TestedFilesFact;
+      for (const filePath of testedFilesFact.data) {
+        paths.add(filePath);
+      }
+    }
+
+    if (fact.type === "jarFingerprints") {
+      const jarFact = fact as JarFingerprintsFact;
+      for (const fingerprint of jarFact.data.fingerprints) {
+        if (fingerprint.location) {
+          paths.add(fingerprint.location);
+        }
+      }
+    }
+  }
+
+  return [...paths];
+}
--- a/lib/analyzer/package-managers/apk-ownership.ts
+++ b/lib/analyzer/package-managers/apk-ownership.ts
@@ -1,0 +1,263 @@
+import * as Debug from "debug";
+import { SymlinkMap } from "../../extractor/types";
+import {
+  AnalysisType,
+  AnalyzedPackageWithVersion,
+  ImageAnalysis,
+  ImagePackagesAnalysis,
+  OSRelease,
+} from "../types";
+import {
+  canonicalizePath,
+  normalizeAbsolutePath,
+  SymlinkGraph,
+} from "./path-canonicalization";
+
+export interface ApkPackageOwnership {
+  distroId: string;
+  packageName: string;
+  packageVersion: string;
+  originPackage: string;
+  evidencePaths: string[];
+}
+
+export type MatchKind = "exact" | "directory";
+
+export interface PathOwnerMatch {
+  owner: AnalyzedPackageWithVersion;
+  matchKind: MatchKind;
+  /** Number of matched path segments; a deeper directory match outranks a shallower one. */
+  prefixLength: number;
+}
+
+interface DirectoryTrieNode {
+  owners: AnalyzedPackageWithVersion[];
+  children: Map<string, DirectoryTrieNode>;
+}
+
+export interface ApkPathIndex {
+  exactFileOwners: Map<string, AnalyzedPackageWithVersion[]>;
+  directoryTrie: DirectoryTrieNode;
+}
+
+const debug = Debug("snyk");
+
+const CHAINGUARD_DISTROS = new Set(["wolfi", "chainguard"]);
+
+export function isChainguardDistro(osRelease: OSRelease): boolean {
+  return CHAINGUARD_DISTROS.has(osRelease.name);
+}
+
+export function toSymlinkGraph(symlinks?: SymlinkMap): SymlinkGraph {
+  const graph: SymlinkGraph = new Map();
+  if (!symlinks) {
+    return graph;
+  }
+  for (const [symlinkPath, target] of Object.entries(symlinks)) {
+    graph.set(normalizeAbsolutePath(symlinkPath), target);
+  }
+  return graph;
+}
+
+export function buildApkPathIndex(
+  packages: AnalyzedPackageWithVersion[],
+  symlinkGraph: SymlinkGraph,
+): ApkPathIndex {
+  const exactFileOwners = new Map<string, AnalyzedPackageWithVersion[]>();
+  const directoryTrie: DirectoryTrieNode = { owners: [], children: new Map() };
+
+  for (const pkg of packages) {
+    for (const filePath of pkg.Files ?? []) {
+      const canonical = canonicalizePath(filePath, symlinkGraph);
+      addToOwnerMap(exactFileOwners, canonical, pkg);
+    }
+
+    for (const dirPath of pkg.Directories ?? []) {
+      const canonical = canonicalizePath(dirPath, symlinkGraph);
+      insertDirectoryOwner(directoryTrie, canonical, pkg);
+    }
+  }
+
+  return { exactFileOwners, directoryTrie };
+}
+
+export function resolveOwnerForEvidencePath(
+  evidencePath: string,
+  index: ApkPathIndex,
+  symlinkGraph: SymlinkGraph,
+): PathOwnerMatch | undefined {
+  const canonical = canonicalizePath(evidencePath, symlinkGraph);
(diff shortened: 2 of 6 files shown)


lib/analyzer/applications/evidence-paths.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  if (scanResult.identity.targetFile) {
+    paths.add(scanResult.identity.targetFile);
+  }
+
+  for (const fact of scanResult.facts) {
+    if (fact.type === "testedFiles") {
+      const testedFilesFact = fact as TestedFilesFact;
+      for (const filePath of testedFilesFact.data) {
+        paths.add(filePath);
```
parker-snyk, line 18: `testedFiles` entries are not always absolute image paths. Several analyzers emit **basenames**, e.g. `lib/analyzer/applications/php.ts:59` and `lib/analyzer/applications/node.ts:318` both do `data: [path.basename(pathPair.manifest), path.basename(pathPair.lock)]` → `["composer.json", "composer.lock"]` / `["package.json", "package-lock.json"]`.

Downstream in `resolveApkOwnership`, each evidence path goes through `normalizeAbsolutePath`, which roots a relative path at `/` → `/composer.json`. No APK package declares `/composer.json`, so `resolveOwnerForEvidencePath` returns `undefined`, and because of the all-or-nothing containment rule the **entire** app result is treated as unowned — even when the real files are genuinely owned by an APK package.

Net effect: ownership resolution can never succeed for PHP apps (and the node path at line 318), because `identity.targetFile` is absolute while these `testedFiles` are bare basenames, guaranteeing at least one `/`-rooted path that matches nothing. It fails *closed* (no incorrect suppression), but it silently defeats the feature for those ecosystems.

Suggestion: resolve relative evidence paths against the app directory (e.g. `path.posix.dirname(scanResult.identity.targetFile)`) before adding them, or skip non-absolute entries rather than rooting them at `/`. Either way, worth a unit test with basename-only `testedFiles`. -- AI Assisted Review

lib/analyzer/package-managers/path-canonicalization.ts
```diff
@@ -0,0 +1,76 @@
...
+      continue;
+    }
+
+    const resolvedTarget = resolveSymlinkChain(
+      normalizeSymlinkTarget(currentPath, linkTarget),
+      symlinkGraph,
+    );
+    const targetSegments = resolvedTarget.split("/").filter(Boolean);
+    resolvedSegments.length = 0;
+    resolvedSegments.push(...targetSegments);
```
parker-snyk, line 41: `canonicalizePath` resolves a symlink only when the *entire* accumulated prefix is a graph key, and after a hit it resets `resolvedSegments` to the target's segments and continues with the remaining **original** segments. It never re-examines the prefixes of the resolved target, and `resolveSymlinkChain` likewise only looks up the full string. So a symlink whose target sits under *another* symlinked directory is not fully resolved.

Example: `{ "/lib64": "/lib/x", "/lib": "/usr/lib" }`, canonicalizing `/lib64/foo`:
- `/lib64` → target `/lib/x`; `resolveSymlinkChain("/lib/x")` finds no key for the full `/lib/x` → returns it unchanged.
- reset to `["lib","x"]`, append `foo` → `/lib/x/foo`. The `/lib → /usr/lib` symlink is never applied.
- Result: `/lib/x/foo`, expected `/usr/lib/x/foo`.

The common single-level usrmerge case this feature targets (`/bin → /usr/bin`, etc.) works fine, so this is an edge-case correctness gap rather than a blocker — but when it hits, ownership silently fails to match the APK file list. Consider re-canonicalizing the resolved target segment-by-segment (feed the target back through the same per-prefix resolution), and add a test with a target whose parent is itself a symlink. -- AI Assisted Review

lib/analyzer/package-managers/apk-ownership.ts
```diff
@@ -0,0 +1,263 @@
...
+  evidencePaths: string[],
+  packages: AnalyzedPackageWithVersion[],
+  osRelease: OSRelease,
+  symlinkGraph: SymlinkGraph,
+): ApkPackageOwnership | undefined {
+  if (!isChainguardDistro(osRelease) || evidencePaths.length === 0) {
+    return undefined;
+  }
+
+  const index = buildApkPathIndex(packages, symlinkGraph);
```
parker-snyk, line 149: `resolveApkOwnership` rebuilds the full path index on every call — `buildApkPathIndex` iterates every APK package and canonicalizes every declared file and directory path (each canonicalization itself walks segments against the symlink graph). But the index depends only on `packages` + `symlinkGraph`, which are image-global and identical for every app scan result.

In PR #872 this function is called once per application scan result inside `buildResponse`'s `.map`, so the cost is O(appResults × apkPackages × declaredPaths) when it should be O(apkPackages × declaredPaths) once. On a Wolfi/Chainguard image with many packages and several app manifests this is a meaningful, fully-redundant chunk of work on the scan hot path.

Suggestion: split the index build from resolution — build the `ApkPathIndex` (and the `SymlinkGraph`) once per image and pass it into a `resolveApkOwnership(evidencePaths, index, symlinkGraph, osRelease)`. The per-result work then reduces to `extractEvidencePaths` + lookups. See the companion comment on the call site in #872. -- AI Assisted Review

lib/analyzer/package-managers/apk-ownership.ts
```diff
@@ -0,0 +1,263 @@
...
+  exactFileOwners: Map<string, AnalyzedPackageWithVersion[]>;
+  directoryTrie: DirectoryTrieNode;
+}
+
+const debug = Debug("snyk");
+
+const CHAINGUARD_DISTROS = new Set(["wolfi", "chainguard"]);
+
+export function isChainguardDistro(osRelease: OSRelease): boolean {
+  return CHAINGUARD_DISTROS.has(osRelease.name);
```
parker-snyk, line 48: Nit: this match is case-sensitive against a hardcoded lowercase set, and `tryOSRelease` (`lib/analyzer/os-release/release-analyzer.ts:11`) does **not** lowercase the `ID` field — unlike `tryLsbRelease`, which does. The os-release spec says `ID` is lowercase, so in practice `wolfi`/`chainguard` match fine; just calling it out as a latent robustness gap. A `osRelease.name.toLowerCase()` here would make it defensive against a non-conforming image without costing anything. -- AI Assisted Review

parker-snyk (commented):
Review of the ownership-resolution library. The structure is clean and the test coverage of the documented selection rules (exact-over-directory, deepest-prefix, split-evidence → none) is genuinely good. A few things worth addressing before this gets wired in:

**Worth fixing**
1. **Relative `testedFiles` defeat ownership** (`evidence-paths.ts`) — basename-only evidence paths get rooted at `/` and never match, silently disabling the feature for PHP and some node apps. Fails closed, but defeats the purpose.
2. **Symlinked-prefix-in-target not resolved** (`path-canonicalization.ts`) — multi-level symlink layouts canonicalize incorrectly. The common single-level usrmerge case works, so edge-case, but worth a test + fix.
3. **Index rebuilt per call** (`apk-ownership.ts`) — `buildApkPathIndex` is image-global but built once per app result via #872. Hoist it.

**Minor / nits**
4. Order-dependent exact-owner tiebreak (nondeterministic fact).
5. Case-sensitive distro check.

Nice work on the all-or-nothing fail-closed design and the debug logging on skipped paths — that's exactly the right instinct given this drives suppression. -- AI Assisted Review

lib/analyzer/package-managers/apk-ownership.ts
LSTu5: done in 872

parker-snyk (approved):

PR #872, opened 2026-06-12
feat: cn-1395 emit apkPackageOwnership fact for chainguard images

LSTu5:

**Stack 4/5** — `main` ← apk file lists ← layer symlinks ← ownership library ← **this PR** ← prettyName.

- [x] Ready for review

#### What does this PR do?

Wires the ownership library into `buildResponse`: every application ScanResult on a Chainguard/Wolfi image whose evidence paths are wholly owned by one APK package gains an `apkPackageOwnership` fact (distro, package name, installed version, **origin**). The origin is the key into Chainguard's advisory feed — it lets downstream matching reconcile app findings against the owning package's fixed-version data instead of reporting raw upstream-range matches.

⚠️ **Public API note:** `apkPackageOwnership` is a new `FactType` — a contract change visible to all plugin consumers (CLI, kubernetes-monitor, registry agents). The fact is only *emitted* for Chainguard distros with fully resolved evidence, so existing scan output is unchanged — zero snapshot churn in this PR (verified: no `.snap` in the repo contains the fact).

Also removes the unreachable `?? depsAnalysis.depTree.targetOS` fallback on the `osRelease` read — `StaticAnalysis.osRelease` is a required field, so the fallback could never fire.

The single block added to `lib/response-builder.ts`, then the fact shape in `lib/facts.ts`. The heavy algorithmic review happened one PR down.

`npx jest test/unit/response-builder-ownership.spec.ts` (full `buildResponse` pass on a synthetic Wolfi analysis asserting the fact and its data) and `npx jest test/system/application-scans/chainguard-ownership.spec.ts` (pulls `chainguard/bash`, asserts an OS-only Chainguard image emits **no** ownership facts — the fail-closed direction). `test/system/package-managers/apk.spec.ts` snapshots stay byte-identical.

Downstream consumption is follow-up work in `registry` (thread the fact through) and `@snyk/vuln` (subtract findings when installed ≥ fixed for the owning origin). Until those land, the fact rides along in scan results unconsumed.

CN-1395

#### Screenshots

N/A

#### Additional questions

N/A

### DIFF
--- a/lib/response-builder.ts
+++ b/lib/response-builder.ts
@@ -1,4 +1,11 @@
 import { legacy } from "@snyk/dep-graph";
+import { extractEvidencePaths } from "./analyzer/applications/evidence-paths";
+import {
+  buildApkPathIndex,
+  getApkPackagesFromResults,
+  resolveApkOwnership,
+  toSymlinkGraph,
+} from "./analyzer/package-managers/apk-ownership";
 import { IntroducingLayerByPackage, StaticAnalysis } from "./analyzer/types";
 import * as facts from "./facts";
 // Module that provides functions to collect and build response after all
@@ -240,6 +247,16 @@ async function buildResponse(
     additionalFacts.push(autoDetectedUserInstructionsFact);
   }
 
+  // The APK ownership inputs are image-global, so build the path index once
+  // here rather than rebuilding it for every application scan result below.
+  // resolveApkOwnership decides which distros the index actually applies to.
+  const osRelease = depsAnalysis.osRelease;
+  const apkSymlinkGraph = toSymlinkGraph(depsAnalysis.symlinks);
+  const apkPathIndex = buildApkPathIndex(
+    getApkPackagesFromResults(depsAnalysis.results),
+    apkSymlinkGraph,
+  );
+
   const applicationDependenciesScanResults: types.ScanResult[] = (
     depsAnalysis.applicationDependenciesScanResults || []
   ).map((appDepsScanResult) => {
@@ -291,6 +308,20 @@ async function buildResponse(
     //     appDepsScanResult.facts.push(historyFact);
     //   }
     // }
+
+    const ownership = resolveApkOwnership(
+      extractEvidencePaths(appDepsScanResult),
+      apkPathIndex,
+      apkSymlinkGraph,
+      osRelease,
+    );
+    if (ownership) {
+      const ownershipFact: facts.ApkPackageOwnershipFact = {
+        type: "apkPackageOwnership",
+        data: ownership,
+      };
+      appDepsScanResult.facts.push(ownershipFact);
+    }
 
     return {
       ...appDepsScanResult,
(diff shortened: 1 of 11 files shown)


lib/response-builder.ts
```diff
@@ -261,6 +267,25 @@ async function buildResponse(
     };
     appDepsScanResult.facts.push(appPluginVersionFact);
 
+    const osRelease = depsAnalysis.osRelease;
+    const evidencePaths = extractEvidencePaths(appDepsScanResult);
+    const apkPackages = getApkPackagesFromResults(depsAnalysis.results);
```
parker-snyk: This block sits inside the `applicationDependenciesScanResults.map(...)`, so it runs once per app scan result — but three of the four inputs are image-global and identical on every iteration:

- `getApkPackagesFromResults(depsAnalysis.results)` — re-scans `results` each time
- `toSymlinkGraph(depsAnalysis.symlinks)` — rebuilds the same `Map` each time
- `resolveApkOwnership(...)` → `buildApkPathIndex(...)` — rebuilds the entire exact-file map + directory trie, canonicalizing every APK declared path, each time

Only `extractEvidencePaths(appDepsScanResult)` is genuinely per-result. For an image with N app results this is N redundant index builds; combined with the per-call rebuild flagged in #871 it's O(N × packages × paths).

Suggestion: hoist the apk packages, symlink graph, and the path index above the `.map`, and pass the prebuilt index into resolution (see the companion comment in #871). The hot-loop body then becomes just `extractEvidencePaths` + lookups. -- AI Assisted Review

lib/response-builder.ts
```diff
@@ -261,6 +267,25 @@ async function buildResponse(
...
+    const ownership =
+      osRelease &&
```
parker-snyk: Nit: `StaticAnalysis.osRelease` is a required (non-optional) field, so the `osRelease &&` guard is dead from the type system's perspective, and it makes `ownership` a `false | ApkPackageOwnership | undefined` union that only works because of the truthy `if (ownership)` below. `resolveApkOwnership` already takes `osRelease` and short-circuits on non-Chainguard distros, so this guard is doing nothing useful. Either drop it, or — if `osRelease` really can be absent at runtime (the PR notes you removed a `?? depTree.targetOS` fallback for exactly this reason) — fix the type to `osRelease?` so the guard is honest. -- AI Assisted Review

parker-snyk (commented):
Wiring looks correct and the fail-closed direction is well tested (the `chainguard/bash` system test asserting OS-only images emit *no* ownership fact is a great choice). Two items:

1. **Per-app-result recomputation** — the apk packages, symlink graph, and path index are image-global but rebuilt inside the `.map`. Hoist them out (paired with the index/resolution split suggested in #871).
2. **Redundant `osRelease &&` guard** — dead given the required type; either drop it or make the type optional.

Credit where due: explicitly calling out the new `apkPackageOwnership` `FactType` as a public-API contract change in the PR description — with the note that it's only *emitted* for Chainguard so existing output is unchanged — is exactly what AGENTS.md asks for. Nicely handled. -- AI Assisted Review

parker-snyk (approved):

PR #873, opened 2026-06-12
feat: cn-1395 emit imageOsReleasePrettyName on app scan results

LSTu5:

**Stack 5/5 (top)** — `main` ← apk file lists ← layer symlinks ← ownership library ← ownership fact ← **this PR**. Independently droppable: nothing below depends on it.

- [x] Ready for review

#### What does this PR do?

Emits the existing `imageOsReleasePrettyName` fact on every application-dependency ScanResult (previously only the OS ScanResult carried it).

**Why:** app-dep scan results carry no distro context at all today — that's half the root cause of the Chainguard false-positive escalation (`registry` strips distro context from app dep graphs; matching gets zero Chainguard signal). This gives the pipeline a routing signal to associate app results with their distro. Note it intentionally fires for **all** distros, not just Chainguard — flagging this explicitly since it's the one behavior change in the stack that affects every scan.

The one block in `lib/response-builder.ts`. Everything else is snapshot churn: the fact appearing once per app scan result across the system suites.

`npx jest test/system/application-scans/node.spec.ts test/system/application-scans/php.spec.ts test/system/flags/app-vulns.spec.ts test/system/operating-systems/sles15.spec.ts test/system/operating-systems/ubi8.spec.ts test/system/application-scans/gomodules.spec.ts` — all snapshot deltas are exactly one `imageOsReleasePrettyName` object per app result. (registry-auth and windows snapshots verified by CI: the former needs private-image creds, the latter the Windows runner.)

The ownership fact (one PR down) already carries `distroId`, so if this fact's all-distro scope is contested, this PR can be dropped or gated without touching the rest of the stack.

CN-1395

#### Screenshots

N/A

#### Additional questions

N/A

### DIFF
--- a/lib/response-builder.ts
+++ b/lib/response-builder.ts
@@ -278,14 +278,6 @@ async function buildResponse(
     };
     appDepsScanResult.facts.push(appPluginVersionFact);
 
-    if (osRelease?.prettyName) {
-      const imageOsReleasePrettyNameFact: facts.ImageOsReleasePrettyNameFact = {
-        type: "imageOsReleasePrettyName",
-        data: osRelease.prettyName,
-      };
-      appDepsScanResult.facts.push(imageOsReleasePrettyNameFact);
-    }
-
     const ownership = resolveApkOwnership(
       extractEvidencePaths(appDepsScanResult),
       apkPathIndex,
--- a/lib/response-builder.ts
+++ b/lib/response-builder.ts
@@ -278,6 +278,14 @@ async function buildResponse(
     };
     appDepsScanResult.facts.push(appPluginVersionFact);
 
+    if (osRelease?.prettyName) {
+      const imageOsReleasePrettyNameFact: facts.ImageOsReleasePrettyNameFact = {
+        type: "imageOsReleasePrettyName",
+        data: osRelease.prettyName,
+      };
+      appDepsScanResult.facts.push(imageOsReleasePrettyNameFact);
+    }
+
     const ownership = resolveApkOwnership(
       extractEvidencePaths(appDepsScanResult),
       apkPathIndex,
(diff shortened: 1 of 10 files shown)


lib/response-builder.ts
```diff
@@ -268,6 +268,14 @@ async function buildResponse(
     appDepsScanResult.facts.push(appPluginVersionFact);
 
     const osRelease = depsAnalysis.osRelease;
+    if (osRelease?.prettyName) {
```
parker-snyk: Looks correct. One thing to be aware of: `osRelease?.prettyName` skips the fact whenever `prettyName` is empty, and several distros set `prettyName: ""` in this codebase (e.g. `scratch`, `chisel`, and every `release-analyzer` branch that doesn't parse `PRETTY_NAME` — debian, alpine via `/etc/alpine-release`, etc.). So "fires for all distros" really means "all distros whose os-release carries a non-empty `PRETTY_NAME`." That's almost certainly the behavior you want (no point emitting an empty string), just flagging that the routing signal won't be present on those images — if downstream relies on it as a distro discriminator, it'll be absent exactly where `PRETTY_NAME` is missing. No change needed unless that matters to the consumer. -- AI Assisted Review

parker-snyk (commented):
Clean, focused change. The non-snapshot diff is a single well-guarded block, and the snapshot churn is exactly what's expected — one `imageOsReleasePrettyName` fact per app scan result, placed consistently after `pluginVersion`, with distro-correct values across the node/php/gomodules/app-vulns/sles15/ubi8 suites. Good call flagging the all-distro scope in the PR description and noting it's independently droppable.

One non-blocking observation inline about the empty-`prettyName` case. Otherwise LGTM. -- AI Assisted Review

(Posting as a comment rather than a formal approval since I'm an automated reviewer — a human on @snyk/container_container should give the actual sign-off.) -- AI Assisted Review

PR #879, opened 2026-06-18
feat: cn-1395 resolve per-dependency apk ownership for npm global modules

LSTu5:

**Follow-up to #872.** Stacked on `feat/cn-1395-apk-ownership-fact`.

#### What does this PR do?

#872 emits an `apkPackageOwnership` fact only when **all** of an application-dependency ScanResult's evidence paths are owned by a single apk package. That works for Java (jar paths) and Go (binaries), but **npm global modules fail closed**: their only evidence path is the shared `/usr/lib/node_modules` root, which is declared by multiple apk packages (`npm`, `node-gyp`), so nothing resolves — even though individual bundled packages **are** owned.

Concretely, `brace-expansion` (a CN-1395 example finding) lives at `/usr/lib/node_modules/npm/node_modules/brace-expansion` and is owned by the `npm` apk package, but #872 never reconciles it.

This PR unifies ownership resolution across all three ecosystems behind a single **ownership-candidate** model, resolving npm per-package while preserving whole-result resolution for Java and Go:

- **`node.ts`** — `collectNodeModulesPackagePaths()` parses each `package.json` under `node_modules/` to derive `{name, version, installDir}` per installed package, attached as an internal `nodeModulesPackagePaths` field. Gated by `isChainguardDistro()`, so non-Chainguard images skip the parsing entirely.
- **`evidence-paths.ts`** — `buildOwnershipCandidates()` produces one `OwnershipCandidate` per npm package, keyed by that package's own install dir and carrying a `name@version` coordinate. Java/Go fall through to the existing path: a single whole-result candidate over all evidence paths, with no coordinate. `.pnpm`, `.bin`, and `.cache` entries are skipped.
- **`apk-ownership.ts`** — `resolveApkOwnership()` now takes `OwnershipCandidate[]` and resolves each candidate independently (fail closed per candidate, so user-installed packages keep their findings). Candidates with coordinates are deduped by `name@version` — keeping an owned occurrence over an unowned one — sorted by first evidence path for deterministic output, with memoized per-path lookups.
- **`response-builder.ts`** — emits a **single** `apkPackageOwnership` fact and strips the internal `nodeModulesPackagePaths` field before the result becomes a public `ScanResult`, so it never reaches the wire.

#### ⚠️ Public API note

The `apkPackageOwnership` `FactType` string is **unchanged**, but its data shape changes from flat `packageName`/`packageVersion`/`originPackage` fields to:

```ts
{ distroId, ownedPackages: OwnedPackage[] }
// OwnedPackage: { evidencePaths, apkPackageName, apkPackageVersion, originPackage, name?, version? }
```

`name`/`version` are present only for per-dependency entries (npm); they're absent for whole-result entries (Go, Java). The fact only emits on Chainguard/Wolfi images → **zero snapshot churn**.

`buildOwnershipCandidates` in `lib/analyzer/applications/evidence-paths.ts`, then `resolveApkOwnership` in `lib/analyzer/package-managers/apk-ownership.ts`, then the wiring in `lib/response-builder.ts`.

#### How was this tested?

- Unit: `npx jest test/unit/apk-ownership.spec.ts test/unit/evidence-paths.spec.ts test/unit/node-package-paths.spec.ts test/unit/response-builder-ownership.spec.ts`
- System (Docker + cgr.dev): `npx jest test/system/application-scans/chainguard-ownership.spec.ts` — scans `cgr.dev/chainguard/node:latest` and asserts `brace-expansion` → origin `npm`. **Verified passing.**
- Full `npm run test:unit` + `npm run build` + `npm run lint` all green, no snapshot churn.

#### Out of scope (follow-ups)

- pnpm virtual store (`.pnpm`) — skipped.
- Java multi-jar mixed-owner results — separate concern.

#### Relevant tickets
CN-1395

### DIFF
--- a/lib/analyzer/applications/node.ts
+++ b/lib/analyzer/applications/node.ts
@@ -23,6 +23,7 @@ import {
   groupNodeModulesFilesByDirectory,
   persistNodeModules,
 } from "./node-modules-utils";
+import { NodeModulesPackagePath } from "../types";
 import {
   AppDepsScanResultWithoutTarget,
   FilePathToContent,
@@ -38,6 +39,10 @@ interface ManifestLockPathPair {
 export async function nodeFilesToScannedProjects(
   filePathToContent: FilePathToContent,
   shouldIncludeNodeModules: boolean,
+  // Per-package install paths are only consumed for APK ownership on
+  // Chainguard/Wolfi images, so collection is gated to avoid parsing every
+  // node_modules package.json on images that will never use it.
+  collectOwnershipPaths = false,
 ): Promise<AppDepsScanResultWithoutTarget[]> {
   const scanResults: AppDepsScanResultWithoutTarget[] = [];
   /**
@@ -81,6 +86,7 @@ export async function nodeFilesToScannedProjects(
           filePathToContent,
           nodeProjects,
           appNodeModulesGroupedByDirectory,
+          collectOwnershipPaths,
         )),
       );
     }
@@ -89,10 +95,68 @@ export async function nodeFilesToScannedProjects(
   return scanResults;
 }
 
+const nodeModulesPackageJsonRegex = /\/node_modules\/.+\/package\.json$/;
+
+/**
+ * Collect the on-disk install directory of each package under a project's
+ * node_modules, from the image paths already gathered for the scan. Used to
+ * resolve per-package APK ownership downstream (response-builder), since the
+ * scan result itself only records the shared node_modules root. Skips pnpm/.bin
+ * virtual entries and any package.json without a name+version.
+ */
+export function collectNodeModulesPackagePaths(
+  project: string,
+  fileNamesGroupedByDirectory: FilesByDirMap,
+  filePathToContent: FilePathToContent,
+): NodeModulesPackagePath[] {
+  const files = fileNamesGroupedByDirectory.get(project);
+  if (!files) {
+    return [];
+  }
+
+  const packages: NodeModulesPackagePath[] = [];
+  for (const filePath of files) {
+    // Skip dot-prefixed node_modules entries (.pnpm virtual store, .bin, .cache).
+    // Safe because npm package names cannot begin with ".", so a real package
+    // dir is never excluded by this guard.
+    if (
+      !nodeModulesPackageJsonRegex.test(filePath) ||
+      filePath.includes("/node_modules/.")
+    ) {
+      continue;
+    }
+
+    const content = filePathToContent[filePath];
+    if (!content) {
+      continue;
+    }
+
+    let parsed: { name?: string; version?: string };
+    try {
+      parsed = JSON.parse(content);
+    } catch {
+      continue;
+    }
+
+    if (!parsed.name || !parsed.version) {
+      continue;
+    }
+
+    packages.push({
+      name: parsed.name,
+      version: parsed.version,
+      installDir: path.posix.dirname(filePath),
+    });
+  }
+
+  return packages;
+}
+
 async function depGraphFromNodeModules(
   filePathToContent: FilePathToContent,
   nodeProjects: string[],
   fileNamesGroupedByDirectory: FilesByDirMap,
+  collectOwnershipPaths: boolean,
 ): Promise<AppDepsScanResultWithoutTarget[]> {
   const scanResults: AppDepsScanResultWithoutTarget[] = [];
   for (const project of nodeProjects) {
@@ -141,6 +205,16 @@ async function depGraphFromNodeModules(
         pkgTree.type || "npm",
       );
 
+      // Only collected on Chainguard/Wolfi (the caller passes the distro gate),
+      // since per-package ownership is the only consumer.
+      const nodeModulesPackages = collectOwnershipPaths
+        ? collectNodeModulesPackagePaths(
+            project,
+            fileNamesGroupedByDirectory,
+            filePathToContent,
+          )
+        : [];
+
       scanResults.push({
         facts: [
           {
@@ -160,6 +234,9 @@ async function depGraphFromNodeModules(
             ? manifestPath
             : path.join(project, "node_modules"),
         },
+        ...(nodeModulesPackages.length > 0
+          ? { nodeModulesPackagePaths: nodeModulesPackages }
+          : {}),
       });
     } catch (error) {
       debug(
--- a/lib/analyzer/applications/node.ts
+++ b/lib/analyzer/applications/node.ts
@@ -23,6 +23,7 @@ import {
   groupNodeModulesFilesByDirectory,
   persistNodeModules,
 } from "./node-modules-utils";
+import { normalizeAbsolutePath } from "../package-managers/path-canonicalization";
 import { NodeModulesPackagePath } from "../types";
 import {
   AppDepsScanResultWithoutTarget,
@@ -95,7 +96,12 @@ export async function nodeFilesToScannedProjects(
   return scanResults;
 }
 
-const nodeModulesPackageJsonRegex = /\/node_modules\/.+\/package\.json$/;
+// Match a package.json only at a real package root (parent is an immediate
+// child of node_modules, optional @scope/), so bundled fixtures nested deeper
+// aren't collected as packages. [\\/] handles both POSIX and Windows paths.
+const nodeModulesPackageJsonRegex =
+  /[\\/]node_modules[\\/](?:@[^\\/]+[\\/])?[^\\/]+[\\/]package\.json$/;
+const nodeModulesDotDirRegex = /[\\/]node_modules[\\/]\./;
 
 /**
  * Collect the on-disk install directory of each package under a project's
@@ -116,12 +122,11 @@ export function collectNodeModulesPackagePaths(
 
   const packages: NodeModulesPackagePath[] = [];
   for (const filePath of files) {
-    // Skip dot-prefixed node_modules entries (.pnpm virtual store, .bin, .cache).
-    // Safe because npm package names cannot begin with ".", so a real package
-    // dir is never excluded by this guard.
+    // Skip dot-prefixed node_modules entries (.pnpm, .bin, .cache); npm package
+    // names never start with ".", so no real package dir is excluded.
     if (
       !nodeModulesPackageJsonRegex.test(filePath) ||
-      filePath.includes("/node_modules/.")
+      nodeModulesDotDirRegex.test(filePath)
     ) {
       continue;
     }
@@ -131,21 +136,29 @@ export function collectNodeModulesPackagePaths(
       continue;
     }
 
-    let parsed: { name?: string; version?: string };
+    let parsed: { name?: unknown; version?: unknown };
     try {
       parsed = JSON.parse(content);
     } catch {
       continue;
     }
 
-    if (!parsed.name || !parsed.version) {
+    // JSON.parse returns any; require non-empty strings since the coordinate
+    // flows into the apkPackageOwnership fact as name/version.
+    if (
+      typeof parsed.name !== "string" ||
+      typeof parsed.version !== "string" ||
+      !parsed.name ||
+      !parsed.version
+    ) {
       continue;
     }
 
     packages.push({
       name: parsed.name,
       version: parsed.version,
-      installDir: path.posix.dirname(filePath),
+      // Normalize first so posix.dirname works on Windows backslash paths.
+      installDir: path.posix.dirname(normalizeAbsolutePath(filePath)),
     });
   }
 
--- a/lib/analyzer/package-managers/apk-ownership.ts
+++ b/lib/analyzer/package-managers/apk-ownership.ts
@@ -1,4 +1,3 @@
-import * as Debug from "debug";
 import { SymlinkMap } from "../../extractor/types";
 import {
   AnalysisType,
@@ -13,12 +12,35 @@ import {
   SymlinkGraph,
 } from "./path-canonicalization";
 
+export interface OwnedPackage {
+  // The file(s)/dir(s) that justify this ownership. Always present — the
+  // universal key across ecosystems (jar files, npm install dirs, Go binaries).
+  evidencePaths: string[];
+  // The apk package that owns the evidence paths.
+  apkPackageName: string;
+  apkPackageVersion: string;
+  // apk Source package (falls back to apkPackageName).
+  originPackage: string;
+  // Dependency coordinate (name@version) for per-package ecosystems where only
+  // some of a result's dependencies are owned — npm global modules. Absent for
+  // whole-result entries (Go binaries, Java jar dirs), where the entire result
+  // is owned and downstream relabels every dependency it produced.
+  name?: string;
+  version?: string;
+}
+
 export interface ApkPackageOwnership {
   distroId: string;
-  packageName: string;
-  packageVersion: string;
-  originPackage: string;
+  ownedPackages: OwnedPackage[];
+}
+
+// A unit of evidence to resolve ownership for. Each candidate is resolved
+// independently (fail closed per candidate): all of its evidence paths must
+// resolve to one consistent apk package, or it is omitted.
+export interface OwnershipCandidate {
   evidencePaths: string[];
+  name?: string;
+  version?: string;
 }
 
 export type MatchKind = "exact" | "directory";
@@ -39,8 +61,6 @@ export interface ApkPathIndex {
   exactFileOwners: Map<string, AnalyzedPackageWithVersion[]>;
   directoryTrie: DirectoryTrieNode;
 }
-
-const debug = Debug("snyk");
 
 const CHAINGUARD_DISTROS = new Set(["wolfi", "chainguard"]);
 
@@ -144,52 +164,104 @@ function uniqueDeclaredOwner(
 }
 
 /**
- * Resolve APK package ownership for app evidence paths on Wolfi/Chainguard images.
- *
- * Per Chainguard's scanner spec, an app dependency is owned by an APK package
- * only when its evidence paths are wholly contained in that package's declared
- * paths; a dependency with any unowned path is not covered by Chainguard's
- * advisory data and must keep its findings. This fact drives downstream
- * suppression, so we skip it rather than guess and risk suppressing real
- * vulnerabilities in user-added software.
+ * Resolve APK package ownership for a set of candidates on Wolfi/Chainguard
+ * images. Each candidate (a single npm package, or a whole Go/Java result) is
+ * resolved independently: all of its evidence paths must be wholly owned by one
+ * consistent APK package, or the candidate is omitted (fail closed per
+ * candidate). Per Chainguard's scanner spec, a dependency with any unowned path
+ * is not covered by advisory data and must keep its findings — we skip rather
+ * than guess and risk suppressing real vulnerabilities in user-added software.
+ * Candidates carrying a coordinate (name@version) are deduplicated, keeping an
+ * owned occurrence over an unowned one.
  * https://github.com/chainguard-dev/vulnerability-scanner-support/blob/main/docs/scanning_implementation.md
  */
 export function resolveApkOwnership(
-  evidencePaths: string[],
+  candidates: OwnershipCandidate[],
   index: ApkPathIndex,
   symlinkGraph: SymlinkGraph,
   osRelease: OSRelease,
 ): ApkPackageOwnership | undefined {
-  if (!isChainguardDistro(osRelease) || evidencePaths.length === 0) {
+  if (!isChainguardDistro(osRelease) || candidates.length === 0) {
     return undefined;
   }
 
-  const perPathMatches: PathOwnerMatch[] = [];
+  const ownedPackages: OwnedPackage[] = [];
+  // `seen` holds matched coordinates only, so an owned occurrence always wins
+  // over an unowned one. It must NOT include unmatched coordinates, or a later
+  // owned copy of the same coordinate would be skipped.
+  const seen = new Set<string>();
+  // Memoize per-path owner lookups so a path shared by several candidates (or an
+  // unmatched coordinate seen repeatedly) is resolved only once.
+  const resolvedByPath = new Map<
+    string,
+    ReturnType<typeof resolveOwnerForEvidencePath>
+  >();
+  // Sort by first evidence path so coordinate dedup is deterministic regardless
+  // of discovery order: when a coordinate is bundled under more than one apk
+  // package, the lexicographically-first evidence path wins. Which origin is
+  // "correct" in that case is a downstream (@snyk/vuln) question.
+  const ordered = [...candidates].sort((a, b) =>
+    (a.evidencePaths[0] ?? "").localeCompare(b.evidencePaths[0] ?? ""),
+  );
 
-  for (const evidencePath of evidencePaths) {
-    const normalized = normalizeAbsolutePath(evidencePath);
-    const match = resolveOwnerForEvidencePath(normalized, index, symlinkGraph);
-    if (!match) {
-      debug(
-        `apk ownership skipped: no owning package for evidence path ${normalized}`,
-      );
-      return undefined;
+  for (const candidate of ordered) {
+    if (candidate.evidencePaths.length === 0) {
+      continue;
     }
-    perPathMatches.push(match);
+    const coordinate =
+      candidate.name && candidate.version
+        ? `${candidate.name}@${candidate.version}`
+        : undefined;
+    if (coordinate && seen.has(coordinate)) {
+      continue;
+    }
+
+    // Fail closed per candidate: every evidence path must resolve to one
+    // consistent owner, else the candidate keeps its findings.
+    const matches: PathOwnerMatch[] = [];
+    let allOwned = true;
+    for (const evidencePath of candidate.evidencePaths) {
+      const normalized = normalizeAbsolutePath(evidencePath);
+      if (!resolvedByPath.has(normalized)) {
+        resolvedByPath.set(
+          normalized,
+          resolveOwnerForEvidencePath(normalized, index, symlinkGraph),
+        );
+      }
+      const match = resolvedByPath.get(normalized);
+      if (!match) {
+        allOwned = false;
+        break;
+      }
+      matches.push(match);
+    }
+    if (!allOwned) {
+      continue;
+    }
+
+    const owner = pickConsistentOwner(matches);
+    if (!owner) {
+      continue;
+    }
+
+    if (coordinate) {
+      seen.add(coordinate);
+    }
+    ownedPackages.push({
+      evidencePaths: candidate.evidencePaths,
+      apkPackageName: owner.Name,
+      apkPackageVersion: owner.Version,
+      originPackage: owner.Source ?? owner.Name,
+      ...(candidate.name ? { name: candidate.name } : {}),
+      ...(candidate.version ? { version: candidate.version } : {}),
+    });
   }
 
-  const owner = pickConsistentOwner(perPathMatches);
-  if (!owner) {
+  if (ownedPackages.length === 0) {
     return undefined;
   }
 
-  return {
-    distroId: osRelease.name,
-    packageName: owner.Name,
-    packageVersion: owner.Version,
-    originPackage: owner.Source ?? owner.Name,
-    evidencePaths,
-  };
+  return { distroId: osRelease.name, ownedPackages };
 }
 
 function ownerKey(pkg: AnalyzedPackageWithVersion): string {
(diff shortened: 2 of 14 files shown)


lib/analyzer/applications/node.ts
```diff
@@ -141,6 +201,12 @@ async function depGraphFromNodeModules(
         pkgTree.type || "npm",
       );
 
+      const nodeModulesPackages = collectNodeModulesPackagePaths(
```
LSTu5: **[perf] This extraction runs on every node image, not just Chainguard.** `collectNodeModulesPackagePaths` `JSON.parse`s every `package.json` under `node_modules` and attaches the carrier fact for *all* images scanned via the global-modules path — yet it's only ever consumed on Chainguard/Wolfi (the resolver gates on `isChainguardDistro`, and response-builder strips the carrier otherwise). On a large node image (thousands of deps) that's thousands of parses + a sizeable transient fact for zero benefit on the common case.

The scanner can't see `osRelease` here (that absence is the original CN-1395 root cause), so we can't cheaply gate at the source. Not a blocker, but let's make the call explicit: measure the overhead on a big node image and accept it, or defer collection. Right now the cost is silent.

lib/analyzer/applications/node.ts
```diff
@@ -89,6 +89,66 @@ export async function nodeFilesToScannedProjects(
   return scanResults;
 }
 
+export interface NodeModulesPackagePath {
+  name: string;
+  version: string;
+  installDir: string;
+}
+
+const nodeModulesPackageJsonRegex = /\/node_modules\/.+\/package\.json$/;
```
LSTu5: **[edge case + test gap] Scoped packages.** This regex matches `.../node_modules/@scope/name/package.json` and `path.posix.dirname` gives the right `installDir`, with the coordinate taken from the `package.json` `name` (`@scope/name`) — so it should work. But there's no test for it, and it's precisely open question #1 (join-key form for `@scope/name`). Please add a unit case.

Separately, the `.includes('/node_modules/.')` guard on line 121 is a broad heuristic (drops `.pnpm`, `.bin`, `.cache`, ...). It's safe because npm package names can't begin with `.`, but a one-line comment stating that assumption would save the next reader a double-take.

LSTu5 (commented):
Senior review of the npm per-dependency ownership change.

Overall: clean, well-scoped, and genuinely additive — new `FactType` + a transient carrier, zero snapshot churn, and the core resolver (`resolveApkOwnershipsPerDependency`) is ecosystem-agnostic, so a future Python/site-packages scanner can reuse it. The system test proves the real `brace-expansion` -> `npm` reconciliation end-to-end, which is exactly the CN-1395 pain.

Nothing here is a correctness blocker. Comments are one perf call-out, a couple of robustness/maintainability nits, and test-durability concerns. Before un-drafting I'd want: (1) the `:latest` / brace-expansion system test made deterministic, (2) an explicit decision on the per-package extraction cost that runs on every node image, and (3) the two downstream contract questions (coordinate join key + dedup policy) confirmed with @snyk/vuln — they directly inform the dedup-tiebreak comment.

lib/analyzer/package-managers/apk-ownership.ts
```diff
@@ -192,6 +206,55 @@ export function resolveApkOwnership(
...
+    const coordinate = `${pkg.name}@${pkg.version}`;
+    if (seen.has(coordinate)) {
+      continue;
+    }
+    const normalized = normalizeAbsolutePath(pkg.installDir);
+    const match = resolveOwnerForEvidencePath(normalized, index, symlinkGraph);
+    if (!match) {
+      continue;
+    }
+    seen.add(coordinate);
```
LSTu5: **[correctness, minor] Dedup tiebreak is iteration-order dependent.** Dedup keeps the *first* owned occurrence of a coordinate, where order derives from `filePathToContent`/`Set` iteration. If the same `name@version` is bundled by two *different* apk packages (different origins), the reported owner — and the single `evidencePath` — is whichever is hit first, non-deterministically, and the alternative is silently dropped.

Harmless for brace-expansion, but it intersects open question #2 (single path vs per-location). Either define an explicit tiebreak (e.g. prefer an exact match / shallowest path) or document that a coordinate intentionally maps to one arbitrary owning origin.

lib/facts.ts
```diff
@@ -173,3 +173,28 @@ export interface ApkPackageOwnershipFact {
     evidencePaths: string[];
   };
 }
+
+export interface ApkPackageOwnershipsFact {
```
LSTu5: **[maintainability] Duplicated shape.** This `data` block is hand-copied from `ApkPackageOwnerships` / `OwnedDependency` in `apk-ownership.ts` rather than imported, so it'll drift silently if the lib type changes (the existing `ApkPackageOwnershipFact` has the same smell). Prefer `data: ApkPackageOwnerships` (import the type) so the compiler keeps them in lockstep.

lib/response-builder.ts
```diff
@@ -292,6 +293,32 @@ async function buildResponse(
       appDepsScanResult.facts.push(ownershipFact);
     }
 
+    // npm global modules live in a shared node_modules root that no single APK
+    // package owns, so the whole-result resolution above fails closed. Resolve
+    // ownership per package instead, using the per-package install dirs the node
+    // scanner attached as a transient fact, then strip that transient fact.
+    const npmPackagePathsFact = appDepsScanResult.facts.find(
```
LSTu5: **[robustness] Three notes on this block:**
1. The carrier-strip (`.filter` on line 317) only runs inside `if (npmPackagePathsFact)`. Correct today, but stripping unconditionally would be more defensive if the fact ever arrives via another path.
2. Mixed mutation style: earlier facts are added via `.push` (in place) but here `.facts` is reassigned to a filtered copy. Works, just inconsistent — consider composing the final facts array once.
3. Implicit invariant: the whole-result `resolveApkOwnership` above runs on this same npm result and only fails closed because the shared `node_modules` root is multi-owned. If that ever changes, an npm result could emit *both* `apkPackageOwnership` and `apkPackageOwnerships`. Worth a guard (skip the per-dependency path when whole-result already fired, or vice-versa) or at least a comment naming the invariant.

test/system/application-scans/chainguard-ownership.spec.ts
```diff
@@ -1,25 +1,23 @@
 import { scan } from "../../../lib/index";
 
+const BASH_IMAGE =
+  "chainguard/bash@sha256:642933df66209814502599053ca3dfa97cccf847badc4219d2b1fd6565f6559a";
+const NODE_IMAGE = "cgr.dev/chainguard/node:latest";
```
LSTu5: **[test brittleness] Three things here:**
1. `:latest` is mutable -> non-reproducible across runs. The bash image above is digest-pinned for exactly this reason; consider `cgr.dev/chainguard/node@sha256:...`.
2. The assertion keys on `brace-expansion`, a *transitive* dep bundled inside the npm package that can change or vanish across npm releases — the test will silently rot. Prefer asserting the invariant instead: at least one `apkPackageOwnerships` entry with `originPackage === 'npm'`.
3. This pulls from `cgr.dev` while the existing test uses `docker.io`. Please confirm CI can pull `cgr.dev/chainguard/*` anonymously, otherwise this is green locally and red in CI.

test/unit/node-package-paths.spec.ts
```diff
@@ -0,0 +1,99 @@
+import { collectNodeModulesPackagePaths } from "../../lib/analyzer/applications/node";
+import {
+  FilePathToContent,
+  FilesByDirMap,
+} from "../../lib/analyzer/applications/types";
+
+describe("collectNodeModulesPackagePaths", () => {
```
LSTu5, line 7: **[test gap]** Add coverage for the two behaviors the PR's open questions hinge on: (a) a scoped package `@scope/name` -> coordinate `@scope/name` with the correct `installDir`; (b) the same coordinate appearing at two `node_modules` paths (hoisted + nested) to lock the keep-owned dedup behavior at the collection layer.

LSTu5:
Worked through the review — resolutions below (all in `47ae1ef`, full unit suite + system test green, no snapshot churn).

**Fixed**

- **#4 (facts.ts duplicated shape)** — `ApkPackageOwnershipsFact.data` now references the resolver's `ApkPackageOwnerships` type. Confirmed no real import cycle: `facts.ts` ↔ `analyzer/types.ts` already have a type-only cycle, and these are erased interfaces, so `tsc` is clean. (Left the pre-existing singular `ApkPackageOwnershipFact` alone — that's #872's code.)
- **#3 (dedup order-dependence)** — now sorts candidates by install dir before keeping the first owned occurrence, so the result is deterministic regardless of file-discovery order. *Reply:* this only makes the tiebreak **stable**; *which* origin is correct when two different apk packages genuinely bundle the same `name@version` is still the open `@snyk/vuln` question (#2), now called out in a code comment.
- **#6 (system-test brittleness)** — all three points addressed: switched to **docker.io** (the registry the rest of the suite already uses — verified no existing test touches cgr.dev), **digest-pinned** like `BASH_IMAGE`, and the assertion now checks the **invariant** (≥1 dependency with `originPackage === 'npm'`) instead of a specific bundled package. Verified on the pinned digest: 128 npm-owned deps surfaced (incl. scoped ones).
- **#2 / #7 (scoped pkgs + test gaps)** — added unit coverage for scoped packages (`@scope/name`, empirically confirmed working) and for one-entry-per-path on a coordinate at multiple depths; documented the `.includes("/node_modules/.")` heuristic.

**Fixed + reply (decisions)**

- **#1 (collection runs on every node image)** — **accepting** this, with a code comment explaining why: the `package.json` contents are already in memory so it's a cheap parse, it only runs on the node global-modules path, and gating by distro is impossible here because the scanner has no `osRelease` (that absence *is* the CN-1395 root cause). Threading distro into the scanner is a larger refactor out of scope for this PR.
- **#5 (response-builder robustness)** — added a comment naming the invariant that an npm result never carries **both** ownership facts (the whole-result path always fails closed on the shared root). *Reply:* kept the strip scoped inside the `if` (the carrier only ever exists there) and kept the `.push`-then-reassign style to match the surrounding code in this function — happy to restructure if preferred.

Still open before un-drafting (unchanged): the two `@snyk/vuln` contract questions (coordinate join key + dedup/per-location policy), which drive #3's semantic.

lib/analyzer/applications/node.ts
LSTu5: Accepting, with a code comment added in 47ae1ef explaining why: the package.json contents are already in memory (cheap parse), it only runs on the node global-modules path, and gating by distro is impossible here because the scanner has no osRelease — that absence is the CN-1395 root cause itself. Threading distro into the scanner is the larger refactor, out of scope for this PR.

lib/analyzer/applications/node.ts
LSTu5: Added a unit case for scoped packages in 47ae1ef, and empirically confirmed they already work — the pinned docker.io node image surfaces scoped owned deps (@npmcli/agent, @gar/promise-retry, ...). Also documented the `.includes("/node_modules/.")` guard (safe because npm names cannot start with ".").

lib/analyzer/package-managers/apk-ownership.ts
LSTu5: Fixed the non-determinism in 47ae1ef: candidates are now sorted by install dir before keeping the first owned occurrence, so the result no longer depends on file-discovery order. The remaining semantic — which origin is correct when two different apk packages genuinely bundle the same name@version — is the open @snyk/vuln question (per-location vs single path); noted in a code comment.

lib/facts.ts
LSTu5: Done in 47ae1ef — `data` now references the resolver`s `ApkPackageOwnerships` type. Confirmed there is no real import cycle: facts.ts and analyzer/types.ts already have a type-only cycle and these are erased interfaces, so tsc is clean. Left the pre-existing singular `ApkPackageOwnershipFact` as-is since it is #872`s code.

lib/response-builder.ts
LSTu5: Added a comment in 47ae1ef naming the invariant: an npm result never carries both facts, because the only path that emits the carrier (global modules) has the shared root as evidence, so the whole-result resolveApkOwnership always fails closed for it. Kept the strip scoped inside the `if` (the carrier only ever exists there) and kept the push-then-reassign style to match the surrounding function — happy to restructure if you prefer.

test/system/application-scans/chainguard-ownership.spec.ts
LSTu5: All three addressed in 47ae1ef: switched to docker.io (verified nothing else in the suite uses cgr.dev), digest-pinned like BASH_IMAGE, and the assertion now checks the invariant (>=1 dependency with originPackage === "npm") rather than brace-expansion specifically. Verified on the pinned digest: 128 npm-owned deps surface.

test/unit/node-package-paths.spec.ts
LSTu5: Added in 47ae1ef: a scoped-package case (@scope/name) and a multi-depth case asserting one entry per install path for the same coordinate (collection does not dedup; the resolver does).

lib/analyzer/applications/node.ts
LSTu5: Update: per team decision (Parker), we are gating to Chainguard rather than accepting the unconditional work. Implemented in f5cab7a with one predicate (isChainguardDistro) at two gates: static-analyzer now threads the check into the node scanner so per-package collection (parsing every node_modules package.json) only runs on Chainguard/Wolfi, and response-builder only builds the APK path index + runs resolution on Chainguard. Non-Chainguard images now skip ownership work entirely. Output is unchanged (ownership only ever fired on Chainguard); widening the distro set later is a one-line change.

lib/types.ts
```diff
@@ -84,7 +84,11 @@ export type FactType =
   // Application files observed in the image
   | "applicationFiles"
   // Chainguard/Wolfi: APK package that owns application dependency evidence paths
-  | "apkPackageOwnership";
+  | "apkPackageOwnership"
+  // Chainguard/Wolfi: per-dependency APK ownership for npm global modules
+  | "apkPackageOwnerships"
```
parker-snyk: We have `apkPackageOwnership` and `apkPackageOwnerships`

What is the difference between these? If the second one is for npm global modules, should it have a different name?

test/system/application-scans/chainguard-ownership.spec.ts
```diff
@@ -1,25 +1,26 @@
 import { scan } from "../../../lib/index";
 
+const BASH_IMAGE =
+  "chainguard/bash@sha256:642933df66209814502599053ca3dfa97cccf847badc4219d2b1fd6565f6559a";
```
parker-snyk: This might break in the CI depending on if it needs to be authed to pull this image. I believe chainguard only lets you pull the `:latest` freely. We may need to self host this to ensure access and that it doesn't change.

lib/analyzer/package-managers/apk-ownership.ts
```diff
@@ -192,6 +206,62 @@ export function resolveApkOwnership(
...
+): ApkPackageOwnerships | undefined {
+  if (!isChainguardDistro(osRelease) || packages.length === 0) {
+    return undefined;
+  }
+
+  const dependencies: OwnedDependency[] = [];
+  const seen = new Set<string>();
+  // Sort by install dir so dedup is deterministic regardless of file-discovery
+  // order. If a coordinate is bundled by more than one apk package (different
+  // origins), the shallowest/lexicographically-first install wins; which origin
```
parker-snyk: Is it shallowest or lexicographically first?

lib/analyzer/package-managers/apk-ownership.ts
```diff
@@ -192,6 +206,62 @@ export function resolveApkOwnership(
...
+ * Resolve APK ownership per dependency for ecosystems whose dependencies share
+ * one directory (npm global node_modules), where the whole-result
+ * resolveApkOwnership fails closed because the shared root is declared by
+ * several packages. Each dependency's own install directory is matched
+ * independently against the APK index; dependencies with no owning package are
+ * omitted (per-package fail-closed) so their findings are preserved. Results are
+ * deduplicated by name@version, keeping an owned occurrence over an unowned one.
+ */
+export function resolveApkOwnershipsPerDependency(
+  packages: Array<{ name: string; version: string; installDir: string }>,
```
parker-snyk: `name: string; version: string; installDir: string` seems to appear several times. Instead of redefining inline, would it make sense to import the type and use that?

lib/analyzer/package-managers/apk-ownership.ts
```diff
@@ -192,6 +206,62 @@ export function resolveApkOwnership(
...
+    a.installDir.localeCompare(b.installDir),
+  );
+  for (const pkg of ordered) {
+    const coordinate = `${pkg.name}@${pkg.version}`;
+    if (seen.has(coordinate)) {
+      continue;
+    }
+    const normalized = normalizeAbsolutePath(pkg.installDir);
+    const match = resolveOwnerForEvidencePath(normalized, index, symlinkGraph);
+    if (!match) {
```
parker-snyk: If the same `coordinate` that doesn't match appears several times, we will attempt to resolve it each time. Since `seen` is actually seen AND matched. Not a blocker, but consider if we can avoid resolving the same package more than once either way

lib/response-builder.ts
```diff
@@ -278,20 +283,55 @@ async function buildResponse(
...
+          };
+          appDepsScanResult.facts.push(ownershipsFact);
+        }
+      }
     }
 
+    // The transient carrier fact must never reach the output. The node scanner
+    // only emits it on Chainguard, but strip unconditionally as a safety net.
+    appDepsScanResult.facts = appDepsScanResult.facts.filter(
+      (fact) => fact.type !== "nodeModulesPackagePaths",
```
parker-snyk: We may not need this strip at all. It's only here because we push `nodeModulesPackagePaths` onto the public `facts[]` array back in `node.ts` (line 235), just to hand the install dirs to this function.

That data never leaves the plugin, so it shouldn't be going through the public fact contract to get here. If `node.ts` returns the install dirs as a normal value and we take them as a parameter, this filter goes away, and so does the `FactType` entry in `types.ts`. 

If we don't need to alter an external / exported type like `FactType` we should look for another way to pass the info

lib/response-builder.ts
```diff
@@ -278,20 +283,55 @@ async function buildResponse(
...
+      // APK package owns, so the whole-result resolution above fails closed.
+      // Resolve ownership per package instead, using the per-package install
+      // dirs the node scanner attached as a transient fact. The two facts are
+      // mutually exclusive: the transient fact is only emitted on the node
+      // global-modules path, whose evidence is the shared root.
+      const npmPackagePathsFact = appDepsScanResult.facts.find(
+        (fact) => fact.type === "nodeModulesPackagePaths",
+      );
+      if (npmPackagePathsFact) {
+        const ownerships = resolveApkOwnershipsPerDependency(
```
parker-snyk: nit: var name is similar to `ownership` used right above, but it's a different type. Similar names but they're both the exact same type `ApkPackageOwnerships`

This may be occuring for other singular / plural var names in other places

lib/analyzer/package-managers/apk-ownership.ts
```diff
@@ -21,6 +21,20 @@ export interface ApkPackageOwnership {
   evidencePaths: string[];
 }
 
+export interface OwnedDependency {
+  name: string;
+  version: string;
+  packageName: string;
+  packageVersion: string;
```
parker-snyk: nit: unclear at a glance what version is vs packageVersion. Consider changing to make more descriptive, are they apkVersion, npmVersion, etc?

parker-snyk (changes requested):

lib/types.ts
parker-snyk: On second thought, do we actually need this added? Do we need to pass registry two different but mostly similar types, one for go/jar another for npm? Or can we make it work for both?

LSTu5: Do you think we should merge these? Might cause more complexity later down the line, but I think now would be the time to decide.

parker-snyk: What additional complexity do you forsee if we were to merge these?

LSTu5: May be simpler and prevent duplicate code later, but would still have to check which kind you're dealing with when interacting with  and go back to 872 to address

lib/analyzer/applications/node.ts
```diff
@@ -89,10 +95,68 @@ export async function nodeFilesToScannedProjects(
   return scanResults;
 }
 
+const nodeModulesPackageJsonRegex = /\/node_modules\/.+\/package\.json$/;
```
parker-snyk: This regex matches a package.json at any depth under node_modules, and the extractor grabs every file named package.json (see `lib/inputs/node/static.ts`). So a bundled test fixture or example manifest gets collected as if it were a real installed package.

Concrete case: `/usr/lib/node_modules/npm/node_modules/brace-expansion/test/fixtures/package.json` declaring `{"name":"lodash","version":"4.17.21"}`. Its `installDir` is under npm's owned subtree, so it resolves to origin `npm` and we emit `lodash@4.17.21` as npm-owned.

If the image also has a real `lodash@4.17.21`, the two coordinates collide and @snyk/vuln can suppress the real dependency's vulns against the wrong owner, which is the false suppression the fail-closed design is supposed to prevent.

Can we anchor the coordinate to an actual package root, e.g. require the install dir's trailing segment to match the declared package name, instead of trusting any nested package.json?

lib/analyzer/applications/node.ts
```diff
@@ -89,10 +95,68 @@ export async function nodeFilesToScannedProjects(
...
+  }
+
+  const packages: NodeModulesPackagePath[] = [];
+  for (const filePath of files) {
+    // Skip dot-prefixed node_modules entries (.pnpm virtual store, .bin, .cache).
+    // Safe because npm package names cannot begin with ".", so a real package
+    // dir is never excluded by this guard.
+    if (
+      !nodeModulesPackageJsonRegex.test(filePath) ||
+      filePath.includes("/node_modules/.")
```
parker-snyk: This one's easy to miss. This guard and the regex on line 98 only match forward slashes, but image paths aren't always forward-slash. When the plugin runs on a Windows host, the extractor builds paths with the OS separator (`layer.ts` uses `path.sep`), so they arrive as `\usr\lib\node_modules\...`. You can see how the rest of the node code copes: `node-modules-utils.ts` matches `node_modules` with `[\\/]` (both separators), and the ownership resolver runs each path through `normalizeAbsolutePath` before matching.

Here the regex never matches a backslash path, so `collectNodeModulesPackagePaths` returns `[]` on Windows and no ownership fact is emitted, even though the same image works on Linux. We run a Windows CI job, so this can look fine locally and behave differently there. The fix is to accept both separators like the rest of the file, e.g. `[\\/]` in the regex. The tightened regex in my other comment already does this.

test/unit/response-builder-ownership.spec.ts
```diff
@@ -69,8 +69,139 @@ describe("buildResponse apkPackageOwnership", () => {
...
+          originPackage: "npm",
+          apkPackageName: "npm",
+        },
+      ],
+    });
+
+    // The internal install-dir data must never reach the public ScanResult.
+    expect((appResult as any).nodeModulesPackagePaths).toBeUndefined();
+    expect(
+      appResult.facts.find((f) => f.type === "nodeModulesPackagePaths"),
```
parker-snyk: `nodeModulesPackagePaths` is a field on the scan result, not a `FactType` (it isn't in the union), so nothing ever emits a fact with that type. This `find()` always returns `undefined`, so the assertion can't fail. jest runs with `isolatedModules`, so the type mismatch never gets caught either.

The line right below it, `(appResult as any).nodeModulesPackagePaths` being undefined, is the one that actually tests the strip. I'd drop this fact check (and the identical one in `chainguard-ownership.spec.ts`) so it doesn't read as a guard that isn't really there. Looks like a leftover from when the carrier was still a fact.

parker-snyk (commented):
A few notes from a local review of the per-dependency npm ownership work. Two correctness ones on `node.ts` (bundled/nested package.json coordinates, and the forward-slash-only matching on Windows) plus one test cleanup. Nothing blocking on my end.

lib/analyzer/applications/node.ts
parker-snyk: Here's a tighter regex that only matches a real package root, a package.json whose parent is an immediate child of a `node_modules` dir (with an optional `@scope/`), instead of one at any depth:

```ts
const nodeModulesPackageJsonRegex =
  /[\\/]node_modules[\\/](?:@[^\\/]+[\\/])?[^\\/]+[\\/]package\.json$/;
```

What changes:
- `.../node_modules/npm/package.json` and `.../node_modules/npm/node_modules/brace-expansion/package.json` still match (real installs, including nested ones).
- `.../node_modules/@npmcli/agent/package.json` still matches (scoped).
- `.../node_modules/brace-expansion/test/fixtures/package.json` no longer matches, so a bundled fixture or example manifest stops being collected as its own package.

I used `[\\/]` instead of `/`, so this also folds in the backslash handling from the Windows comment on the guard below.

One caveat: this still trusts that the install dir reflects the package. If you also want to defend against a manifest whose `name` disagrees with its directory, add a check that the install dir's trailing segment matches the declared `name` (handling the `@scope/` case) before pushing it.

test/unit/apk-ownership.spec.ts
```diff
@@ -277,3 +292,181 @@ describe("apk-ownership", () => {
...
+      ],
+      index,
+      symlinkGraph,
+      wolfi,
+    );
+
+    expect(ownership).toBeUndefined();
+  });
+
+  it("dedupes a coordinate by name@version, keeping the owned occurrence", () => {
```
parker-snyk, line 402: These are quick to add in this per-package `describe`, and each pins a behavior the resolver relies on:

1. Owned-beats-unowned when the owned copy is first. The current dedup test lists the unowned copy first (`/app/...` sorts before `/usr/...`), so it only checks one order. Add a case with the same `name@version` where the owned copy comes first and an unowned copy second, and assert the owned one still wins. That's the only order that exercises the `seen.has(coordinate)` skip.
2. Same `name@version` owned by two different apk packages. This is the case the `localeCompare` sort exists to make deterministic. Add two candidates with that coordinate under different apk owners and assert which origin wins. Without it, someone could delete the sort and nothing would fail.
3. A candidate whose `evidencePaths` is `[]`. The "empty input" test passes an empty array of candidates, which is a different path from one candidate with empty `evidencePaths`. The latter is what hits the `evidencePaths[0] ?? ""` guard in the sort and the length check in the loop.

test/unit/evidence-paths.spec.ts
parker-snyk, line 1: On the existing `extractEvidencePaths` test at line 25 (`collects targetFile, testedFiles, and jar fingerprint locations`): it asserts with `arrayContaining` and no length check, while the sibling tests use exact `toEqual([...])`. This module exists to keep bogus paths out (the `/`-rooted and char-split cases), and a regression that emitted extra junk paths would still pass here as long as the three expected paths show up. Prefer `toEqual` plus `toHaveLength`. Not introduced by this PR, but the new coordinate work leans on this function, so worth tightening while we're here.

parker-snyk (commented):
Test coverage notes for the per-package ownership work: a few resolver cases the current specs don't pin down, plus one existing assertion worth tightening.

lib/analyzer/applications/node.ts
```diff
@@ -89,10 +95,68 @@ export async function nodeFilesToScannedProjects(
...
+    }
+
+    let parsed: { name?: string; version?: string };
+    try {
+      parsed = JSON.parse(content);
+    } catch {
+      continue;
+    }
+
+    if (!parsed.name || !parsed.version) {
```
parker-snyk: `JSON.parse` returns `any`, so the `{ name?: string; version?: string }` annotation isn't enforced at runtime. This guard only rejects falsy values, so a manifest like `{"name": 123, "version": 4.5}` gets through and we push non-string values into `NodeModulesPackagePath`, whose fields are typed `string`. They flow into the coordinate (`123@4.5`, or `[object Object]@` for an object name) and out to the `apkPackageOwnership` fact, where a consumer doing string ops on `name`/`version` could break.

Low-risk in practice, since anything published to npm has string `name`/`version`. It mostly matters alongside the bundled-manifest concern from the other thread, where the input isn't guaranteed to be a real package.

Which are you after here, that `name`/`version` are actually strings, or just that they're present? If you want the string guarantee, a `typeof` check covers it, and it still needs to reject empty strings:

```ts
if (
  typeof parsed.name !== "string" ||
  typeof parsed.version !== "string" ||
  !parsed.name ||
  !parsed.version
) {
  continue;
}
```

parker-snyk (commented):
One more on the package.json parsing: the name/version guard only checks for presence, not type. Left a question inline on whether that's intentional.

parker-snyk (approved):
LGTM

PR #891, opened 2026-07-28
feat(provenance): fetch provenance attestations from the registry

bdemeo12:
## What

When Snyk scans a **registry image reference** (ex: `snyk container monitor myregistry/app:1.2.3`), the image is obtained via `docker pull` / `docker save`, which **drop the `unknown/unknown` attestation manifest**. So provenance never reaches the parser and the asset shows none, even though the registry has it.

This change attempts to fetch the provenance attestation from the source registry when a scan shows no attestations locally.

## How
- New `lib/extractor/registry-attestations.ts`:
  - `fetchAttestationsFromRegistry()` uses `@snyk/docker-registry-v2-client`'s **`getAttestationManifest`** (finds the platform's attestation manifest) + **`getLayer`** (downloads each in-toto blob), parses the in-toto statements, and returns `ResolvedAttestationManifest[]`.
  - Best-effort: any failure (no attestation, unsupported registry, auth/network) → `[]`, so it never breaks a scan.
- `lib/static.ts`: for `ImageType.Identifier` scans where extraction found no attestations, fetch from the registry and set `analysis.attestations` before `buildResponse` (which already turns them into the `provenanceMetadata` fact via `parseProvenanceAttestations`).

## Why it matters

Tested end-to-end in container-integrations: a CI job builds the image with SLSA provenance, pushes it to a registry, and runs snyk container monitor on the pushed reference. The scan fetched the attestation from the registry and the image appeared in the dev inventory with its provenance populated!

Screenshot 2026-07-15 at 10 58 17 AM

https://app.circleci.com/pipelines/gh/snyk/container-integrations/117/details?useNewPipelines=true&job=8e5fc580-cbdd-4bf6-bf03-1c23d20b0a2f&workflowId=f8bd1925-1f54-435f-aaec-8053f80a2115&buildNumber=1712&jobType=build#step-102-


bdemeo12:
addressing: 

>When scanning an image by digest (e.g., my-repo@sha256:123...), extractImageDetails returns the digest in a separate field, and the tag field may be empty. In analyzeStatically, the code only passes tag as the imageReference. This will cause fetchAttestationsFromRegistry to fail to locate the correct attestation manifest for digest-pinned images, as it will be missing the required identifier.

When you scan by digest, the digest gets put into the tag variable. So passing tag already passes the digest.

bdemeo12:
Responding to this review comment: 

>[Missing Digest Validation](https://github.com/snyk/snyk-docker-plugin/pull/891/changes#diff-455f0ad65de5470abd061037cd35ad337a211f612f9dadfe77c7af0900f01583R69-R103) 🟠 [major]
The function fetchAttestationsFromRegistry fetches attestation manifests but does not verify that the attestation actually refers to the image being scanned. While it filters layers by SLSA predicate type, it lacks the logic found in lib/extractor/oci-archive/layer.ts (specifically extractAttestations) which checks the vnd.docker.reference.digest annotation against the image's manifest digest. Without this, a registry could return attestations for a different tag or version of the image if the referrer implementation is loose, leading to incorrect provenance being associated with the scan result.

This is not a major risk.  Even when it triggers, the provenance content is the same. All platforms are apart of the same build, so they share the same Dockerfile, git commit, source repo, build type, and build time.

kateeselius (approved):
🚢 🥳

